PR #15, opened 2016-12-26
feat(): captcha to mitigate abuse, closes #9

nils-wisiol:
Introduce captcha to prevent abuse while maintaining easy usability.

PR #18, opened 2016-12-26
fix(): update email signatures

peterthomassen:


peterthomassen:
wait for merge of #15 and then adjust `api/desecapi/templates/emails/captcha/content.txt` as well

nils-wisiol:
#15 is merged, so this PR should be updated and rebased

peterthomassen:
I rebased this on top of `20170120_POSTwithRecords` which is supposed to be merged first.

Regarding putting the footer in a separte file -- can we do this next time we touch this? Then this would be ready to merge now.

PR #27, opened 2017-01-20
fix pdns A/AAAA record sync for new domains, add test

peterthomassen:


api/desecapi/models.py
```diff
@@ -128,13 +128,15 @@ def pdns_sync(self):
         if new_domain:
             pdns.create_zone(self.name)
 
-        # for existing domains, see if records are changed
-        if not new_domain:
+        # check if current A and AAAA record values require updating pdns
+        if new_domain:
+            changes_required = (self.arecord or self.aaaarecord)
```
nils-wisiol: should be `bool(a or b)`, otherwise changes_required holds a string, which is unexpected

peterthomassen:
done. ready to merge?

PR #28, opened 2017-01-20
fix(devadmin): enforce PHP 5

peterthomassen:
The docker image richarvey/nginx-php-fpm that we're using has switched
its default to PHP 7, breaking the way our poweradmin is set up.

This PR enforces the use of the PHP 5 docker image.


peterthomassen:
Test is now working after rebasing on top of the other PRs.

PR #29, opened 2017-01-23
feat(network): BREAKING CHANGE internal network refactoring

peterthomassen:
- make IPv4 internal net prefix configurable
- restructure internal networks into several smaller ones, to restrict
  container visibility

NOTE: Changing the IPv4 internal net prefix breaks existing installations,
and MySQL tables need to be updated manually. See README.


docker-compose.yml
```diff
@@ -161,20 +168,52 @@ volumes:
   dbmaster_mysql:
 
 networks:
-  back1: # TODO can we declare this internal? compose 1.9.0 will allow it: https://github.com/docker/compose/pull/3488 May break apt inside nslord etc.
+  # Note that it is required that the front network ranks lower (in lexical order)
```
nils-wisiol: lexical order by name?

peterthomassen: yes, fixed.

peterthomassen:
I agree with creating an `UPGRADE.md` in the future. At that point, we also should have release tags so that we can refer properly to which upgrade path is being described.

nils-wisiol:
alright, then this PR lgtm

PR #31, opened 2017-01-26
feat(api): implement zone deletion

peterthomassen:
This implements zone deletion.

To do:
- write tests


README.md
```diff
@@ -54,6 +54,7 @@ Although most configuration is contained in this repository, some external depen
       - `DESECSTACK_DBMASTER_SUBJECT_ns1replication`: slave 1 replication SSL certificate subject name
       - `DESECSTACK_DBMASTER_PASSWORD_ns2replication`: slave 2 replication password
       - `DESECSTACK_DBMASTER_SUBJECT_ns2replication`: slave 1 replication SSL certificate subject name
+      - `DESECSTACK_NSMASTER_APIKEY`: pdns API key on nsmaster (required so that we can execute zone deletions on the slaves)
```
nils-wisiol: on master, to be acurate

api/desecapi/pdns.py
```diff
@@ -14,31 +19,44 @@ def normalize_hostname(name):
         raise Exception('Invalid hostname ' + name)
     return name if name.endswith('.') else name + '.'
 
+def _pdns_delete(url):
```
nils-wisiol: add some doc why two requests are needed

api/desecapi/pdns.py
```diff
@@ -98,7 +123,7 @@ def set_dyn_records(name, a, aaaa):
     """
     Commands pdns to set the A and AAAA record for the zone with the given name to the given record values.
     Only supports one A, one AAAA record.
-    If a or aaaa is None, pdns will be commanded to delete the record.
+    If a or aaaa is empty, pdns will be commanded to delete the record.
```
nils-wisiol: hah, good catch

api/desecapi/pdns.py
```diff
@@ -108,3 +133,18 @@ def set_dyn_records(name, a, aaaa):
             _delete_or_replace_rrset(name, 'aaaa', aaaa),
         ]
     })
+
+
+def set_rrset(zone, name, type, value):
```
nils-wisiol: The `name` and `zone` argument seem redundant, or I could not figure out why they ever should be set to different values. Cf. code comment below: `... for the zone with the given name`.

The `type` argument shadows the python3 built-in function `type()`, which should be avoided. Suggestion: `rr_type`. (This is also true for already existing `type` arguments in the code.)

nils-wisiol: This function is very similar to `set_dyn_records()`. I'd prefer that `set_dyn_records()` uses the `set_rrset()` function to do what needs to be done, however in the current state this is only possible with two pdns api requests, which is certainly not desirable. Should `set_rrset` be extended to `set_rrsets`?

api/desecapi/views.py
```diff
@@ -91,13 +91,6 @@ class DomainDetail(generics.RetrieveUpdateDestroyAPIView):
     def get_queryset(self):
         return Domain.objects.filter(owner=self.request.user.pk)
 
-    def put(self, request, pk, format=None):
```
nils-wisiol: extend commit msg to explain why this can be safely deleted

nsmaster/conf/pdns.conf.var
```diff
@@ -1,10 +1,15 @@
...
 allow-unsigned-supermaster=yes
+api=yes
+api-key=${DESECSTACK_NSMASTER_APIKEY}
 disable-axfr=yes
 setgid=pdns
 setuid=pdns
 slave=yes
 version-string=powerdns
+webserver=yes
+webserver-address=0.0.0.0
```
nils-wisiol: could that cause the travis error?

test-api
```diff
@@ -44,4 +43,3 @@ docker-compose -f docker-compose.yml -f docker-compose.dev.yml -f docker-compose
 docker-compose ps
 
 exit $(docker-compose ps -q | xargs docker inspect -f '{{ .State.ExitCode }}' | grep -v 0 | wc -l | tr -d ' ')
-
```
nils-wisiol: why remove the newline? git always warns me about missing newlines at the end of files

api/desecapi/tests/testdomains.py
```diff
@@ -177,13 +196,48 @@ def testDomainDetailURL(self):
...
 
+    def testCanDeleteOwnedDynDomain(self):
+        httpretty.enable()
+        httpretty.register_uri(httpretty.DELETE, settings.NSLORD_PDNS_API + '/zones/' + self.ownedDomains[1].name + '.')
+        httpretty.register_uri(httpretty.DELETE, settings.NSMASTER_PDNS_API + '/zones/' + self.ownedDomains[1].name+ '.')
+        httpretty.register_uri(httpretty.PATCH, settings.NSLORD_PDNS_API + '/zones/dedyn.io.')
+
+        url = reverse('domain-detail', args=(self.ownedDomains[1].pk,))
+        response = self.client.delete(url)
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
```
nils-wisiol: What's the status code for deletion of non-existing domains? Should it be 200? Cf next comment.

api/desecapi/tests/testdomains.py
```diff
@@ -177,13 +196,48 @@ def testDomainDetailURL(self):
...
+        self.assertTrue(isinstance(httpretty.last_request(), httpretty.core.HTTPrettyRequestEmpty))
+
+    def testCantDeleteOtherDynDomains(self):
+        httpretty.enable()
+        httpretty.register_uri(httpretty.DELETE, settings.NSLORD_PDNS_API + '/zones/' + self.otherDomains[1].name + '.')
+        httpretty.register_uri(httpretty.DELETE, settings.NSMASTER_PDNS_API + '/zones/' + self.otherDomains[1].name+ '.')
+
+        url = reverse('domain-detail', args=(self.otherDomains[1].pk,))
+        response = self.client.delete(url)
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
```
nils-wisiol: If 200 should be returned for non-existing domains, should be also return that for domains the user does not have access to? Or should we return 404 to increase privacy? But isn't the domain data public anyways? (So it should be 403?)

api/desecapi/pdns.py
peterthomassen: Regarding `zone` and `name`: When deleting `DS` records for `foobar.dedyn.io`, those are part of the parent zone `dedyn.io`. We thus call `set_rrset("dedyn.io", "foobar.dedyn.io", "DS", "")`.

Regarding the similarity between `set_dyn_records` and `set_rrset`: I agree that it would be nice is the former used the latter, but the two resulting requests are not desirable. Extension of `set_rrset` to `set_rrsets` is indeed the solution, but I am planning to refactor this when we generalize the API to allow general request proxying to pdns, for arbitrary record management. At that point, we'll have to rethink the design of `pdns.py` anyways.

nsmaster/conf/pdns.conf.var
peterthomassen: No, the error referred to port 3306 specifically, which is the MySQL port and unrelated to the nsmaster API (port 8081). I realized that nsmaster/dbmaster don't need to run because everything is caught by httpretty. I just verified that removing nsmaster/dbmaster from the test suite fixes the problem. (And it's in line with the pdns-independent test design, so nice!)

test-api
peterthomassen: I did not know that git gives you a warning. I don't care, adding it back.

api/desecapi/tests/testdomains.py
peterthomassen: We resource to be deleted can't be fetched, i.e. `/api/v1/domains/<N>/` where the number `<N>` does not identify a domain cannot be found. This causes 404.

api/desecapi/tests/testdomains.py
peterthomassen: Hm, that's a good point. But I think 404 is ok. Let me know if you have strong preference for 403, but I don't really think we need it.

peterthomassen:
Nils, thanks a lot for the thorough review!

I addressed your comments and the Travis build went through. Looks ok?

api/desecapi/pdns.py
nils-wisiol: Ah, the names and/or docs should be clarified then.

api/desecapi/tests/testdomains.py
nils-wisiol: `DELETE` requests should be idempotent, that requires us to return 200/204 if the resource requested does not exist: https://tools.ietf.org/html/rfc7231#page-23

I can see though that some interpretations of this RFC may conclude that the idempotency does not extend to the response code given. However, this would only require the client to handle more reponse codes with the same meaning; hence I'd return 200/204 for deleted and non-existing resources, 403 for domains of other users. (Unless there is a good point why this is an unacceptable privacy intrusion.)

api/desecapi/tests/testdomains.py
nils-wisiol: see answer above.

nils-wisiol:
other than the status code discussion, this lgtm

api/desecapi/tests/testdomains.py
peterthomassen: I'll open a separate issue for this.

PR #34, opened 2017-02-13
fix(db): improve foreign keys for dblord/dbmaster

peterthomassen:
Previously, we only had foreign key contraints for the `records` and the `domainmetadata` table.

Now we're also adding them for the `comments` and `cryptokeys` table. The SQL statements are taken from https://github.com/PowerDNS/pdns/blob/master/modules/gmysqlbackend/enable-foreign-keys.mysql.sql

### DIFF
--- a/dblord/initdb.d/11-pdns-lord-REFERENCES.sql
+++ b/dblord/initdb.d/11-pdns-lord-REFERENCES.sql
@@ -1,5 +1,16 @@
 USE pdns;
 
--- As recommended by https://doc.powerdns.com/md/authoritative/backend-generic-mysql/
-ALTER TABLE `records` ADD CONSTRAINT `records_ibfk_1` FOREIGN KEY (`domain_id`) REFERENCES `domains` (`id`) ON DELETE CASCADE;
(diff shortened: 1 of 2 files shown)


dblord/initdb.d/11-pdns-lord-REFERENCES.sql
```diff
@@ -1,5 +1,16 @@
 USE pdns;
 
--- As recommended by https://doc.powerdns.com/md/authoritative/backend-generic-mysql/
-ALTER TABLE `records` ADD CONSTRAINT `records_ibfk_1` FOREIGN KEY (`domain_id`) REFERENCES `domains` (`id`) ON DELETE CASCADE;
-ALTER TABLE `domainmetadata` ADD CONSTRAINT `domainmetadata_ibfk_1` FOREIGN KEY (`domain_id`) REFERENCES `domains` (`id`) ON DELETE CASCADE;
+-- From https://github.com/PowerDNS/pdns/blob/master/modules/gmysqlbackend/enable-foreign-keys.mysql.sql
```
nils-wisiol: should reference specific commit

Issue #25, opened 2017-01-15
api: add creation/update timestamps to domain list view


peterthomassen:
Currently, the domain list view does not return creation/update timestamps. Should we add them?

(And why is it not the case so far?)

PR #36, opened 2017-02-20
api: Improve error handling

peterthomassen:
- better error codes
- make domain DELETE idempotent
- wrap Domain database writes in transaction, commit only when pdns operations are successful

TODO:
- fix testCantDeleteOtherDomains test
    - done
- return a reasonable error if pdns operations fail (currently 500)
    - done (if domain already exists, return 409)
- If a user is acting (GET, DELETE, ...) on a domain that belongs to another user, we return 404 (because the Django queryset is empty). 403 may be more appropriate, but it leaks a bit of information
    - not doing this now, problem is rather esoteric

### DIFF
--- a/api/desecapi/models.py
+++ b/api/desecapi/models.py
@@ -1,5 +1,5 @@
 from django.conf import settings
-from django.db import models
+from django.db import models, transaction
 from django.contrib.auth.models import (
     BaseUserManager, AbstractBaseUser
 )
@@ -98,6 +98,26 @@ class Domain(models.Model):
     arecord = models.CharField(max_length=255, blank=True)
     aaaarecord = models.CharField(max_length=1024, blank=True)
     owner = models.ForeignKey(settings.AUTH_USER_MODEL, related_name='domains')
+    _dirtyRecords = False
+
+    def __setattr__(self, attrname, val):
+        setter_func = 'setter_' + attrname
+        if attrname in self.__dict__ and callable(getattr(self, setter_func, None)):
+            super(Domain, self).__setattr__(attrname, getattr(self, setter_func)(val))
+        else:
+            super(Domain, self).__setattr__(attrname, val)
+
+    def setter_arecord(self, val):
+        if val != self.arecord:
+            self._dirtyRecords = True
+
+        return val
+
+    def setter_aaaarecord(self, val):
+        if val != self.aaaarecord:
+            self._dirtyRecords = True
+
+        return val
 
     def pdns_resync(self):
         """
@@ -112,7 +132,7 @@ class Domain(models.Model):
         # update zone to latest information
         pdns.set_dyn_records(self.name, self.arecord, self.aaaarecord)
 
-    def pdns_sync(self):
+    def pdns_sync(self, new_domain):
         """
         Command pdns updates as indicated by the local changes.
         """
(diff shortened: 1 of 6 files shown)


nils-wisiol:
also fix error code for domains that already exist in the pdns db

peterthomassen:
also add test verifying transaction behavior

peterthomassen:
@nils-wisiol I addressed all action items brought up here so far. Can you please review? Thanks

peterthomassen:
Added commit 209f70ed05643798ba26086cbbd1128be7046277, changing serializer behavior (see issues #16, #25) and enforcing an error when a user tries to change a domain name.

Questions:
- Should [this line](https://github.com/desec-io/desec-stack/pull/36/commits/209f70ed05643798ba26086cbbd1128be7046277#diff-d93a995458c0f3a31fb831c4f43b5d27R159) somehow be moved into the model instance constructor, so that we always store a copy of the unmodified instance in `self._orig`? That way, accessing the original instance (like in `clean_fields()`) is easier, and we save a database lookup.
- When registering a user, a JSON representation with the `email` and `id` fields is returned. I cannot find where this is specified (the user model has many more fields, and we have not written our own serializer). I'll need to answer this question in order to remove the `id` field from the response.

api/desecapi/pdns.py
```diff
@@ -100,14 +104,14 @@ def create_zone(name, kind='NATIVE'):
             "ns2.desec.io."
         ]
     }
-    _pdns_post('/zones', payload)
+    return _pdns_post('/zones', payload)
```
nils-wisiol: do not expose internal objects, i.e. don't return the response object

api/desecapi/pdns.py
```diff
@@ -100,14 +104,14 @@ def create_zone(name, kind='NATIVE'):
...
 
 
 def delete_zone(name):
     """
     Commands pdns to delete a zone with the given name.
     """
-    _pdns_delete('/zones/' + normalize_hostname(name))
+    return _pdns_delete('/zones/' + normalize_hostname(name))
```
nils-wisiol: do not expose internal objects, i.e. don't return the response object

api/desecapi/pdns.py
```diff
@@ -131,7 +135,7 @@ def set_dyn_records(name, a, aaaa):
     """
     name = normalize_hostname(name)
 
-    _pdns_patch('/zones/' + name, {
+    return _pdns_patch('/zones/' + name, {
```
nils-wisiol: do not expose internal objects, i.e. don't return the response object

api/desecapi/pdns.py
```diff
@@ -147,7 +151,7 @@ def set_rrset(zone, name, type, value):
     zone = normalize_hostname(zone)
     name = normalize_hostname(name)
 
-    _pdns_patch('/zones/' + zone, {
+    return _pdns_patch('/zones/' + zone, {
```
nils-wisiol: do not expose internal objects, i.e. don't return the response object

api/desecapi/models.py
```diff
@@ -132,25 +131,33 @@ def pdns_sync(self):
...
         pdns.delete_zone(self.name)
         if self.name.endswith('.dedyn.io'):
             pdns.set_rrset('dedyn.io', self.name, 'DS', '')
             pdns.set_rrset('dedyn.io', self.name, 'NS', '')
-        super(Domain, self).delete(*args, **kwargs)
 
+    @transaction.atomic
     def save(self, *args, **kwargs):
+        try:
+            self._orig = Domain.objects.get(pk=self.pk)
```
nils-wisiol: use `__setattr__` to track changes https://stackoverflow.com/questions/2898711/django-model-fields-getter-setter

api/desecapi/views.py
```diff
@@ -62,7 +62,15 @@ def perform_create(self, serializer):
             ex.status_code = status.HTTP_403_FORBIDDEN
             raise ex
 
-        obj = serializer.save(owner=self.request.user)
+        try:
+            obj = serializer.save(owner=self.request.user)
+        except Exception as e:
+            if str(e).endswith(' already exists"}'):
+                ex = ValidationError(detail={"detail": "This domain name is unavailable.", "code": "domain-unavailable"})
```
nils-wisiol: use the same error message for 'domain-taken' above

api/desecapi/models.py
```diff
@@ -99,6 +100,12 @@ class Domain(models.Model):
     aaaarecord = models.CharField(max_length=1024, blank=True)
     owner = models.ForeignKey(settings.AUTH_USER_MODEL, related_name='domains')
 
+    def clean_fields(self, exclude=None):
+        if self._orig is not None and self.name != self._orig.name:
```
nils-wisiol: needs to be updated, see comment above

peterthomassen:
I addressed your suggestions, and things are pretty much ready. The last remaining question is question 2 from https://github.com/desec-io/desec-stack/pull/36#issuecomment-281702688

peterthomassen:
The last remaining question was also addressed (the answer was `djoser.serializers`). Merge?

PR #37, opened 2017-04-10
feat(api): DNS challenge support for Let's Encrypt

nils-wisiol:
Provides easy setting of TXT records at `_acme-challenge.yourname.dedyn.io.`

A certbot hook script can be found at https://github.com/desec-io/certbot-hook

PR #38, opened 2017-04-11
cronhook, privacy chores

nils-wisiol:
Added cronhook that executs the privacy chores.

Chosen 2:55am UTC without any particular reason.

Environment get dumped into container at /etc/environment. Potential security thread? It contains sensitive data.

Closes nils-wisiol/desec-internal#9.

### DIFF
--- a/api/cronhook/start-cron.sh
+++ b/api/cronhook/start-cron.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+# start-cron.sh
+
+printenv >> /etc/environment
+touch /var/log/cron.log
+cron
+tail -F -v /var/log/cron.log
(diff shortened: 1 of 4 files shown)


api/cronhook/start-cron.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# start-cron.sh
+
+printenv >> /etc/environment
```
peterthomassen: As you already said in https://github.com/desec-io/desec-stack/pull/38#issue-220883537, this writes out sensitive data. Can you justify why it is needed?

nils-wisiol: Django needs to talk to `dbapi` to perform correctly, and it needs credentials to do so. Those credentials are stored in the environment; however cron does not by default provide the env it is running in to the started cron jobs. Hence, we need to pass on the env to the command line django by ourselves.

I will also argue that it does not pose an additional security thread: an attacker that can read from `/etc/environment` can likely also read env variables.

nils-wisiol:
rebased onto updated #37

peterthomassen:
lgtm

PR #39, opened 2017-04-11
provide $DESECSTACK_DOMAIN env to www container

nils-wisiol:
provide `$DESECSTACK_DOMAIN` env to www container

this is needed for PR desec-io/desec-www#5


docker-compose.yml
```diff
@@ -10,6 +10,8 @@ services:
     volumes:
     - ${DESECSTACK_WWW_CERTS}:/etc/ssl/private:ro
     - ./www/html:/usr/share/nginx/html:ro
+    environment:
+    - DESECSTACK_DOMAIN
```
peterthomassen: This variable is undocumented. I seems that it is only needed for the nginx configuration introduced in https://github.com/desec-io/desec-www/pull/5.

- Doesn't this overlap with `DESECSTACK_API_ALLOWED_HOSTS`? Why not stick to just one variable? (We can rename the latter if that would be more appriopriate.)
- If there's a good reason to have two variables, let's add documentation in both the `README.md` and `.env.default` files.

nils-wisiol: They do overlap, in fact we have

    DESECSTACK_API_ALLOWED_HOSTS = desec.DESECSTACK_DOMAIN update.dedyn.DESECSTACK_DOMAIN update6.dedyn.DESECSTACK_DOMAIN

Adopted `README.md`, `.env.default`, `settings.py`, `docker-compose.yml` to reflect that fact. `DESECSTACK_API_ALLOWED_HOSTS` is not needed anymore.

A disadvantage of this clean cut is a dependency on desec-io/desec-www@9485f5b, any previous version won't work due to missing `DESECSTACK_API_ALLOWED_HOSTS` provided to the `api` environment.

PR #48, opened 2017-04-18
IP Address Validation

nils-wisiol:
This PR introduces IP address validation before we pass values on to pdns. To that end, we use django and rest framework validation routines, hence the model field change.

This PR also establishes the policy that IP address values cannot be `""` anymore, but need to be `null` in order to trigger the removal of the a or aaaa record.

The PR also cleans up `pdns.py` a bit by unshadowing the type keyword and by removing the exposure of 'internal' objects.

Last, the unit tests are adopted to deal with the new IP address validation.

**Upon deployment, database needs to be migrated and will likely ask for a default value for the fields that cannot be 'blank' anymore.**


peterthomassen:
deployed, default values updated in db

Issue #44, opened 2017-04-18
api: REST endpoints for RRset at domains/<domain>/rrsets/<type>


peterthomassen:
Checklist:
- [ ] implementation
- [ ] unit tests
- [ ] docs

peterthomassen:
implemented by https://github.com/desec-io/desec-stack/commit/f13b45673ee44bd710cbc2fabc51f1c66da111db

Issue #43, opened 2017-04-18
api: REST endpoint domains/<domain>/rrsets/ to proxy pdns API requests


peterthomassen:
Limitations:
- do not allow zone rename
- do not support pdns fields disabled, set-ptr, comments

Checklist:
- [ ] implementation
- [ ] unit tests
- [ ] docs (try readthedocs.org?)

peterthomassen:
implemented by https://github.com/desec-io/desec-stack/commit/f13b45673ee44bd710cbc2fabc51f1c66da111db

PR #50, opened 2017-06-07
20170418 rrsets endpoint

peterthomassen:
This addresses issues #43, #44 by introducing a REST API for manipulation of DNS information. The PR contains the following:

- REST API for manipulation of DNS information. Before reviewing, it may be helpful to read about usage and limitations (e.g. the inaccessibility of the `SOA` record) in the docs.
- Tests for the new API
- Documentation for both "old" (domain management) and the new additions, available in the `docs/` directory. For easier reading, a compiled version of the current docs status is provided at https://a4a.de/.u/1496884799_index.html (URL valid for 7 days)
- A management command that imports some or all DNS information from pdns, from zones that have corresponding domain objects in the API database. (This allows pdns to operate zones independently of this API.)

What's left to do?

- Improve error handling. There are some uncaught or not very smart exceptions at the moment.
- Figure out how to deal with pdns' domain ID (we currently assume it's identical to the domain name)
- Think about how `HEAD` and `OPTIONS` requests should be answered (currently we have the default Django behavior, which does not always seem to be useful)
- Add tests for inaccessibility of the SOA record and for the pdns import
- Limit manual storage of DNSSEC-related information: Currently, there are no limitations to which record types can be stored / accessed via the API (except SOA). We may want to limit the ones related to DNSSEC, as we're taking care of that stuff automatically.


peterthomassen:
Updated with improved exceptions and according docs. Compiled docs: https://a4a.de/.u/1497020624_index.html

peterthomassen:
I addressed a few issues in the most recent commits:

- Thought-through exception handling / HTTP status codes
- disallowed POST'ing on read-only endpoints
- disallowed tampering with RRSIG, DNSKEY and NSEC3PARAM records
- added tests for tampring with SOA, RRSIG, DNSKEY and NSEC3PARAM records
- added tests for importing zones from pdns database
- improved some other tests (e.g. made sure the right number of RRsets is returned when querying)
- Bring docs up to date for all of this

@nils-wisiol I think this PR is now in good shape, definitely good enough for a tight look / review.

Remaining questions:
- Do we need the created/updated fields for RRsets? (When syncing from pdns, everything is deleted, so these get reset. In such cases, their contents may turn out misleading.)
- Do we need IsDomainOwner permission in the RRset views? (We also require ownership in the queryset, so that we get 404 instead of 401/403 when reading a foreign domain; the IsDomainOwner permission thus seems redundant.)
- Should we lowercase domain names on Domain object creation (via POST)? (We do lowercase the `subname` field of RRset.)
- Currently, domain names cannot contain `--`. Should we lift this restriction to support IDNs? Should we block any other patterns?
- Deal with pdns' domain ID (this is almost always identical to the zone name -- but not always).

peterthomassen:
Upated documentation to reflect changes in the two previous commits: https://a4a.de/.u/1497888712_index.html

api/desecapi/permissions.py
```diff
@@ -9,3 +9,12 @@ class IsOwner(permissions.BasePermission):
     def has_object_permission(self, request, view, obj):
         return obj.owner == request.user
 
+
+class IsDomainOwner(permissions.BasePermission):
+    """
+    Custom permission to only allow owners of a domain to view or edit an object owned by that domain.
```
nils-wisiol: Misleading doc: this permission doesn't do anything on it's own; only if used it makes sure that the requests user is the object's domain owner.

api/desecapi/views.py
```diff
@@ -120,6 +122,95 @@ class DomainDetailByName(DomainDetail):
...
 
+class RRsetDetail(generics.RetrieveUpdateDestroyAPIView):
+    lookup_field = 'subname'
+    serializer_class = RRsetSerializer
+    permission_classes = (permissions.IsAuthenticated, IsDomainOwner,)
+
+    def get_queryset(self):
+        name = self.kwargs['name']
+        subname = self.kwargs['subname']
+        type = self.kwargs['type']
```
nils-wisiol: shadows python `type`

api/desecapi/views.py
```diff
@@ -120,6 +122,95 @@ class DomainDetailByName(DomainDetail):
...
+        try:
+            return super().delete(request, *args, **kwargs)
+        except django.core.exceptions.ValidationError as e:
+            ex = ValidationError(detail={"detail": str(e)})
+            ex.status_code = status.HTTP_409_CONFLICT
+            raise ex
+
+    def get(self, request, *args, **kwargs):
+        subname = self.kwargs['subname']
+        type = self.kwargs['type'].upper()
```
nils-wisiol: shadows python `type`

api/desecapi/views.py
```diff
@@ -120,6 +122,95 @@ class DomainDetailByName(DomainDetail):
...
+
+
+class RRsetsDetail(generics.ListCreateAPIView):
+    serializer_class = RRsetSerializer
+    permission_classes = (permissions.IsAuthenticated, IsDomainOwner,)
+
+    def get_queryset(self):
+        name = self.kwargs['name']
+        subname = self.kwargs.get('subname')
+        type = self.kwargs.get('type')
```
nils-wisiol: shadows python `type`

api/desecapi/migrations/0015_rrset.py
```diff
@@ -0,0 +1,38 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='RRset',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created', models.DateTimeField(auto_now_add=True)),
+                ('updated', models.DateTimeField(null=True)),
+                ('subname', models.CharField(blank=True, max_length=178)),
+                ('type', models.CharField(max_length=10)),
```
nils-wisiol: `type` should also be renamed here, cf other comments

api/desecapi/models.py
```diff
@@ -227,3 +230,68 @@ def save(self, *args, **kwargs):
...
+    subname = models.CharField(max_length=178, blank=True)
+    type = models.CharField(max_length=10)
+    records = models.CharField(max_length=64000, blank=True)
+    ttl = models.PositiveIntegerField(validators=[MinValueValidator(1)])
+    _dirty = False
+
+    class Meta:
+        unique_together = (("domain","subname","type"),)
+
+    def __setattr__(self, attrname, val):
```
nils-wisiol: duplicate code (cf `Domain.__setattr__`). Move to mixin?

api/desecapi/models.py
```diff
@@ -227,3 +230,68 @@ def save(self, *args, **kwargs):
...
+        serializer = RRsetSerializer(self)
+
+        pdns.set_rrsets(self.domain.name, [serializer.data])
+        pdns.notify_zone(self.domain.name)
+
+
+    @transaction.atomic
+    def delete(self, *args, **kwargs):
+        if self.type == 'SOA':
+            raise ValidationError('You cannot touch the SOA record')
```
nils-wisiol: Note that the SOA record is not actually in our database, and this error is just for the user's convenience

api/desecapi/models.py
```diff
@@ -227,3 +230,68 @@ def save(self, *args, **kwargs):
...
+
+
+class RRset(models.Model):
+    # TODO Do these two fields really make sense? Meaning is limited when deleting + recreating an RRset
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(null=True)
+    domain = models.ForeignKey(Domain, on_delete=models.CASCADE, related_name='rrsets')
+    subname = models.CharField(max_length=178, blank=True)
+    type = models.CharField(max_length=10)
+    records = models.CharField(max_length=64000, blank=True)
```
nils-wisiol: Storing multiple values per field violates the first DB normal form

api/desecapi/models.py
```diff
@@ -227,3 +230,68 @@ def save(self, *args, **kwargs):
...
+        return val
+
+    def setter_ttl(self, val):
+        if val != self.ttl:
+            self._dirty = True
+
+        return val
+
+    def update_pdns(self):
+        from .serializers import RRsetSerializer
```
nils-wisiol: why not import globally?

api/desecapi/pdns.py
```diff
@@ -121,6 +126,40 @@ def delete_zone(name):
...
+    rrsets = get_zone(name)['rrsets']
+    rrsets = [rrset for rrset in rrsets \
+              if (subname == None or rrset['name'] == fullname) and (type == None or rrset['type'] == type) \
+        ]
+
+    # Make sure that if subname and type are given, we get one or none RRsets
+    assert subname == None or type == None or len(rrsets) <= 1, \
+            "pdns returned multiple RRsets of type %s for %s (zone: %s)" % (type, fullname, name)
+
+    return jq("[.[] | {name: .name, records: [.records[] | .content], ttl: .ttl, type: .type}]").transform(rrsets)
```
nils-wisiol: This interface breaks current `pdns.py` design pattern: up to now, all `pdns.py` interfaces are ("proper") python objects; `get_rrsets` and `set_rrsets` however both rely on serialized JSON objects. Unless there is good reason, this should be changed to python objects.

api/desecapi/pdns.py
```diff
@@ -121,6 +126,40 @@ def delete_zone(name):
...
+
+
+def set_rrsets(name, rrsets):
+    name = normalize_hostname(name)
+
+    data = jq('{"rrsets": [ .[] | { "name": .name, "type": .type, "ttl": .ttl, "changetype": "REPLACE", "records": [ .records[] | { "content": ., "disabled": false } ] } ] }').transform(rrsets)
```
nils-wisiol: `jq` programs are hard to read, I believe that [python list comprehensions](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) are a better alternative.

api/desecapi/serializers.py
```diff
@@ -1,6 +1,71 @@
...
+        return json.loads(obj)
+
+    def to_internal_value(self, data):
+        return json.dumps(data)
+
+
+class RecordsSerializer(JSONSerializer):
+    def to_internal_value(self, records):
+        if not all(isinstance(record, str) for record in records):
+            msg = 'Incorrect type. Expected a list of strings'
```
nils-wisiol: What if it's not even a list?

api/desecapi/serializers.py
```diff
@@ -1,6 +1,71 @@
...
+
+        # https://lists.isc.org/pipermail/bind-users/2008-April/070148.html
+        if not len(records) < 4092:
+            msg = 'Records too long. Must be less than 4092 characters, but was %d'
+            raise serializers.ValidationError(msg % len(records))
+
+        return super().to_internal_value(records)
+
+
+class FromContext(object):
```
nils-wisiol: add some doc here

api/desecapi/serializers.py
```diff
@@ -1,6 +1,71 @@
...
+
+    def __call__(self):
+        return self.value
+
+
+class RRsetSerializer(serializers.ModelSerializer):
+    # Disallow moving RRset to different zone
+    domain = serializers.SlugRelatedField(read_only=True, slug_field='name',
+                                          default=FromContext(
+                                              lambda context: Domain.objects.get(name=context['view'].kwargs['name'], owner=context['request'].user.pk)
```
nils-wisiol: line too long to properly review on GitHub; the viewer here imposes a 120ish char limit

api/desecapi/serializers.py
```diff
@@ -1,6 +1,71 @@
...
+                                          ))
+
+    # Disallow renaming RRset (RRset.update_pdns() does not cope with this)
+    subname = serializers.CharField(allow_blank=True, read_only=True,
+                                 default=FromContext(
+                                     lambda context: context['request'].data.get('subname', '').lower()
+                                 ))
+
+    # Disallow changing RRset type (RRset.update_pdns() does not cope with this)
+    type = serializers.CharField(read_only=True,
```
nils-wisiol: - shadows `type`
 - imposes a restriction that is due to implementation elsewhere. If possible, the layer that imposes the restriction should also handle validation of required prerequisites

api/desecapi/serializers.py
```diff
@@ -1,6 +1,71 @@
...
+                                 default=FromContext(
+                                     lambda context: context['request'].data['type'].upper()
+                                 ))
+
+    name = serializers.SerializerMethodField()
+    records = RecordsSerializer()
+
+    def get_name(self, obj):
+        return '.'.join(filter(None, [obj.subname, obj.domain.name])) + '.'
```
nils-wisiol: duplicate code, cf `pdns.py`

api/desecapi/views.py
```diff
@@ -120,6 +122,95 @@ class DomainDetailByName(DomainDetail):
...
+
+        # Replace by proper exception
```
nils-wisiol: add TODO

api/desecapi/views.py
```diff
@@ -120,6 +122,95 @@ class DomainDetailByName(DomainDetail):
...
+        if subname is not None:
+            return RRset.objects.filter(domain__name=name, domain__owner=self.request.user.pk, subname=subname)
+        elif type is not None:
+            return RRset.objects.filter(domain__name=name, domain__owner=self.request.user.pk, type=type)
+        else:
+            return RRset.objects.filter(domain__name=name, domain__owner=self.request.user.pk)
+
+    def create(self, request, *args, **kwargs):
+        if self.kwargs.get('subname') is not None or self.kwargs.get('type') is not None:
+            raise MethodNotAllowed(method=request.method)
```
nils-wisiol: Shouldn't this be `CONFLICT`? Method not allowed tells IMHO that this method is not allowed at this endpoint under any circumstances.

api/desecapi/views.py
```diff
@@ -120,6 +122,95 @@ class DomainDetailByName(DomainDetail):
...
+            ex = ValidationError(detail={"detail": str(e)})
+            ex.status_code = status.HTTP_409_CONFLICT
+            raise ex
+
+    def get(self, request, *args, **kwargs):
+        subname = self.kwargs['subname']
+        type = self.kwargs['type'].upper()
+
+        # TODO Do we need this functionality at all? We could just disregard the SOA record completely.
+        if type == 'SOA' and subname == '':
```
nils-wisiol: aggregate those SOA checks in one place, currently they are split across different layers

peterthomassen: This already has been consolidated; there is now one check in each of the views that handle an endpoint under `rrsets/`.

api/desecapi/views.py
peterthomassen: Error handling has already been fixed in a later commit

api/desecapi/views.py
peterthomassen: This already has been fixed in a later commit.

api/desecapi/migrations/0015_rrset.py
peterthomassen: fixed

api/desecapi/models.py
```diff
@@ -227,3 +230,68 @@ def save(self, *args, **kwargs):
...
         ordering = ('created',)
+
+
+class RRset(models.Model):
+    # TODO Do these two fields really make sense? Meaning is limited when deleting + recreating an RRset
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(null=True)
+    domain = models.ForeignKey(Domain, on_delete=models.CASCADE, related_name='rrsets')
+    subname = models.CharField(max_length=178, blank=True)
+    type = models.CharField(max_length=10)
```
peterthomassen: fixed

api/desecapi/pdns.py
```diff
@@ -121,6 +126,40 @@ def delete_zone(name):
...
+def get_zone(name):
+    """
+    Retrieves a JSON representation of the zone from pdns
+    """
+    r = _pdns_get('/zones/' + normalize_hostname(name))
+
+    return r.json()
+
+
+def get_rrsets(name, subname = None, type = None):
```
peterthomassen: Fixed the shadowing, and will fix whitespace style in upcoming commit

api/desecapi/permissions.py
peterthomassen: I don't think this is misleading.

1) The comment doesn't say that the permission is active without being called. It is still a "permission" in Django speech. Consequently, the parent class is a BasePermission (and not a "CallablePermissionCheck" or something).

2) This follows the same language as the other permission in that file that you implemented ;-)

api/desecapi/serializers.py
peterthomassen: In a later commit, this was changed to check `isinstance(records, str) or not all(isinstance(record, str) for record in records)` because inputting a string would pass the check in the way it was done previously. I believe this is safe now -- can you let me know your thoughts?

api/desecapi/views.py
peterthomassen: fixed

api/desecapi/views.py
peterthomassen: fixed

api/desecapi/views.py
peterthomassen: fixed

api/desecapi/models.py
peterthomassen: Our interface provides access to RRsets, not to individual resource records. Also internally in our model, individual resource records are not accessible. Therefore, each (JSON representation of an) RRset is atomic. The DBMS or even our application are not concerned with accessing sub-parts of the `records` field. We receive it as a whole from the user, and pass it on to pdns.

(On the way, we check whether the value is well-formed JSON. However, it is common to validate atomic values, such as domain names, phone numbers, ...)

Thus, in the context of what we're doing, we are storing one value here (i.e., the list of record values).

api/desecapi/models.py
peterthomassen: done

api/desecapi/models.py
peterthomassen: Because `serializers.py` globally imports `Domain` which is in `models.py`. If `models.py` now imports something from `serializers.py` globally, there will be an `import` loop going back and forth when loading either file.

I know this is ugly, but I don't know how to solve, except by breaking the loop by a local import. Any other ideas?

api/desecapi/models.py
peterthomassen: This is gone now (checks are done only on view-level; if you have model access, you can mess with the SOA record)

api/desecapi/serializers.py
peterthomassen: done

api/desecapi/serializers.py
peterthomassen: fixed

api/desecapi/models.py
nils-wisiol: I think the problem lies deeper within the organization of the JSON parsing and passing. You use `RRsetSerializer` to obtain data passed to `pdns.set_rrsets`, hence the import. The serializer here is needed as the data is stored serialized in the database -- a choice that I criticized in another comment on this PR. While I can see the point of doing this, the import issue here is a consequence of that: if the data was "properly" (in the sense of my other comment) stored, we didn't need any serializer here.

If you insist that storing serialized JSON is the way to go here, then I think `pdns.py` should provide an appropriate interface to deal with this; rather than the model deserializing the DB-stored values. (As you also brought up the pdns API as one reason to store the entire rrset serialized.)

That being said, I propose to store records not serialized, as this will save us the `jq` usage, the serializer import and parsing across the project on different layers (that were not intended to deal with (de)serialization).

api/desecapi/models.py
nils-wisiol: Please consider [this comment](https://github.com/desec-io/desec-stack/pull/50/files/e24690761137d31315045721ac1716b71fff7600#r123376800).

api/desecapi/pdns.py
nils-wisiol: This discussion is continued in [this comment](https://github.com/desec-io/desec-stack/pull/50/files/e24690761137d31315045721ac1716b71fff7600#r123376800).

api/desecapi/pdns.py
nils-wisiol: Related to [this comment](https://github.com/desec-io/desec-stack/pull/50/files/e24690761137d31315045721ac1716b71fff7600#r123376800).

api/desecapi/permissions.py
nils-wisiol: point taken

api/desecapi/serializers.py
nils-wisiol: The custom serializer here can be avoided, cf [this comment](https://github.com/desec-io/desec-stack/pull/50/files/e24690761137d31315045721ac1716b71fff7600#r123376800).

api/desecapi/serializers.py
peterthomassen: fixed

docs/domains.rst
```diff
@@ -0,0 +1,196 @@
...
+    String with an IPv4 address that will be written to the ``A`` RRset of the
+    zone apex, or ``null``.  If ``null``, the RRset is removed.
+
+    This was originally introduced to set an IPv4 address for deSEC's dynamic
+    DNS service dedyn.io.  However, it has some drawbacks (redundancy with
+    `Modifying an RRset`_ as well as inability to set multiple addresses).
+
+    *Do not rely on this field; it may be removed in the future.*
+
+``created``
```
nils-wisiol: remove this field from doc and spec

docs/domains.rst
```diff
@@ -0,0 +1,196 @@
...
+
+    Domain name.  Restrictions on what is a valid domain name apply on a
+    per-user basis.  The maximum length is 191.
+
+``owner``
+    :Access mode: read-only
+
+    Email address of the user owning the domain.
+
+``update``
```
nils-wisiol: remove this field from doc and spec

docs/domains.rst
```diff
@@ -0,0 +1,196 @@
...
+To create a new domain, issue a ``POST`` request to the ``/api/v1/domains/``
+endpoint, like this::
+
+    http POST \
+        https://desec.io/api/v1/domains/ \
+        Authorization:"Token {token}" \
+        name:='"example.com"'
+
+Only the ``name`` field is mandatory; ``arecord``, ``acme_challenge``, and
+``aaaarecord`` are optional.
```
nils-wisiol: optional and depricated

docs/domains.rst
```diff
@@ -0,0 +1,196 @@
...
+        Authorization:"Token {token}"
+
+This will return only one domain (i.e., the response is not a JSON array).
+
+If you own a domain with that name, the API responds with ``200 OK`` and
+returns the domain object in the reponse body.  Otherwise, the return status
+code is ``404 Not Found``.
+
+
+Modifying a Domain
```
nils-wisiol: add "(depricated)" to the headline

docs/domains.rst
```diff
@@ -0,0 +1,196 @@
...
+        Authorization:"Token {token}" \
+        acme_challenge:='""' arecord:='null'
+
+If the domain was updated successfully, the response status code is ``200 OK``
+and the updated domain object is returned in the response body.  In case of
+malformed request data such as syntactically invalid field contents for
+``arecord`` or ``aaaarecord``, ``400 Bad Request`` is returned.  If the domain
+does not exist or you don't own it, the status code is ``404 Not Found``.
+
+Note: Do not use the ``PUT`` method.  It currently behaves identically to the
```
nils-wisiol: should this PR remove `PUT`?

docs/endpoint-reference.rst
```diff
@@ -0,0 +1,39 @@
...
++================================================+============+=============================================+
+| ...\ ``/``                                     | ``GET``    | Retrieve all domains owned by you           |
+|                                                +------------+---------------------------------------------+
+|                                                | ``POST``   | Create a domain                             |
++------------------------------------------------+------------+---------------------------------------------+
+| ...\ ``/{domain}/``                            | ``GET``    | Retrieve a specific domain                  |
+|                                                +------------+---------------------------------------------+
+|                                                | ``PATCH``  | Modify a domain                             |
+|                                                +------------+---------------------------------------------+
+|                                                | ``PUT``    | Replace a domain                            |
```
nils-wisiol: remove if decided so (see previous comment)

docs/endpoint-reference.rst
```diff
@@ -0,0 +1,39 @@
...
+|                                                +------------+---------------------------------------------+
+|                                                | ``DELETE`` | Delete a domain                             |
++------------------------------------------------+------------+---------------------------------------------+
+| ...\ ``/{domain}/rrsets/``                     | ``GET``    | Retrieve all RRsets from zone               |
+|                                                +------------+---------------------------------------------+
+|                                                | ``POST``   | Create an RRset                             |
++------------------------------------------------+------------+---------------------------------------------+
+| ...\ ``/{domain}/rrsets/{type}/``              | ``GET``    | Retrieve all RRsets of a specific type      |
++------------------------------------------------+------------+---------------------------------------------+
+| ...\ ``/{domain}/rrsets/{subname}.../``        | ``GET``    | Retrieve all RRsets with a specific subname |
```
nils-wisiol: needs updating according to refactoring with queries

docs/rrsets.rst
```diff
@@ -0,0 +1,277 @@
...
+        "domain": "example.com",
+        "name": "www.example.com.",
+        "records": [
+            "127.0.0.1",
+            "127.0.0.2"
+        ],
+        "subname": "www",
+        "ttl": 3600,
+        "type": "A",
+        "updated": "2017-06-05T17:10:32.880217Z"
```
nils-wisiol: order JSON by semantics

docs/rrsets.rst
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+Field details:
+
+``created``
```
nils-wisiol: remove

docs/rrsets.rst
```diff
@@ -0,0 +1,277 @@
...
+
+    Subdomain string which, together with ``domain``, defines the RRset name.
+    Typical examples are ``www`` or ``_443._tcp``.  The maximum length is 178.
+
+``ttl``
+    :Access mode: read, write
+
+    TTL (time-to-live) value, which dictates for how long resolvers may cache
+    this RRset.  Only positive integer values are allowed.  Additional
+    restrictions may apply.
```
nils-wisiol: in seconds

docs/rrsets.rst
```diff
@@ -0,0 +1,277 @@
...
+    :Access mode: read, write-once (upon RRset creation)
+
+    RRset type (uppercase).  We support all `RRset types supported by
+    PowerDNS`_, with the exception of DNSSEC-related types (the backend
+    automagically takes care of setting those records properly).  You also
+    cannot access the ``SOA`` record, see `SOA caveat`_.
+
+.. _RRset types supported by PowerDNS: https://doc.powerdns.com/md/types/
+
+``updated``
```
nils-wisiol: remove

docs/rrsets.rst
```diff
@@ -0,0 +1,277 @@
...
+Retrieving RRsets of a Specific Type
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To retrieve an array of all RRsets from your zone that have a specific type
+(e.g. all ``A`` records, regardless of ``subname``), issue a ``GET`` request
+with the (uppercase) RRset type appended to the ``rrsets/`` endpoint, like
+this::
+
+    http GET \
+        https://desec.io/api/v1/domains/{domain}/rrsets/{type}/ \
```
nils-wisiol: change according to spec changes (query)

docs/rrsets.rst
```diff
@@ -0,0 +1,277 @@
...
+Retrieving RRsets with a Specific Subname
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To retrieve an array of all RRsets from your zone that have a specific subname
+(e.g. all records in the ``www`` subdomain, regardless of their type), issue a
+``GET`` request with the ``subname`` information appended to the ``rrsets/``
+endpoint, like this::
+
+    http GET \
+        https://desec.io/api/v1/domains/{domain}/rrsets/{subname}.../ \
```
nils-wisiol: change according to spec changes (query)

docs/rrsets.rst
```diff
@@ -0,0 +1,277 @@
...
+Retrieving a Specific RRset
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To retrieve an RRsets with a specific name and type from your zone (e.g. the
+``A`` record for the ``www`` subdomain), issue a ``GET`` request with the
+``subname`` information and the type appended to the ``rrsets/`` endpoint,
+like this::
+
+    http GET \
+        https://desec.io/api/v1/domains/{domain}/rrsets/{subname}.../{type}/ \
```
nils-wisiol: change according to spec changes (query)

docs/rrsets.rst
```diff
@@ -0,0 +1,277 @@
...
+.. _`SOA caveat`:
+
+- The ``SOA`` record cannot be read or written through this interface.  When
+  attempting to create, modify or otherwise access an ``SOA`` record, ``403
+  Forbidden`` is returned.
+
+  The rationale behind this is that the content of the ``SOA`` record is
+  entirely determined by the DNS operator, and users should not have to bother
+  with this kind of metadata.  Upon zone changes, the backend automatically
+  takes care of updating the ``SOA`` record accordingly.
```
nils-wisiol: maybe add: if you're interested in the SOA, retrieve it via DNS queries

api/desecapi/views.py
```diff
@@ -184,8 +172,8 @@ def get_queryset(self):
         subname = self.kwargs.get('subname')
         type = self.kwargs.get('type')
 
-        # Replace by proper exception
-        assert(subname is None or type is None)
+        if subname is not None and type is not None:
+            raise ValueError("Invalid combination of arguments")
```
nils-wisiol: this could be removed with query-access to rrsets

api/desecapi/views.py
```diff
@@ -198,19 +186,18 @@ def create(self, request, *args, **kwargs):
         if self.kwargs.get('subname') is not None or self.kwargs.get('type') is not None:
             raise MethodNotAllowed(method=request.method)
 
+        type = request.data.get('type', '').upper()
```
nils-wisiol: shadows `type`

api/desecapi/views.py
```diff
@@ -127,52 +130,37 @@ class RRsetDetail(generics.RetrieveUpdateDestroyAPIView):
...
+    def get_queryset(self):
+        name = self.kwargs['name']
         subname = self.kwargs['subname']
         type = self.kwargs['type'].upper()
 
-        # TODO Do we need this functionality at all? We could just disregard the SOA record completely.
-        if type == 'SOA' and subname == '':
-            name = self.kwargs['name']
+        if type == 'SOA':
+            raise PermissionDenied("You cannot tamper with the SOA record.")
```
nils-wisiol: Should we implement this as a Django permission? The `DoNotQuerySOARecordPermission` or something?

api/desecapi/views.py
```diff
@@ -198,19 +186,18 @@ def create(self, request, *args, **kwargs):
         if self.kwargs.get('subname') is not None or self.kwargs.get('type') is not None:
             raise MethodNotAllowed(method=request.method)
 
+        type = request.data.get('type', '').upper()
+        if type == 'SOA':
+            raise PermissionDenied("You cannot tamper with the SOA record.")
```
nils-wisiol: This could use the above-mentioned permission class

api/desecapi/views.py
nils-wisiol: Should we implement this as a Django permission? The `DoNotQuerySOARecordPermission` or something?

nils-wisiol:
"further improvements (see details)" is not an allowed commit message

nils-wisiol:
check for TODOs before merging

nils-wisiol:
Should we make the beta API available under `/api/vbeta/` or something?

api/desecapi/views.py
```diff
@@ -175,18 +175,32 @@ def get_queryset(self):
         if subname is not None and type is not None:
             raise ValueError("Invalid combination of arguments")
 
+        if type == 'SOA':
+            raise PermissionDenied("You cannot tamper with the SOA record.")
```
nils-wisiol: use Django permission

api/desecapi/views.py
```diff
@@ -163,7 +163,7 @@ def update(self, request, *args, **kwargs):
         return super().update(request, *args, **kwargs)
 
 
-class RRsetsDetail(generics.ListCreateAPIView):
+class RRsetsDetailSpecific(generics.ListAPIView):
```
nils-wisiol: replace by query endpoint

api/desecapi/management/commands/sync-from-pdns.py
```diff
@@ -0,0 +1,51 @@
...
+
+        if options['domain-name']:
+            domains = domains.filter(name__in=options['domain-name'])
+
+        for domain in domains:
+            print(domain.name + ' ', end='', flush=True)
+
+            try:
+                rrsets_pdns = pdns.get_rrsets(domain.name)
+                rrsets_pdns = jq('map(select( .type != "SOA" ))').transform(rrsets_pdns)
```
peterthomassen: fixed

api/desecapi/management/commands/sync-from-pdns.py
```diff
@@ -0,0 +1,51 @@
...
+
+                rrsets = []
+                for rrset_pdns in rrsets_pdns:
+                    records = json.dumps(rrset_pdns['records'])
+                    ttl = rrset_pdns['ttl']
+                    type = rrset_pdns['type']
```
peterthomassen: fixed

api/desecapi/pdns.py
```diff
@@ -62,7 +62,7 @@ def _pdns_patch(url, body):
 
 def _pdns_get(url):
     r = requests.get(settings.NSLORD_PDNS_API + url, headers=headers_nslord)
-    if r.status_code < 200 or r.status_code >= 500:
+    if r.status_code < 200 or r.status_code >= 400:
```
peterthomassen: Sounds like a good idea, but let's treat it separately

api/desecapi/serializers.py
```diff
@@ -36,13 +37,25 @@ def set_context(self, serializer_field):
...
+        return domain
+
 
 class RRsetSerializer(serializers.ModelSerializer):
     # Disallow moving RRset to different zone
     domain = serializers.SlugRelatedField(read_only=True, slug_field='name',
                                           default=FromContext(
-                                              lambda context: Domain.objects.get(name=context['view'].kwargs['name'], owner=context['request'].user.pk)
-                                          ))
+                                              lambda context: FromContext.get_domain(context)
```
peterthomassen: The default value is because if you `POST` on `rrsets/example.com/` to create an RRset, you need to extract `example.com` from the URL and use it to set the `domain` field of the RRset. So, the field is read-only (the user cannot supply it), but if empty (on creation), we fill it with the required value.

How do you suggest solving the layer architecture issue? If the serializer needs to transform the domain name into a domain id, it seems unavoidable for the serializer to invoke a database query.

docs/domains.rst
peterthomassen: done

docs/domains.rst
peterthomassen: done

docs/domains.rst
peterthomassen: done

docs/domains.rst
peterthomassen: done

docs/domains.rst
peterthomassen: removed from docs

docs/endpoint-reference.rst
peterthomassen: done

docs/endpoint-reference.rst
peterthomassen: will update in the commit that introduces the queries

docs/rrsets.rst
peterthomassen: done

docs/rrsets.rst
peterthomassen: done

docs/rrsets.rst
peterthomassen: done

docs/rrsets.rst
peterthomassen: will update in the commit that introduces the queries

docs/rrsets.rst
peterthomassen: will update in the commit that introduces the queries

docs/rrsets.rst
peterthomassen: will update in the commit that introduces the queries

peterthomassen:
> Should we make the beta API available under `/api/vbeta/` or something?

I think it's better if people don't have to change their stuff once beta ends. I don't think API changes are likely during beta (once the query mechanism is implemented)

docs/rrsets.rst
peterthomassen: done

api/desecapi/views.py
peterthomassen: `type` shadowing was fixed in later commit

api/desecapi/models.py
peterthomassen: I refactored the serialization logic for the use case of the API talking to pdns, see https://github.com/desec-io/desec-stack/commit/67f0bb19f508967f318d977ad72be2a8f0afff62. Everything is now done in `pdns.get_rrsets()` and `pdns.set_rrsets()` where `GenericRRsetSerializer` is used.

`GenericRRsetSerializer` is a `ModelSerializer`-based serializer. (It is extended by `RRsetSerializer` which is used by the frontend views.) The serializer transforms between `RRset` objects and a Python dictionary whose structure is driven by `ModelSerializer`.

`pdns.get_rrsets()` and `pdns.set_rrsets()` still have some list comprehension logic which is used to transform between the structure received from pdns and the Python dictionary produced by `GenericRRsetSerializer`.

At this point, we face two problems:

1. The list comprehension could be removed if the pdns JSON structure could be reproduced based on a `ModelSerializer` (such as `GenericRRsetSerializer`). However, it is not straightforward to write such a serializer, due to the (in our pdns usecase, constant but mandatory) extra fields in the pdns JSON structure (e.g. `'changetype': 'REPLACE'`). Note that this has nothing to do with whether `records` (list of strings) is deserialized into separate record objects or not.

   I am not sure if the extra work of making the serializer pdns-compatible is justified by the little gain of getting rid of the list comprehension. (I would take a different stand if this functionality was reused elsewhere in the code.)

2. Even with all transformations encapsulated in `pdns.get_rrsets()`, `pdns.set_rrsets()`, and `serializers.py`, a circular import issue occurs: `RRset.save()` calls `RRset.update_pdns()` which calls functions from `pdns.py`, so we need to import `pdns.py`. In `pdns.py`, we use `GenericRRsetSerializer`. Even if Problem 1 was solved, we would still need to call some serializer for `RRset` here. In that serializer's `Meta.model` attribute, the `RRset` model is referenced (as long as the serializer is a `ModelSerializer`), closing the loop.

   Summary: If within `RRset`, we call something that serializes an `RRset` object, and if the serializer used in that process is a `ModelSerializer`, then we have an import loop.

   Therefore, `pdns.get_rrsets()` and `pdns.set_rrsets()` have local imports at the moment, and I don't see a way of getting rid of them (except not using `ModelSerializer`). On the other hand, I don't see why a local import would be a problem. (The loop is not a logical inconsistency; it has merely to do with the fact that global import statements are not handled very smartly.)

api/desecapi/management/commands/sync-from-pdns.py
```diff
@@ -0,0 +1,51 @@
...
+        for domain in domains:
+            print(domain.name + ' ', end='', flush=True)
+
+            try:
+                rrsets_pdns = pdns.get_rrsets(domain.name)
+                rrsets_pdns = jq('map(select( .type != "SOA" ))').transform(rrsets_pdns)
+
+                rrsets = []
+                for rrset_pdns in rrsets_pdns:
+                    records = json.dumps(rrset_pdns['records'])
```
peterthomassen: fixed

api/desecapi/views.py
```diff
@@ -200,15 +200,25 @@ def create(self, request, *args, **kwargs):
 
         try:
             return super().create(request, *args, **kwargs)
+        except Domain.DoesNotExist:
+            raise Http404
         except django.core.exceptions.ValidationError as e:
             ex = ValidationError(detail={"detail": str(e)})
             ex.status_code = status.HTTP_409_CONFLICT
             raise ex
         # TODO improve error handling
```
peterthomassen: this was fixed in a later commit

docs/rrsets.rst
peterthomassen: done

api/desecapi/views.py
peterthomassen: done

api/desecapi/views.py
peterthomassen: done

api/desecapi/views.py
peterthomassen: This is more difficult than it seems. Explanation:

- Object permission operates on objects (not on queries etc.), so they only apply to objects that have actually been saved to the database. However, RRsets of restricted type are not expected to be in the database.
- I considered `type` field validation in the serializer as an alternative. However, serializer-based validation only works for user input (on `POST`); if we want to restrict `GET` requests for restricted RRset types (even if they are not present in the database), the serializer will not be involved as the restriction is before the serialization step.
- Thus, of the three "can't tamper with restricted record types" exceptions, only the one in the view's `create()` method (dealing with `POST`) can be replaced by a field validator in the serializer; the other two exceptions cannot be moved.
- An alternative _may_ be model permissions (instead of object permissions), see http://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissions. Conceptually, however, this sort of implies that there are users which have "restricted types access permission" and others that don't. AFAIU, this would also create an extra entry in the `auth_permission` database table. I'm not sure this solution is desirable.

All in all, I favor doing nothing about this at this point (not even moving the single check for which that is possible). This gives the advantage of having all checks in more or less one place (namely, in the `RRset`-related views).

api/desecapi/views.py
peterthomassen: regarding the permission, see https://github.com/desec-io/desec-stack/pull/50#discussion_r124585051

api/desecapi/views.py
peterthomassen: see https://github.com/desec-io/desec-stack/pull/50#discussion_r124585051

api/desecapi/views.py
peterthomassen: see https://github.com/desec-io/desec-stack/pull/50#discussion_r124585051

api/desecapi/serializers.py
peterthomassen: - does not shadow (referenced as `self.type`)
- moved write restriction to model layer, https://github.com/desec-io/desec-stack/commit/16ef52342c657c0b1dfdb11099ccb4c78b7c4eae

peterthomassen:
I addressed all your comments with the commits from June 27 on. Next round of review? (thanks for the first!)

docs/rrsets.rst
```diff
@@ -246,32 +248,114 @@ Upon success or if the RRset did not exist in the first place, the response
...
 
 - All operations are performed on RRsets, not on the individual Resource
   Records.
 
+- The TTL (time-to-live: time for which resolvers may cache DNS information)
+  is a property of an RRset (and not of a record).  Thus, all records in an
+  RRset share the record type and also the TTL.  (This is actually a
+  requirement of the DNS specification and not an API design choice.)
+
+- We have not done extensive testing for reverse DNS, but hings should work in
```
nils-wisiol: "things"

docs/rrsets.rst
```diff
@@ -246,32 +248,114 @@ Upon success or if the RRset did not exist in the first place, the response
...
-- The TTL (time-to-live: time for which resolvers may cache DNS information)
-  is a property of an RRset (and not of a record).  Thus, all records in an
-  RRset share the record type and also the TTL.  (This is actually a
-  requirement of the DNS specification and not an API design choice.)
+Record types with priority field
+    The deSEC DNS API does not explicitly support priority fields (as used for
+    ``MX`` or ``SRV`` records and the like).
+
+    Instead, the priority is expected to be specified at the beginning of the
+    record content, separated from the rest of it by whitespace.
```
nils-wisiol: whitespace as in \t, \r, \n, and so on?

docs/rrsets.rst
```diff
@@ -246,32 +248,114 @@ Upon success or if the RRset did not exist in the first place, the response
...
+    mail.a4a.de.``
+
+``NS`` record
+    The use of wildcard RRsets (with one component of ``subname`` being equal
+    to ``*``) of type ``NS`` is **discouraged**.  This is because the behavior
+    of wildcard ``NS`` records in conjunction with DNSSEC is undefined, per
+    RFC 4592, Sec. 4.2.
+
+``TXT`` record
+    The contents of the ``TXT`` record must be encloded in double quotes.
```
nils-wisiol: "enclosed"

api/desecapi/pdns.py
```diff
@@ -142,15 +140,25 @@ def get_rrsets(domain, subname = None, type_ = None):
...
 
 
 def set_rrsets(domain, rrsets):
-    data = jq('{"rrsets": [ .[] | { "name": .name, "type": .type, "ttl": .ttl, "changetype": "REPLACE", "records": [ .records[] | { "content": ., "disabled": false } ] } ] }').transform(rrsets)
+    data = {'rrsets':
+        [{'name': rrset['name'], 'type': rrset['type'], 'ttl': rrset['ttl'],
+          'changetype': 'REPLACE',
+          'records': [{'content': record, 'disabled': False}
+                      for record in rrset['records']]
+          } for rrset in rrsets]
```
nils-wisiol: improve formatting for the list comprehension

api/desecapi/pdns.py
```diff
@@ -142,15 +140,25 @@ def get_rrsets(domain, subname = None, type_ = None):
...
-              if (subname == None or rrset['name'] == fullname) and (type_ == None or rrset['type'] == type_) \
-        ]
-
-    return jq("[.[] | {name: .name, records: [.records[] | .content], ttl: .ttl, type: .type}]").transform(rrsets)
+    rrsets = [
+        {'name': rrset['name'],
+         'records': [record['content'] for record in rrset['records']],
+         'ttl': rrset['ttl'], 'type': rrset['type']
+         } for rrset in rrsets
+        if (subname == None or rrset['name'] == fullname)
```
nils-wisiol: improve formatting for the list comprehension

docs/rrsets.rst
peterthomassen: shall fix

docs/rrsets.rst
peterthomassen: shall fix

api/desecapi/views.py
```diff
@@ -120,6 +125,90 @@ class DomainDetailByName(DomainDetail):
...
+
+        try:
+            return super().update(request, *args, **kwargs)
+        except django.core.exceptions.ValidationError as e:
+            ex = ValidationError(detail=e.message_dict)
+            ex.status_code = status.HTTP_409_CONFLICT
+            raise ex
+
+
+class RRsetsDetail(generics.ListCreateAPIView):
```
nils-wisiol: rename to `RRsetList`

api/desecapi/views.py
```diff
@@ -120,6 +125,90 @@ class DomainDetailByName(DomainDetail):
...
+
+    def get_queryset(self):
+        rrsets = RRset.objects.filter(domain__owner=self.request.user.pk,
+                                      domain__name=self.kwargs['name'])
+
+        for filter_field in ('subname', 'type'):
+            value = self.request.query_params.get(filter_field)
+
+            if value is not None:
+                # This may be more cleanly implemented using object level permissions.
```
nils-wisiol: mark with TODO

api/desecapi/views.py
```diff
@@ -120,6 +125,90 @@ class DomainDetailByName(DomainDetail):
...
+        except Http404:
+            pass
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def get_queryset(self):
+        name = self.kwargs['name']
+        subname = self.kwargs['subname'].replace('=2F', '/')
+        type_ = self.kwargs['type']
+
+        # This may be more cleanly implemented using object level permissions.
```
nils-wisiol: mark with TODO

api/desecapi/views.py
```diff
@@ -120,6 +125,90 @@ class DomainDetailByName(DomainDetail):
...
+                if filter_field == 'type' and value in RRsetDetail.restricted_types:
+                    raise PermissionDenied("You cannot tamper with the %s RRset." % value)
+
+                rrsets = rrsets.filter(**{'%s__exact' % filter_field: value})
+
+        return rrsets
+
+    def create(self, request, *args, **kwargs):
+        if self.kwargs.get('subname') is not None or self.kwargs.get('type') is not None:
+            raise MethodNotAllowed(method=request.method)
```
nils-wisiol: should be bad request

api/desecapi/pdns.py
peterthomassen: shall do

docs/rrsets.rst
peterthomassen: shall write "... rest of it by a space"

api/desecapi/models.py
peterthomassen: For the record, discussed in person and resolved that there is no good way around this, so we'll keep this as is.

api/desecapi/views.py
```diff
@@ -120,6 +125,90 @@ class DomainDetailByName(DomainDetail):
...
+
+        type_ = request.data.get('type', '')
+        if type_ in RRsetDetail.restricted_types:
+            raise PermissionDenied("You cannot tamper with the %s RRset." % type_)
```
nils-wisiol: rephrase: manually changing the %s resource record set is prohibited.

api/desecapi/models.py
```diff
@@ -227,3 +235,98 @@ def save(self, *args, **kwargs):
...
+                              code='invalid',
+                              params={'value': value})
+
+
+class RRset(models.Model, mixins.SetterMixin):
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(null=True)
+    domain = models.ForeignKey(Domain, on_delete=models.CASCADE, related_name='rrsets')
+    subname = models.CharField(max_length=178, blank=True)
+    type = models.CharField(max_length=10, validators=[validate_upper])
```
peterthomassen: no, is always referenced as a member of class or instance

api/desecapi/views.py
peterthomassen: shall do

api/desecapi/views.py
peterthomassen: This code was no-op and left over from the past. I'll remove it

api/desecapi/serializers.py
peterthomassen: fixing via `perform_create()`

api/desecapi/views.py
peterthomassen: This is harder than it seems, see https://github.com/desec-io/desec-stack/pull/50#discussion_r124585051
Bottom line: Don't use permissions here. I removed the comment.

api/desecapi/views.py
peterthomassen: This is harder than it seems, see https://github.com/desec-io/desec-stack/pull/50#discussion_r124585051
Bottom line: Don't use permissions here. I removed the comment.

api/desecapi/views.py
peterthomassen: I'd like the same message for denying access to read and to write (i.e. `GET .../rrsets/?type=SOA`, `GET .../rrsets/.../SOA/`, and `PATCH .../rrsets/.../SOA/`). Your proposal doesn't fit all cases, and I can't come up with something better than tamper ...

peterthomassen:
I addressed all your comments and believe that this review is complete now.

api/desecapi/views.py
nils-wisiol: tamper IMHO does not include just looking at things, i.e. read. "You may not access"?

peterthomassen:
fixed the wording issue, using "tinker" now. See https://en.wiktionary.org/wiki/tinker#Verb ;-)

PR #70, opened 2017-09-08
integrate www and autocert

nils-wisiol:
This PR moves all files from desec-www into desec-stack and updates the required files (travis, README).

It then adds capability to www to autogenerate self-signed certificates on startup, if no certificate files were found in the folder specified in `.env`.

Those changes are in preperation to run desec-stack on travis for testing.


www/Dockerfile
```diff
@@ -1,11 +1,9 @@
-FROM nginx:1.11-alpine
-
-# Add bash for our entrypoint script
-RUN apk add --no-cache bash
+FROM nginx:1.11
```
peterthomassen: Breaks HTTP2, see https://github.com/desec-io/desec-www/commit/a892c415b1f1dccd14883c7ff3840c347a7440f0

www/README.md
```diff
@@ -1,57 +1,21 @@
...
-- `checkipv4.dedyn.${DOMAIN}` (return IP address)
-- `checkipv6.dedyn.${DOMAIN}` (return IP address)
-- `update.dedyn.${DOMAIN}` (update IP address)
-- `update6.dedyn.${DOMAIN}` (update IP address)
-
-It may contact the following hostnames:
-
-- `api` for API requests
+Various SSL certificates for the external hostnames. For the following hostnames, put certificate and private key files into the certificate directory specified in your `.env` file. Name the files ${NAME}.cer and ${NAME}.key respectively.
+   - `desec.${DESECSTACK_DOMAIN}` (e.g. desec.io)
```
peterthomassen: `... where ${NAME} is one of:`

www/README.md
```diff
@@ -1,57 +1,21 @@
...
-   - `MAIN` (e.g. desec.io)
-   - `www`
-   - `checkip.dedyn`
-   - `checkipv4.dedyn`
-   - `checkipv6.dedyn`
-   - `update.dedyn`
-   - `update6.dedyn`
-   - `dedyn`
-   - `www.dedyn`
-2. DH Parameter file (create this with `openssl dhparam -out dhparam.pem 2048`)
```
peterthomassen: Is this not needed anymore? I can't find this being done automatically in this commit.

www/autocert.sh
```diff
@@ -0,0 +1,39 @@
...
+    dedyn.${DESECSTACK_DOMAIN} \
+    desec.${DESECSTACK_DOMAIN} \
+    update6.dedyn.${DESECSTACK_DOMAIN} \
+    update.dedyn.${DESECSTACK_DOMAIN} \
+    www.dedyn.${DESECSTACK_DOMAIN} \
+    www.desec.${DESECSTACK_DOMAIN} \
+    ; do
+
+    if [ ! -f ${DOMAIN}.cer ] || [ ! -f ${DOMAIN}.key ] ; then
+        rm -f ${DOMAIN}.cer ${DOMAIN}.key
```
peterthomassen: I don't think we should delete any files that exist. It's clear that both files need to be absent for us to generate a new set of files. However, if only one is absent, it may be due to a filename typo, and we'd end up deleting the user's key or certificate file.

If only one file is missing, the most likely scenario is that something is wrong, so my proposal is to error out.

www/conf/nginx.conf
```diff
@@ -35,10 +35,10 @@ http {
     # rate limits are exceeded when the 'leaky bucket' is full
 
     # set up one bucket per remote ip for general purpose
-    limit_req_zone $binary_remote_addr zone=perip-general:100m rate=10r/s;
+    limit_req_zone $binary_remote_addr zone=perip-general:100m rate=100000r/s;
```
peterthomassen: This looks like is was for testing only, right? Is is in this PR intentionally (and if so, why)?

www/conf/nginx.conf
```diff
@@ -35,10 +35,10 @@ http {
...
 
     # set up one bucket per remote ip for (costly) API access
-    limit_req_zone $binary_remote_addr zone=perip-api:100m rate=30r/m;
+    limit_req_zone $binary_remote_addr zone=perip-api:100m rate=300000r/m;
```
peterthomassen: same

www/entrypoint.sh
```diff
@@ -1,4 +1,22 @@
-#!/bin/bash
+#!/bin/bash -e
+
+# generate certs
+mkdir -p /autocert/
+./autocert.sh /autocert/
+
+# link certs
+if [ ! -f /etc/ssl/private/desec.${DESECSTACK_DOMAIN}.cer ] ||
+   [ ! -f /etc/ssl/private/desec.${DESECSTACK_DOMAIN}.key ] ; then
```
peterthomassen: What if `desec.${DESECSTACK_DOMAIN}.{cer,key}` are there, but not one of the other certificates? (Why do these two files have a special role?)

More generally, I wondered why we are doing the `$CERT_PATH` magic instead of just making sure that `/etc/ssl/private/` has a complete set of key+certificate files? The simplest approach would be to have `autocert.sh` loop over the set of domain names, and generate any missing file pairs in `/etc/ssl/private/`.

www/autocert.sh
```diff
@@ -0,0 +1,39 @@
...
+        cert ${DOMAIN} "${DOMAIN}"
```
peterthomassen: Why is this a function at all, and why does it have to identical arguments?

www/Dockerfile
nils-wisiol: Then we can just undo this change and install openssl :+1:

www/README.md
nils-wisiol: This was dead documentation, i.e. the configuration didn't behave that way. Any dhparam.pem file will be ignored (before and after this PR)

nils-wisiol: [This appears to be a non-issue](https://www.ssllabs.com/ssltest/analyze.html?d=desec.io&s=2a01%3a4f8%3a10a%3a1044%3adeec%3a642%3aac10%3a80&latest), as our configuration doesn't support cipher suits relying on dh parameters.

www/README.md
nils-wisiol: fixed

www/entrypoint.sh
nils-wisiol: The `$CERT_PATH` magic is due to the fact that the certificate volume is mounted read-only (a feature that I wanted to keep).

nils-wisiol: That is also the reason we can't mix and match user-provided and auto-generated certificates without an irresponsibly high effort.

www/autocert.sh
nils-wisiol: Files will now be generated into an folder that is guaranteed to be empty before, so the checks are obsolete.

www/autocert.sh
nils-wisiol: Dropped function definition and simplified code structure.

www/entrypoint.sh
nils-wisiol: I kept the fact that we (due to read-only) cannot really mix auto-generated and user-provided certificates (for now?). I rewrote the condition in a way that all required files are checked, if a single one is missing, we fall back to autogeneration. The user is provided with a list of files that are missing in the certificate folder.

nils-wisiol:
I resolved all comments; please double-check on

- a lot of certificate autogeneration code that was basically re-written 
- the dh params, that are said to be unnecessary (by SSL Labs)

peterthomassen:
Looks good to me. Tested, and works. Let's talk about the DH parameters in person and merge afterwards.

peterthomassen:
oh, I forgot -- this also needs some squashing and a rebase (master has changed).

nils-wisiol:
> Deploy (Ephemeral) Elliptic-Curve Diffie-Hellman (ECDHE). Elliptic-Curve Diffie-Hellman (ECDH) key exchange avoids all known feasible cryptanalytic attacks, and modern web browsers now prefer ECDHE over the original, finite field, Diffie-Hellman. The discrete log algorithms we used to attack standard Diffie-Hellman groups do not gain as strong of an advantage from precomputation, and individual servers do not need to generate unique elliptic curves.

From the [weakdh.org](https://weakdh.org/sysadmin.html), the logjam inventors webpage.

IMHO, we do not need DH params.

nils-wisiol:
rebased and squashed, imho this is ready to merge

PR #80, opened 2017-12-04
20171129 fix race condition

peterthomassen:
This PR

- eliminates the race condition that occurs when users send multiple dynDNS IP update requests, by locking RRsets for update (previously, RRsets with same type/subname would have been attempted to be inserted twice, violating uniqueness)
- minimizes pdns interaction when records have not changed

It also contains a minor docs clarification (based on user feedback).

### DIFF
--- a/api/desecapi/models.py
+++ b/api/desecapi/models.py
@@ -183,18 +183,54 @@ class Domain(models.Model, mixins.SetterMixin):
 
     @transaction.atomic
     def write_rrsets(self, rrsets):
-        q = models.Q()
-        for rrset, rrs in rrsets.items():
-            for rr in rrs:
-                if rr.rrset is not rrset:
-                    raise ValueError('RR has wrong parent RRset')
-            q |= models.Q(subname=rrset.subname) & models.Q(type=rrset.type)
-        RRset.objects.filter(domain=self).filter(q).delete()
+        # Always-false Q object: https://stackoverflow.com/a/35894246/6867099
+        rrsets_index = {}
+        q_update = models.Q(pk__isnull=True)
+        q_delete = models.Q(pk__isnull=True)
 
-        RRset.objects.bulk_create([rrset for (rrset, rrs) in rrsets.items() if rrs])
+        # Determine which RRsets need to be updated or deleted
+        for rrset, rrs in rrsets.items():
+            if rrset.domain is not self:
+                raise ValueError('RRset has wrong domain')
+            if (rrset.subname, rrset.type) in rrsets_index:
+                raise ValueError('RRset repeated with same subname and type')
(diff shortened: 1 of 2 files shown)


api/desecapi/models.py
```diff
@@ -183,18 +183,54 @@ def sync_from_pdns(self):
 
     @transaction.atomic
     def write_rrsets(self, rrsets):
-        q = models.Q()
+        # Always-false Q object: https://stackoverflow.com/a/35894246/6867099
+        rrsets_index = {}
```
nils-wisiol: This variable is used to validate that no two given rrsets have the same subname and type. Is here the right place to do such validation? I believe not. Also, the database would deny writing such data. Hence, I'd prefer to move the validation somewhere else (serializer?) or drop it entirely.

api/desecapi/models.py
```diff
@@ -183,18 +183,54 @@ def sync_from_pdns(self):
...
+                                                     type=rrset.type,
+                                                     ttl=rrset.ttl)
+            rrsets[rrset] = [RR(rrset=rrset, content=rr.content) for rr in rrs]
 
-        RRset.objects.bulk_create([rrset for (rrset, rrs) in rrsets.items() if rrs])
         RR.objects.bulk_create([rr for rrs in rrsets.values() for rr in rrs])
 
-        if not self.owner.captcha_required:
+        # Send changed RRsets to pdns
+        if rrsets and not self.owner.captcha_required:
```
nils-wisiol: Don't we need to tell pdns if we got an empty rrsets array?

api/desecapi/models.py
peterthomassen: The variable was actually introduced to do a reverse lookup of the contents in the `rrsets` argument, by subname and type. It is used in https://github.com/desec-io/desec-stack/pull/80/files#diff-d93a995458c0f3a31fb831c4f43b5d27R216.

Since it was there, I figured that another validation would not hurt. Furthermore, `write_rrsets()` is a generic method of `Domain`, and not necessarily called through a serializer. In fact, it is used by the dynDNS view which is just an `APIview` without a `ModelSerializer` involved. So, there is currently no other place to put this validation. What do you think?

api/desecapi/models.py
peterthomassen: Yes, and we do, but the check here is whether there is anything to do at all.

Background: The user supplies a few RRsets in `rrsets`, and then we go and check if those which are empty exist at all (in which case we need to tell pdns to delete them), and if those which are not empty have changed. For non-existing empty RRsets and unchanged ones, nothing needs to happen, so we remove them from the `rrsets` variable, see lines 224 and 236. We may end up with an empty todo list, i.e. an empty `rrsets` dict, in which case we don't need to talk to pdns.

PR #90, opened 2018-01-31
20171211 misc

peterthomassen:
This PR fixes a number of issues, including the bug where unlocking a user would overwrite a domain that is managed in pdns exclusively (without using our API).

I think the items are pretty much clear based on the commit titles. If anything is unclear, let's talk! :-)


api/desecapi/models.py
```diff
@@ -216,80 +216,87 @@ def write_rrsets(self, datas):
...
+        # Base queryset for all RRset of the current domain
+        rrset_qs = RRset.objects.filter(domain=self)
+
+        # Set to check RRset uniqueness
+        rrsets_seen = set()
+
+        # To-do list for non-empty RRsets, indexed by (subname, type)
+        rrsets_meaty_todo = {}
+
+        # RRsets to send to pdns
```
nils-wisiol: better: dictionary of sets of RRs to send to pdns, indexed by their rrsets objects

api/desecapi/models.py
```diff
@@ -216,80 +216,87 @@ def write_rrsets(self, datas):
...
-            pdns.set_rrsets(self, rrsets)
+                rrsets_to_write[rrset] = []
+                q_empty |= q
+
+        # Construct querysets representing RRsets that do (not) have RR
+        # contents and lock them
+        qs_meaty = rrset_qs.filter(q_meaty).select_for_update()
+        qs_empty = rrset_qs.filter(q_empty).select_for_update()
+
+        # For existing RRsets, execute TTL updates and mark for RR update
```
nils-wisiol: "and/or"

api/desecapi/models.py
```diff
@@ -216,80 +216,87 @@ def write_rrsets(self, datas):
...
+            if changed_ttl:
+                rrset.ttl = rrset_temp.ttl
+                rrset.save()
+            if changed_ttl or changed_rrs:
+                rrsets_to_write[rrset] = [RR(rrset=rrset, content=rr_content)
+                                          for rr_content in rrs_temp]
+            if not changed_rrs:
+                rrsets_same_rrs.append(rrset)
+
+        # At this point, rrsets_meaty_todo points to new RRsets only. Let's
```
nils-wisiol: "new, non-empty RRsets only"

api/desecapi/models.py
```diff
@@ -216,80 +216,87 @@ def write_rrsets(self, datas):
...
+                q_empty |= q
+
+        # Construct querysets representing RRsets that do (not) have RR
+        # contents and lock them
+        qs_meaty = rrset_qs.filter(q_meaty).select_for_update()
+        qs_empty = rrset_qs.filter(q_empty).select_for_update()
+
+        # For existing RRsets, execute TTL updates and mark for RR update
+        rrsets_same_rrs = []
+        for rrset in qs_meaty.all():
```
nils-wisiol: Can this loop also include the "new, non-empty" case from below (line 280ff)? Then the "todo-list pattern" makes full sense. ("for item in todo")

api/desecapi/models.py
```diff
@@ -216,80 +216,87 @@ def write_rrsets(self, datas):
...
+
+        # Delete empty RRsets
+        qs_empty.delete()
+
+        # Update contents of modified RRsets
+        RR.objects.filter(rrset__in=qs_meaty).exclude(rrset__in=rrsets_same_rrs).delete()
+        RR.objects.bulk_create([rr
+                                for (rrset, rrs) in rrsets_to_write.items()
+                                for rr in rrs
+                                if rrset not in rrsets_same_rrs])
```
nils-wisiol: is there any advantage in actually updating django objects instead of deleting and re-creating them with new content?

api/desecapi/models.py
peterthomassen: done

api/desecapi/models.py
peterthomassen: done

api/desecapi/models.py
peterthomassen: No, `qs_meaty` is a queryset, i.e. it contains stuff that is currently from the database, and thus it can't contain any new RRsets.

api/desecapi/models.py
peterthomassen: done

api/desecapi/models.py
peterthomassen: You mean, in updating the `RR` objects? It could be done, but their number could change (say, someone adds a second IP address to an RRset of type `A`). We would then have to figure out 1) in case of increase, how many can be updated and then create the other ones only, 2) in case of decrease how many to update and how many to delete.

Since the deletion and the creation actually are bulk SQL statements, I think that this approach is not only much easier and less error prone, but also potentially more efficient.

peterthomassen: ... not to mention wiring it all up correctly with the parent `RRset`

api/desecapi/serializers.py
```diff
@@ -22,8 +23,16 @@ class Meta:
...
     def _set_records(self, instance):
-        records_data = [{'content': x}
-                        for x in self.context['request'].data['records']]
+        # Although serializer fields have required=True by default, that
+        # setting does not work for the SerializerMethodField "records".
+        # Thus, let's wrap our read access to include the validation check.
+        try:
+            records_data = [{'content': x}
+                            for x in self.context['request'].data['records']]
+        except KeyError:
```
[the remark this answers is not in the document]
peterthomassen: done via `records = self.context['request'].get('records')` and raising exception if `None`

peterthomassen:
Thanks for the constructive review! I integrated all the proposed changes. Looks good?

nils-wisiol:
lgtm, with the exception of the absence of e2e tests, which is covered in #82

nils-wisiol:
nice work

PR #95, opened 2018-03-24
Donations: use creditor ID setting, some minor improvements

nils-wisiol:
This PR fixes a bug where the donation would not use the setting's creditor ID, but instead a hard coded one.

Touching this code, I also fixed some spelling and language, updated the code to remove any whitespace from IBANs and added some basic e2e tests. A new setting is added for the account name that will debit the donor.

DEPLOYMENT: please see commit message f22b738


api/desecapi/models.py
```diff
@@ -366,7 +366,6 @@ def save(self, *args, **kwargs):
         self.iban = self.iban[:6] + "xxx" # do NOT save account details
         super().save(*args, **kwargs) # Call the "real" save() method.
 
-
```
peterthomassen: PEP 8 states:
> Surround top-level function and class definitions with two blank lines

While not my style here, the PEP text is actually very clear. But let's not have this hold up the PR.

api/desecapi/templatetags/sepa_extras.py
```diff
@@ -11,4 +11,10 @@ def clean(value):
     cleaned = re.sub(r'[^A-Za-z0-9 ]','',normalized)
     return cleaned
 
+def remove_whitespaces(value):
+    """removes whitespace from the string"""
+    cleaned = re.sub(r'[\s]','',value)
+    return cleaned
+
 register.filter('clean', clean)
+register.filter('remove_whitespaces', remove_whitespaces)
```
peterthomassen: For brevity, you could have said (same for `clean()`):

    @register.filter
    def remove_whitespaces(value):
        ....

Not a problem though.

test/e2e/spec/donation_spec.js
```diff
@@ -0,0 +1,79 @@
...
+            });
+
+            it("follows donation schema", function () {
+                return expect(response).to.have.schema(apiDonationSchema);
+            });
+
+            it("does not return the full iban", function () {
+                return response.then(function (donationResponse) {
+                    expect(donationResponse.body.iban).to.not.contain(iban);
+                });
```
peterthomassen: Here it would have been nice to check that the response contains the cleaned IBAN.

PR #109, opened 2018-06-18
fix(docs): add missing quotes in bulk examples

peterthomassen:


nils-wisiol:
this is now part of #115

peterthomassen:
moved commit to #115

PR #115, opened 2018-09-01
feat(auth): multiple auth tokens per user

nils-wisiol:
To increase authentication flexibility, this change allows users to
create and name additional authentication tokens. They can be used
independently of the current login session.

This change needs some database migration that IS NOT AUTOMATED.
Automatic migration requires either more priviliges or a bunch of code.
Not worth for the number of deployments (one).

Migrate with e.g.

```
INSERT INTO desecapi_token(`created`, `key`, `user_id`, `name`)
  SELECT `created`, `key`, `user_id`, 'default' FROM authtoken_token;
TRUNCATE TABLE authtoken_token;
```

Failure to apply this will result in all login tokens invalid.

This commit does not break any available API, but only extends it.


api/desecapi/models.py
```diff
@@ -53,6 +55,9 @@ class Token(rest_framework.authtoken.models.Token):
     )
     name = models.CharField(_("Name"), max_length=64)
 
+    def generate_key(self):
+        return b64encode(urandom(21)).decode('utf-8')
```
peterthomassen: fyi, utf-8 is default -- no need to remove though

api/desecapi/authentication.py
```diff
@@ -1,9 +1,14 @@
...
 import base64, os
 from rest_framework import exceptions, HTTP_HEADER_ENCODING
-from rest_framework.authtoken.models import Token
 from rest_framework.authentication import BaseAuthentication, get_authorization_header, authenticate
-from desecapi.models import Domain
+from desecapi.models import Domain, Token
+from rest_framework.authentication import TokenAuthentication as RestFrameworkTokenAuthentication
+
+
+
```
peterthomassen: extra newline

api/desecapi/authentication.py
```diff
@@ -96,6 +101,8 @@ def authenticate_credentials(self, userid, key):
 
 
 class IPAuthentication(BaseAuthentication):
+    model = Token
```
peterthomassen: why is this necessary? `IPAuthentication` does nothing with tokens.

api/desecapi/models.py
```diff
@@ -43,6 +44,20 @@ def create_superuser(self, email, password):
...
+    # relation to user is a ForeignKey, so each user can have more than one token
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, related_name='auth_tokens',
+        on_delete=models.CASCADE, verbose_name=_("User")
+    )
+    name = models.CharField(_("Name"), max_length=64)
+
+    class Meta:
+        abstract = False
+        unique_together = (('user', 'name', 'key'),)
```
peterthomassen: `key` must also be unique, otherwise it may be unclear which user a key points to

api/desecapi/models.py
```diff
@@ -53,6 +55,9 @@ class Token(rest_framework.authtoken.models.Token):
     )
```
peterthomassen: `key` field `max_length` is too large for what's returned by `generate_key()`. (Adding here because I can only comment on line 53 and after)

api/desecapi/models.py
```diff
@@ -6,8 +6,9 @@
 from desecapi import pdns, mixins
 import datetime, uuid
 from django.core.validators import MinValueValidator
-from rest_framework.authtoken.models import Token
 from collections import OrderedDict
+import rest_framework.authtoken.models
+from django.utils.translation import ugettext_lazy as _
```
peterthomassen: Why using internationalization? We're not using it anywhere else.

api/desecapi/models.py
```diff
@@ -43,6 +44,20 @@ def create_superuser(self, email, password):
...
 
 
+class Token(rest_framework.authtoken.models.Token):
+    key = models.CharField(_("Key"), max_length=40, db_index=True, unique=True)
+    # relation to user is a ForeignKey, so each user can have more than one token
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL, related_name='auth_tokens',
+        on_delete=models.CASCADE, verbose_name=_("User")
+    )
+    name = models.CharField(_("Name"), max_length=64)
```
peterthomassen: No default? (The migration uses `'default'`.)

test/e2e/spec/api_spec.js
```diff
@@ -881,6 +957,36 @@ describe("API", function () {
...
 
+            describe("tokens/ endpoint", function () {
+
+                var tokenId;
+
+                function createToken () {
+                    return chakram.post('/tokens/', { name: "e2e-token" }).then(function (response) {
+                        expect(response).to.have.status(201);
+                        tokenId = response.body['id'];
+                    })
```
peterthomassen: check that `name` is set correctly

api/desecapi/views.py
```diff
@@ -41,6 +46,52 @@ def get_client_ip(request):
     return request.META.get('REMOTE_ADDR')
 
 
+class TokenCreateView(djoser.views.TokenCreateView):
+
+    def _action(self, serializer):
+        user = serializer.user
+        token = Token(user=user)
```
peterthomassen: Elsewhere I suggested to have a default token name (because the migration assigns the name `'default'` to existing tokens). Seeing this line now, I agree that a token created upon regular login should have empty name, or name `'login'` or something similar.

api/desecapi/views.py
```diff
@@ -41,6 +46,52 @@ def get_client_ip(request):
     return request.META.get('REMOTE_ADDR')
 
 
+class TokenCreateView(djoser.views.TokenCreateView):
+
+    def _action(self, serializer):
```
peterthomassen: The original Djoser `login_user()` function in `utils.py` uses `settings.TOKEN_MODEL.objects.get_or_create()`. The right way to handle this would be to assign a non-standard `Manager` to the token model, so that `get_or_create()` does what you want to do (in the new `Token` model). Then, this view (which is essentially code duplication) is unnecessary.

api/desecapi/views.py
```diff
@@ -41,6 +46,52 @@ def get_client_ip(request):
...
+        _, token = auth.TokenAuthentication().authenticate(request)
+        token.delete()
+        user_logged_out.send(
+            sender=request.user.__class__, request=request, user=request.user
+        )
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+class TokenViewSet(mixins.CreateModelMixin,
+                   mixins.DestroyModelMixin,
```
peterthomassen: I am not sure how exactly this works -- I thought that create and list operations are for list endpoints while the destroy action aims at a specific model instance. Where is that taken care of? I suppose it's some magic interaction with the `SimpleRouter` used in `urls.py`. Can you elaborate briefly?

nils-wisiol:
current docs snapshot: https://a4a.de/.upload/1536244733_index.html

nils-wisiol:
new migration:

```
INSERT INTO desecapi_token(`created`, `key`, `user_id`, `name`. `user_specific_id`)
  SELECT `created`, `key`, `user_id`, 'default', 1 FROM authtoken_token;
TRUNCATE TABLE authtoken_token;
```

PR #132, opened 2019-02-15
20190215 bugfixes

peterthomassen:


nils-wisiol:
I think this looks good, but I still need to run the tests. Also, I believe some clients could break with the changes to the API, how should we proceed here? Emails? We could also hold back some of the changes until a later date and publish them with more breaking changes together.

nils-wisiol:
Tests run through.

peterthomassen:
The only changes of working functionality are:

- When updating dynamic records via https://update.dedyn.io/, send the `good` response as `text/plain` instead of `text/html`
- When the dynamic domain cannot be found or the auth token is invalid, respond with `nohost` and `badauth`, resp. (as `text/plain`). These are error scenarios anyways.

These changes are expected to have negligible impact on working installations. DId I miss any breaking changes?

nils-wisiol:
I'm mostly concerned by changing the status code for bad authentication from 401 to 200. I looked into the dyn.com specifications, which states that the http status is irrelevant. Nevertheless, I feel like we could either delay that change until we inevitably have breaking changes anyways, or we update the code to keep the 401 http status

peterthomassen:
For the domain-not-found scenario, I changed the status back to 404 as discussed offline. Ready for merge?

PR #139, opened 2019-02-22
feat(api): return status 503 when database is unavailable

### LINKED ISSUE #134: Disguise database connection errors as 503?
When the database connection cannot be established, this is usually because of the deployment process restarting the cluster. Currently, this leads to `500 Internal Server Error` due to an error raised by the local database client library. As the error cause is known and intended, it would be justified to replace this with `503 Service Unavailable`.

The question is for which error codes to do that. Here's the server error codes reference: https://mariadb.com/kb/en/library/mariadb-error-codes/ However, errors generated by the database server (as opposed to connection errors on the client side) are real errors and should not be disguised.

This is the client error code reference: https://dev.mysql.com/doc/refman/8.0/en/client-error-reference.html The following list appears reasonable: 2002, 2003, 2005, 2007, 2009, 2026

@nils-wisiol

nils-wisiol:
Previously, we returned an unspecific 500 error code. Fixes #134.

Technical details: https://www.django-rest-framework.org/api-guide/exceptions/#custom-exception-handling

### DIFF
--- a/api/desecapi/exception_handlers.py
+++ b/api/desecapi/exception_handlers.py
@@ -1,0 +1,29 @@
+from django.db.utils import OperationalError
+from rest_framework import status
+from rest_framework.response import Response
+from rest_framework.views import exception_handler, set_rollback
+
+
+def handle_db_unavailable(exc, context):
+    """
+    desecapi specific exception handling. If no special treatment is applied,
+    we default to restframework's exception handling. See also
+    https://www.django-rest-framework.org/api-guide/exceptions/#custom-exception-handling
+    """
+
+    if isinstance(exc, OperationalError):
+        if isinstance(exc.args, (list, dict, tuple)) and exc.args and \
+            exc.args[0] in (
+                2002,  # Connection refused (Socket)
+                2003,  # Connection refused (TCP)
+                2005,  # Unresolved host name
+                2007,  # Server protocol mismatch
+                2009,  # Wrong host info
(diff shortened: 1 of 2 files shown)


api/desecapi/exception_handlers.py
```diff
@@ -0,0 +1,32 @@
...
+            exc.args[0] in (
+                2002,  # Connection refused (Socket)
+                2003,  # Connection refused (TCP)
+                2005,  # Unresolved host name
+                2007,  # Server protocol mismatch
+                2009,  # Wrong host info
+                2026,  # SSL connection error
+        ):
+            # Gracefully let clients know that we cannot connect to the database
+            msg = _('Please try again later.')
```
peterthomassen: Why do we use i18n here while we use it nowhere else?

nils-wisiol: This follows the implementation that's also [used in rest framework 3.7](https://github.com/encode/django-rest-framework/blob/3.7.7/rest_framework/views.py#L63), however was changed in 3.8 and 3.9 (current). That is, our code currently follows 3.7 style, which is in line with our `requirements.txt` (as upgrading is blocked by the bulk restframework module).

Which style should we follow?

nils-wisiol: As discussed offline, we will stick to restframework 3.7 style, but remove the i18n support.

PR #140, opened 2019-02-22
Introduce API Versions and Documentation

### LINKED ISSUE #137: Rethink API Versions
Currently, we are developing the API under the path prefix `api/v1`. This proposes to introduce the path prefix `api/v0` and update documentation to only use `api/v0`. In fact, `api/v1` shall be an alias for `api/v0` for now, but make it clear to users that version 0 is unstable and may change on short notice or even without notice.

When we're satisfied with version 0 (and version 1, as it is an alias), we freeze version 1 and introduce `api/v2` in a similar manner. (This will raise the question if v2 will be stable or not, but that we do not necessarily decide right now.)

This is mainly a proposal to communicate clearly with versions are stable and which ones are not. It is explicitly not aimed at making any technical implementation easier.

If we go forward with this proposal, here's a list of things we need to do:

- [ ] introduce v0 alias technically
- [ ] update docs to use v0 everywhere
- [ ] extend docs to state how we number API versions, including a policy when and how versions change, possibly including support periods (having in mind that old, frozen versions may need updating because they are broken or insecure)

nils-wisiol:
This PR introduces API Version 2, that is a currently an alias for Version 1 and shall later be a stable copy of Version 1. Along the way, this PR cleans up the API definitions and re-organizes them. It also includes end-user documentation for API versioning.

Closes #137

### DIFF
--- a/api/api/urls.py
+++ b/api/api/urls.py
@@ -1,38 +1,61 @@
-from django.conf.urls import include, url
-from desecapi import views
-from djoser.views import UserView
-from rest_framework.routers import SimpleRouter
+from django.urls import include, path
 
-tokens_router = SimpleRouter()
-tokens_router.register(r'', views.TokenViewSet, base_name='token')
 
-auth_urls = [
-    url(r'^users/create/$', views.UserCreateView.as_view(), name='user-create'),  # deprecated
-    url(r'^token/create/$', views.TokenCreateView.as_view(), name='token-create'),  # deprecated
-    url(r'^token/destroy/$', views.TokenDestroyView.as_view(), name='token-destroy'),  # deprecated
-    url(r'^users/$', views.UserCreateView.as_view(), name='register'),
-    url(r'^token/login/$', views.TokenCreateView.as_view(), name='login'),
-    url(r'^token/logout/$', views.TokenDestroyView.as_view(), name='logout'),
-    url(r'^tokens/', include(tokens_router.urls)),
-    url(r'^me/?$', UserView.as_view(), name='user'),
-    url(r'^', include('djoser.urls.authtoken')),
-]
+#
+# On Reversing URLs
+# =================
+#
+# Recommended Usage:
+# Always use rest_framework.reverse.reverse, do not directly use django.urls.reverse.
+# If a request object r is available, use reverse(name, request=r). With the name as defined
+# in desecapi.urls.v1 or desecapi.urls.v2. It will return an URL maintaining the currently requested API version.
+# If there is no request object available, e.g. in commands, a mock object can be constructed
+# carrying all information that is necessary to construct a full URL:
+#
+#         from django.test import RequestFactory
+#         from rest_framework.versioning import NamespaceVersioning
+#         from api import settings
+#
+#         r = RequestFactory().request(HTTP_HOST=settings.ALLOWED_HOSTS[0])
+#         r.version = 'v1'
+#         r.versioning_scheme = NamespaceVersioning()
+#
+# Also note in this context settings.REST_FRAMEWORK['ALLOWED_VERSIONS'] and
+# settings.REST_FRAMEWORK['DEFAULT_VERSIONING_CLASS']. (The latter is of type string.)
+#
+# Advanced Usage:
+# Prefix the name of any path with 'desecapi' to get the default version,
+# or prefix the name of any path with the desired namespace, e.g. 'v1:root'.
+# In this case, the version information of the request will be ignored and
+# providing a request object is optional. However, if no request object is provided,
+# only a relative URL can be generated.
+#
+# Examples:
+# The examples refer to the version used by the client to connect as the REQUESTED version,
+# the version specified by the first argument to reverse as the SPECIFIED version, and to the
+# version defined as default (see below) as the DEFAULT version.
+#
+#         reverse('root', request) -> absolute URL, e.g. https://.../api/v1/, with the REQUESTED version
+#         reverse('root') -> django.urls.exceptions.NoReverseMatch
+#         reverse('desecapi:root') -> relative URL, e.g. api/v1/, with the DEFAULT version
+#         reverse('v2:root') -> relative URL, e.g. api/v2/, with the SPECIFIED version
+#         reverse('v2:root', request) -> absolute URL, e.g. https://.../api/v2/, with the SPECIFIED version
+#         reverse('desecapi:root', request) -> absolute URL, e.g. https://.../api/v1/, with the DEFAULT version
+#         reverse('v1:root', request) -> absolute URL, e.g. https://.../api/v1/, with the SPECIFIED version
+#
+# See Also:
+# https://github.com/encode/django-rest-framework/issues/5659
+# https://github.com/encode/django-rest-framework/issues/3825
+#
+# Note that from the client's perspective, there is no default version: each request needs to
+# specify the version in the request URL.
+#
 
-api_urls = [
-    url(r'^$', views.Root.as_view(), name='root'),
-    url(r'^domains/$', views.DomainList.as_view(), name='domain-list'),
-    url(r'^domains/(?P<name>[a-zA-Z\.\-_0-9]+)/$', views.DomainDetail.as_view(), name='domain-detail'),
-    url(r'^domains/(?P<name>[a-zA-Z\.\-_0-9]+)/rrsets/$', views.RRsetList.as_view(), name='rrsets'),
-    url(r'^domains/(?P<name>[a-zA-Z\.\-_0-9]+)/rrsets/(?P<subname>(\*)?[a-zA-Z\.\-_0-9=]*)\.\.\./(?P<type>[A-Z][A-Z0-9]*)/$', views.RRsetDetail.as_view(), name='rrset'),
-    url(r'^domains/(?P<name>[a-zA-Z\.\-_0-9]+)/rrsets/(?P<subname>[*@]|[a-zA-Z\.\-_0-9=]+)/(?P<type>[A-Z][A-Z0-9]*)/$', views.RRsetDetail.as_view(), name='rrset@'),
-    url(r'^dns$', views.DnsQuery.as_view(), name='dns-query'),
-    url(r'^dyndns/update$', views.DynDNS12Update.as_view(), name='dyndns12update'),
-    url(r'^donation/', views.DonationList.as_view(), name='donation'),
-    url(r'^unlock/user/(?P<email>.+)$', views.unlock, name='unlock/byEmail'),
-    url(r'^unlock/done', views.unlock_done, name='unlock/done'),
-]
-
+# IMPORTANT: specify default version as the last element in the list
+# if no other information is available, the last-specified version will be used as default for reversing URLs
 urlpatterns = [
-    url(r'^api/v1/auth/', include(auth_urls)),
-    url(r'^api/v1/', include(api_urls)),
+    # other available versions in no particular order
+    path('api/v2/', include('desecapi.urls.version_2', namespace='v2')),
+    # the DEFAULT version
+    path('api/v1/', include('desecapi.urls.version_1', namespace='v1')),
 ]
--- a/api/desecapi/tests/testdomains.py
+++ b/api/desecapi/tests/testdomains.py
@@ -1,7 +1,7 @@
-from django.urls import reverse
+from rest_framework.reverse import reverse
 from rest_framework import status
 from rest_framework.test import APITestCase
-from .utils import utils
+from desecapi.tests.utils import utils
 from desecapi.models import Domain
 from django.core import mail
 import httpretty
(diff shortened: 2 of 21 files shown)


peterthomassen:
I don't quite understand the approach. Based on the lifecycle section that you added in the docs, it seems like `v1` is the dev version and `v2` is production, but unreleased. How can that be?

Wouldn't we want to currently have `v1` with "release candiate"/"staging" status, and once we declare it stable have a follow up version `v2`?

Also, would it be worth considering `v0` for whatever we have before full launch, and at that point freeze the current state, naming in `v1`?

Also, we'd need to align the tests with whatever scheme we decide to use.

nils-wisiol:
Let's have an offline discussion about this, then report back with the results

nils-wisiol:
I dismissed the idea of a `v0` version and updated the docs to have a life cycle for each API version, starting from v1.

Also, I prepared code and docs for future versions.

nils-wisiol:
Before this is ready to review, we should decide how to handle the docs. readthedocs.io currently maintains a version of the docs for each branch, and I believe for each tag (we currently have none) of the desec-stack repository. We could therefore rely on readthedocs.io to maintain a API version 1, API version 2 documentation. The other option is to include all documentation in the master branch. Which way should we go? @peterthomassen

peterthomassen:
I prefer that each branch / tag carries the docs that are applicable to it.

nils-wisiol:
I updated the PR to reflect the tag-based approach. It has the downside that the life cycle document now cannot be placed inside the docs, or needs to be updated in old versions. I moved the status table in to README, keeping the remaining information in the docs. The tagging process is somewhat tedious, as tags need to be created and pushed separately, like so:

    git commit
    git tag -f apiv1
    git push
    git push -f origin apiv1

I'm afraid this can easily be forgotten, leading to outdated docs. An alternative way could be to use branches, but I'm not sure if that's any better.

On the other hand, the docs (in particular, the table of contents on [readthedocs.io](https://desec.readthedocs.io/en/latest/)) will not be cluttered with outdated information.

nils-wisiol:
README extension: https://github.com/desec-io/desec-stack/tree/20190222_api_versions#api-versions-and-road-map
readthedocs: https://desec.readthedocs.io/en/apiv1/#api-versions-and-life-cycle (notice the "apiv1")

nils-wisiol:
after much fiddling with the markdown table, this is finally ready for review :)

nils-wisiol:
This PR still has the problem that the unit tests only run for one set of API versions.

nils-wisiol:
Ready for further review.

- I introduced rest framework versioning together with django instance namespacing for URLs, [in a bit of an abuse](https://docs.djangoproject.com/en/2.2/topics/http/urls/#introduction) (see api/urls.py, where version 1 and version 2 are included as different instances of the same app).
- I added extensive documentation for how to use rest framework's `reverse()` in this setting in `/api/api/urls.py`.
- Default behavior for all linking is now to maintain the version that the user used to contact the API. If that is not provided or not available (let's say in the context of a django command), the code needs to define the behavior, otherwise an exception is raised.
- The unit tests are only run for version 1 of the API, this makes sense, as a new version also requires new tests. The fact that version 2 is a copy of version 1 as of now is immaterial.
- As a bit of (unrelated) sugar, I added a coverage report to the unit tests run on travis. We can later upgrade this to be shown on the readme or required to increase (not decrease) with every PR or what not.
- README now shows API version status, readthedocs.io shows API docs for the current and tagged versions. Updating a version now requires that the tag is newly set by using `git tag -f v1` and `git push -f origin v1`, **otherwise the docs will be out of date**.
- What should be up for debate is the fact that I decided [against all advise](https://www.django-rest-framework.org/api-guide/versioning/#acceptheaderversioning) that we keep the API version number in the URLs. I believe using versioning based on the accept header may be confusing for newcomers.

api/api/urls.py
```diff
@@ -1,38 +1,57 @@
...
+# If there is no request object available, e.g. in commands, a mock object can be constructed
+# carrying all information that is necessary to construct a full URL:
+#
+#         from django.test import RequestFactory
+#         from rest_framework.versioning import NamespaceVersioning
+#         from api import settings
+#
+#         r = RequestFactory().request(HTTP_HOST=settings.ALLOWED_HOSTS[0])
+#         r.version = 'v1'
+#         r.versioning_scheme = NamespaceVersioning()
```
peterthomassen: How about writing a class for this, like `desecapi.AbstractRequest`? This would remove code duplication wherever this pattern is needed, and also the need to know about the details, like which versioning scheme is specified in the settings.

peterthomassen: I suppose one could also retrieve `NamespaceVersioning` from `settings.py` and assign the property here dynamically.

api/api/urls.py
```diff
@@ -1,38 +1,57 @@
...
+#         reverse('desecapi:root', request) -> absolute URL, e.g. https://.../api/v1/, with the DEFAULT version
+#         reverse('v1:root', request) -> absolute URL, e.g. https://.../api/v1/, with the SPECIFIED version
+#
+# See Also:
+# https://github.com/encode/django-rest-framework/issues/5659
+# https://github.com/encode/django-rest-framework/issues/3825
+#
 
+# IMPORTANT: specify default version last
+# if no other information is available, the last version will be used as default for reversing URLs
```
peterthomassen: The "last" version, meaning the latest (highest number), or the last element of the list? I'm like 70:30 certain that you mean the latter, but maybe we can make this more clear by referring to the last list element explicitly.

api/desecapi/tests/testregistration.py
```diff
@@ -105,20 +108,23 @@ def test_multiple_registration_not_locked_same_ip_long_time(self):
...
+        url = reverse('v1:register')
         data = {'email': utils.generateUsername(),
                 'password': utils.generateRandomString(size=12), 'dyn': True}
         response = self.client.post(url, data, REMOTE_ADDR="1.3.3.10")
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         user = models.User.objects.get(email=data['email'])
-        send_account_lock_email(None, user)
+        r = RequestFactory().request(HTTP_HOST=settings.ALLOWED_HOSTS[0])
+        r.version = 'v1'
+        r.versioning_scheme = NamespaceVersioning()
```
peterthomassen: See my comment in `urls.py`. Knowledge of the versioning scheme class is overspecific here, I think. A wrapper would solve this.

api/desecapi/views.py
```diff
@@ -297,14 +297,15 @@ class Root(APIView):
...
             return Response({
-                'domains': reverse('domain-list'),
-                'user': reverse('user'),
-                'logout': reverse('token-destroy'),  # TODO change interface to token-destroy, too?
+                'domains': reverse('domain-list', request=request),
+                'user': reverse('user', request=request),
+                'logout': reverse('token-destroy', request=request),  # TODO change interface to token-destroy, too?
             })
         else:
+            print(request.version)
```
peterthomassen: remove

api/desecapi/tests/testdomains.py
```diff
@@ -11,22 +11,22 @@
 
 class UnauthenticatedDomainTests(APITestCase):
     def testExpectUnauthorizedOnGet(self):
-        url = reverse('domain-list')
+        url = reverse('v1:domain-list')
```
peterthomassen: The version string is duplicated here many many times. How about importing a wrapped version of this, e.g. `from desec.test.utils import reverse`? We could have this function use some test-global state (as we have for the test user settings) to store the version number, and then just call the native `reverse()` with the first argument augmented.

api/desecapi/tests/testdomains.py
```diff
@@ -1,7 +1,7 @@
 from django.urls import reverse
```
peterthomassen: In `urls.py`, you wrote: 

> Always use `rest_framework.reverse.reverse`, do not directly use `django.urls.reverse`

Should we follow this recommendation here (and in the rest of the code)?

docs/lifecycle.rst
```diff
@@ -0,0 +1,31 @@
+API Versions and Life Cycle
```
peterthomassen: This section is generally clear, but I think we can smoothen the language a bit. If you don't mind, I'd proofread this and add a commit with suggestions, which you can either accept or pick from. Ok for you?

api/desecapi/urls/version_1.py
```diff
@@ -0,0 +1,61 @@
...
+    url(r'^dns$', views.DnsQuery.as_view(), name='dns-query'),
+
+    # DynDNS update endpoint
+    url(r'^dyndns/update$', views.DynDNS12Update.as_view(), name='dyndns12update'),
+
+    # Donation endpoints
+    url(r'^donation/', views.DonationList.as_view(), name='donation'),
+
+    # Unlock endpoints
+    url(r'^unlock/user/(?P<email>.+)$', views.unlock, name='unlock/byEmail'),
```
peterthomassen: I just realized that this causes PII to be logged (in this case, email address). This is considered bad practice. Also, there may be issues with URI encoding edge cases etc.

I created an issue for this (#161), but I don't think it needs to be in scope for launch.

api/desecapi/urls/version_1.py
```diff
@@ -0,0 +1,61 @@
...
+    url(r'^token/create/$', views.TokenCreateView.as_view(), name='token-create'),  # deprecated
+    url(r'^token/destroy/$', views.TokenDestroyView.as_view(), name='token-destroy'),  # deprecated
+
+    # New user management
+    url(r'^users/$', views.UserCreateView.as_view(), name='register'),
+
+    # Token management
+    url(r'^token/login/$', views.TokenCreateView.as_view(), name='login'),
+    url(r'^token/logout/$', views.TokenDestroyView.as_view(), name='logout'),
+    url(r'^', include('djoser.urls.authtoken')),  # note: this is partially overwritten by the two lines above
```
peterthomassen: This line has now moved up a bit, in comparison to the previous arrangement (https://github.com/desec-io/desec-stack/blob/224eeb9fef477332dc5eb3e67702d5fc6707cb37/api/api/urls.py#L18). Does this have any implications, and if so, what's the reasoning behind the change?

api/requirements.txt
```diff
@@ -8,3 +8,4 @@ requests~=2.21.0
 uwsgi~=2.0.0
 django-nocaptcha-recaptcha==0.0.20  # updated manually
 djangorestframework-bulk~=0.2.0
+coverage~=4.5.3
```
peterthomassen: Is there a way to only have such dependencies when testing, but not in the final build?

This does not need to be solved now, but if we think about hardening later, we should make sure that stuff not used at runtime does not get built into the image.

api/api/urls.py
nils-wisiol: It's an option, but it's not needed right now. We do not use `resolve()` in django commands right now. Once we do, we can see if there is actual code duplication. Also, rest_framework is working on a better solution that does not rely on the `django.test.RequestFactory`.

api/api/urls.py
nils-wisiol: If in line "last" is replaced with "latest", the sentence does not make sense. Nevertheless, I replace it with the next push

api/desecapi/tests/testregistration.py
nils-wisiol: It is very specific, and it needs to be, as the framework has no other source of knowing which versioning scheme is used. Since we're testing if there is an email sent, it's also not important which scheme we're using.

It's the only place where we need to provide a mocked request, which is why I do not want to wrap it.

api/desecapi/views.py
nils-wisiol: woops.

api/desecapi/tests/testdomains.py
nils-wisiol: I considered and dismissed that idea. Each test is written for a specific version of the API, reusing tests is complicated and may be of limited use. Because we cannot decide today how to reuse tests, as we do not know how the future version may look like, it is fair to update all tests to their specific version. Admittedly, it looks hideous in the commit diff, but it's not a lot of work with search and replace if it needs to be changed later. Again, today, I do not see a scenario where it is useful to change all tests to another API version.

api/desecapi/tests/testdomains.py
nils-wisiol: yup. Fixed with next push

docs/lifecycle.rst
nils-wisiol: Go nuts, commit your changes, and let me know which id to cherry-pick

api/desecapi/urls/version_1.py
nils-wisiol: It's rearranged and documented with line comments

api/requirements.txt
nils-wisiol: It's possible, but it introduces additional complexity (such as waiting for coverage to be installed with every test run or having a second requirements.txt file). I believe as of now, the disadvantages outweigh the advantages. (Also considering that we do not even install the latest version of djangorestframework and use the abandoned django-nocaptcha-recaptcha and djangorestframework-bulk)

nils-wisiol:
I noticed while cleaning up the import statements that we're using the outdated `url()` function to specify the URL configuration. `url()` is an alias for `re_path()`, I updated the code accordingly with a46eb30ffd29ef02a6314f9d96b7e9add4d41119. Also, I removed regex where it was not necessary.

api/desecapi/tests/testregistration.py
peterthomassen: The code _can_ figure out which versioning scheme is used, by looking at the settings. Anyways, I agree with your conclusion.

docs/lifecycle.rst
peterthomassen: See https://github.com/desec-io/desec-stack/pull/140/commits/d5b82b5e986382548ee87acf7ab634aa4c0c14be

api/desecapi/urls/version_1.py
peterthomassen: Yes, of course. Does it have any implications though (for example, has the way that routes can shadow each other changed)?

nils-wisiol: imho it's functionally equivalent

peterthomassen:
I added my docs modifications. Once you reviewed them, I think we're good to go (modulo my question on the order in `urls.py`).

I suggest to only keep the `import` cleanup commit and the coverage integration commit separate, and squash the rest.

nils-wisiol:
imho ready to merge

Issue #47, opened 2017-04-18
api: make sure unit tests are not using nslord/nsmaster


peterthomassen:

peterthomassen:
Most crucially, put an end to the tests interacting with `nslord` and `nsmaster`. Those containers should not even be required to run, as long as their responses are mocked properly.

This will allow us to also add a test verifying that `NS` delegation records are created under `dedyn.io` for dynamic domains. Since this requires creating a `dedyn.io` zone in the test database, such a test would currently lead to complications due to said interactions.

PR #144, opened 2019-03-04
20190304 simplify apex access

peterthomassen:


peterthomassen:
@coderbyheart whohoo! Thanks for discussing this together.

nils-wisiol:
Lgtm. It could include a little more e2e testing, but I'm okay with the "unit" tests for now. Also, see #140 for a little more thoughts on urls.py.

peterthomassen:
Yes, I agree about the e2e tests, but it seemed quite some work, and I figured improvement will be covered by #47. I'm planning to rework this before public launch.

PR #148, opened 2019-04-07
feature(api): remove write-support for locked users, closes #147

### LINKED ISSUE #147: Remove RRset write-support for locked users
Currently, locked users can write RRsets to the application database. Only upon unlocking the user account, these RRsets will be propagated to pdns.

This leads to various difficulties regarding validation and consistency, see https://github.com/PowerDNS/pdns/issues/7565#issuecomment-476397025

As the benefit of the feature is questionable, nobody will miss it if it's not there, and it even may lead to confusion ("My PATCH was successful, but changes don't get propagated?!"), it is better to remove the feature.

peterthomassen:
This removes all write-support for locked users.  Exception: Users
which have the dyn flag set can created domains under .dedyn.io
(but they will not be published until the account is unlocked).

This PR is ready for review. It is only missing a few tests which I will add soon.

The tests will check what happens when a locked account tries
- PATCH/PUT/DELETE domains/:domain/
- POST rrsets/
- PATCH/PUT rrsets/
- PATCH/PUT/DELETE rrsets/.../


peterthomassen:
@nils-wisiol I'd also like your input about the following: With this PR, we impose the write-lock for locked users in each view (`DomainList`, `RRsetDetail`, ...) in each of the write-related methods on the view-level (`create()`, `update()`, ...). I wondered if a permission class would be more appropriate. What I don't know though is whether permission classes allow distinguishing between read and write, or if it's all or nothing.

nils-wisiol:
Restframework does provide for permissions. One particular interesting example from the docs is [IsAuthenticatedOrReadOnly](https://www.django-rest-framework.org/api-guide/permissions/#isauthenticatedorreadonly), which we should be apply to apply to "is locked or read only".

api/desecapi/models.py
```diff
@@ -121,9 +121,12 @@ def is_staff(self):
...
 
     def unlock(self):
-        # self.locked is used by domain.sync_to_pdns(), so call that first
-        for domain in self.domains.all():
-            domain.sync_to_pdns()
+        # Create domains on pdns that were created after the account was locked.
+        # We filter by created__gt=self.locked, published=None should give the
+        # same result as the published timestamp is updated at domain creation.
+        # TODO Should we check somewhere whether the two filters agree?
+        for domain in self.domains.filter(created__gt=self.locked):
```
nils-wisiol: I propose to follow the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle): "Be conservative in what you do, be liberal in what you accept from others". Applying it to the situation at hand, I propose that `unlock()` will unlock all locked domains, regardless of when and why they have been locked. One way to do that could be to wrap each `domain.create_on_pdns()` in a `try`/`except` block and ignore exceptions due to already existing domains.

peterthomassen: I agree with the principle, but I don't think it applies to the case at hand: If we determine that there is an unpublished domain from the time before the account lock, then that's not a user input to the unlock action. In other words, it is an unexpected internal condition and not something "from others" that we may be willing to accept.

I'm leaning towards checking that the internal condition is in a well-defined state, and otherwise do something conservative, including notifying us.

Also, it may happen in the future that we have unpublished domains for some valid use case, and we may not want to publish them automatically when the user account is unlocked.

So, if you don't have a strong opinion on this, I'll `assert` that either filter condition returns the same number of domains (this is not a perfect check).

peterthomassen:
I switched to permissions classes and added the tests. Should be good to go now; pls review.

peterthomassen:
rebased on top of master

api/desecapi/models.py
nils-wisiol: By "from others", I meant the current state of the database. Let it be legacy code that created that, or manual iteration with it. I'm proposing to state the exact behavior of unlocking in a function doc comment and remove the check, but I do not have a strong opinion.

nils-wisiol:
I added a commit with four proposed changes: a few words on the docs, a dead import and formatting. Please pick or discard at will. Otherwise, this lgtm.

(I noticed that this PR is building new code on top of the deprecated `dyn` flag. Since it's fixing a bug, I believe that's okay.)

api/desecapi/models.py
peterthomassen: I tried to state behavior more clearly now in the comment, see https://github.com/desec-io/desec-stack/pull/148/commits/4f778bce174302273d09f4dcbe2247f4e0e8d16d#diff-d93a995458c0f3a31fb831c4f43b5d27R127

If this is ok for you, we can merge.

peterthomassen:
i squashed all of your proposed changes into my commits. Thanks for the suggestions!

I also improved the comment regarding the unlocking mechanism. Lgty?

PR #172, opened 2019-04-26
20190426 lowercase domain

peterthomassen:
This PR

- fixes up the previous migration squash
- enforces strictly lowercase domain names and subnames (see explanation below) on the model layer (regex validation has mostly been moved there; slashes and lowercase are gone)
- adds and fixes a few tests
- updates the docs

It turns out that it is /very/ complicated to allow any-case input while lowercasing things internally. There are a very large number of locations in the code where lookups start to fail; also in the tests, the checks for the number of expected internal requests containing certain strings in the body etc. start failing.

I spent a few hours with this and then decided to strictly limit domain names and subnames to lowercase for now. If people start asking for more flexibility, we can still spend more time on this.


peterthomassen:
Blocker: We should add one more commit to allow arbitrary case in the dynDNS update view so that people's uppercase dynDNS configurations don't break.

peterthomassen:
done (added uppercase tolerance for dynDNS)

peterthomassen:
I modified the migration such that Domain.name and RRset.subname are lowercased in bulk. Tested locally and worked.

PR #174, opened 2019-04-27
Restrict all Domain Names and Subnames to Lowercase Fixes #54 Closes #172 Closes #173

### LINKED ISSUE #54: api: lowercase domain names upon creation
Per RFC 4343, Section 4, case can be changed at input, and per Section 6, it is more or less recommended to do so. This allows us to deal with uniqueness constraints more easily.

If possible, allow users to see their preferred spelling.

nils-wisiol:
In the main commits 245fff0 and 245fff0 of this PR, we extend the model layer to enforce lowercase-only Domain.names and RRset.subnames. The serializer layer is extended to restrict user-input to lowercase.

This also

- fixes the squashed migration
-    fixes some bugs in the Django tests, extends the test's base.py to provide more helpful output for failing tests
-    adds tests for domain names containing underscores _
-    adds a test to see if users can steal other user's zones
-    fixes the docs with regard to slahes / in subnames

This PR replaces #172 and #173: closes #172, closes #173


nils-wisiol:
Lgtm, but let's test that for the known production cases were it it broken right now

peterthomassen:
works. rebasing and merging

PR #176, opened 2019-04-29
fix(tests): typo in RRsets filter

peterthomassen:
I found a typo in the RRsets filter used in tests. I appears to have been without consequence; I wonder how that can be? You can probably figure it out quicker as it's your code.

### DIFF
--- a/api/desecapi/tests/testrrsets.py
+++ b/api/desecapi/tests/testrrsets.py
@@ -110,7 +110,7 @@ class AuthenticatedRRSetTestCase(DomainOwnerTestCase):
     @staticmethod
     def _filter_rr_sets(rr_sets, **kwargs):
         return [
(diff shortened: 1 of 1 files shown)


nils-wisiol:
`_filter_rr_sets` is currently only used in the two methods directly below: [`assertRRSetCount`](https://github.com/desec-io/desec-stack/blob/b092660e6fe8b6172d97ab1b07101718befcf72a/api/desecapi/tests/testrrsets.py#L117) and [`assertContainsRRSet`](https://github.com/desec-io/desec-stack/blob/b092660e6fe8b6172d97ab1b07101718befcf72a/api/desecapi/tests/testrrsets.py#L124). Both only care about the length of the object returned, hence the error did not surface.

PR #177, opened 2019-05-01
Add bulks tests

peterthomassen:
This PR generalizes some of the testing helper functions and then adds testing of bulk RRsets functionality.


api/desecapi/tests/testrrsets.py
```diff
@@ -61,11 +61,11 @@ def _test_rr_sets(cls, subname=None, type_=None, records=None, ttl=None):
             (subname_, 'TXT', ['"hey ho, let\'s go!"'], 134)
             for subname_ in cls.SUBNAMES
         ] + [
-            (subname_, type_, ['"10 mx1.example.com."'], 101)
+            (subname_, type_, ['10 mx1.example.com.'], 101)
             for subname_ in cls.SUBNAMES
             for type_ in ['MX', 'SPF']
         ] + [
-            (subname_, 'A', ['"1.2.3.4"'], 187)
+            (subname_, 'A', ['1.2.3.4'], 187)
```
nils-wisiol: Note in the test class docs that we do not test validation

api/desecapi/tests/testrrsets.py
```diff
@@ -108,24 +108,25 @@ def assertRRSet(self, response_rr, domain=None, subname=None, records=None, type
...
+        return [len(_filter_rr_sets_by_mask(rr_sets, mask)) for mask in masks]
 
-    def assertRRSetCount(self, rr_sets, count, **kwargs):
-        filtered_rr_sets = self._filter_rr_sets(rr_sets, **kwargs)
-        if len(filtered_rr_sets) != count:
-            self.fail('Expected to find %i RR set(s) with %s, but only found %i in %s.' % (
-                count, kwargs, len(filtered_rr_sets), rr_sets
+    def assertRRSetsCount(self, rr_sets, masks, count=1):
+        counts = self._count_occurrences_by_mask(rr_sets, masks)
+        if not all([x == count for x in counts]):
```
nils-wisiol: rename to `expected_count` and `is_counts` to ease reading?

api/desecapi/tests/testrrsets.py
```diff
@@ -189,6 +190,107 @@ def test_create_my_rr_sets(self):
                 self.assertStatus(response, status.HTTP_200_OK)
                 self.assertRRSet(response.data, **data)
 
+    def test_bulk_post_my_rr_sets(self):
+        data = []
+        for subname in ['', 'create-my-rr-sets', 'foo.create-my-rr-sets', 'bar.baz.foo.create-my-rr-sets']:
```
nils-wisiol: use `self.SUBNAMES`? (if necessary can be used with `self.my_empty_domain`)

api/desecapi/tests/testrrsets.py
```diff
@@ -189,6 +190,107 @@ def test_create_my_rr_sets(self):
...
+
+        with self.assertPdnsRequests(self.requests_desec_rr_sets_update(name=self.my_empty_domain.name)):
+            response = self.client.bulk_post_rr_sets(domain_name=self.my_empty_domain.name, payload=data)
+            self.assertStatus(response, status.HTTP_201_CREATED)
+
+        response = self.client.get_rr_sets(self.my_empty_domain.name)
+        self.assertStatus(response, status.HTTP_200_OK)
+        self.assertRRSetsCount(response.data, data)
+
+        for value in data:
```
nils-wisiol: move this into a single call, e.g. `self.assertAPIRRSet(domain_name, subname, type_, kwargs)` where `kwargs = value`

api/desecapi/tests/testrrsets.py
```diff
@@ -189,6 +190,107 @@ def test_create_my_rr_sets(self):
...
+
+        data_no_type = copy.deepcopy(data_full)
+        data_no_type[1].pop('type')
+
+        # Fresh records need TTL and type
+#        with self.assertPdnsNoRequestsBut():
+#            response = self.client.bulk_patch_rr_sets(domain_name=self.my_empty_domain.name, payload=data_no_records)
+#            self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
+#            self.assertEqual(response.data, [{'ttl': ['This field is required for new RRsets.']}, {}])
+        with self.assertPdnsNoRequestsBut():
```
nils-wisiol: this context manager has no effect

api/desecapi/tests/testrrsets.py
```diff
@@ -189,6 +190,107 @@ def test_create_my_rr_sets(self):
...
+            response = self.client.bulk_patch_rr_sets(domain_name=self.my_empty_domain.name, payload=data_no_ttl)
+            self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
+            self.assertEqual(response.data, [{'ttl': ['This field is required for new RRsets.']}, {}])
```
nils-wisiol: maybe we should transform these kind of statements into `assertData` or `assertBody` similar to what we did with `assertStatus`

api/desecapi/tests/testrrsets.py
nils-wisiol: Similar, we could implement `self.assertDBRRSet` to check directly against the database (see also your TODO)

api/desecapi/tests/testrrsets.py
```diff
@@ -189,6 +190,107 @@ def test_create_my_rr_sets(self):
...
+        with self.assertPdnsNoRequestsBut():
+            response = self.client.bulk_patch_rr_sets(domain_name=self.my_empty_domain.name, payload=data_no_ttl)
+            self.assertStatus(response, status.HTTP_200_OK)
+        with self.assertPdnsNoRequestsBut():
+            response = self.client.bulk_patch_rr_sets(domain_name=self.my_empty_domain.name,
+                                                      payload=data_no_records)
+            self.assertStatus(response, status.HTTP_200_OK)
+        ### TODO check that RRsets actually exist
+
+    def test_bulk_put_my_rr_sets(self):
```
nils-wisiol: instead write in above test case: `for method in [self.client....put, self.client....patch]`?

api/desecapi/tests/testrrsets.py
peterthomassen: no necessary as of now, we just get all RRsets via API and then use `assertRRSetsCount()`

api/desecapi/tests/testrrsets.py
peterthomassen: sure, but I think this is independent of adding bulk request tests

api/desecapi/tests/testrrsets.py
peterthomassen: the various methods have different requirements regarding presence of fields, so we can't simply use a loop, unfortunately.

nils-wisiol:
I suggest to split up the new tests, where possible, along the comments you made. Why not use the comments as test name? This way, it's easier to see what fails in the test log.

I'm opposing to take advantage of str(Domain) == Domain.name. It is implicit and may change in the future. If you still use the whole domain object as argument to the client method, consider renaming it from domain_name to domain.

peterthomassen:
> I suggest to split up the new tests, where possible, along the comments you made. Why not use the comments as test name? This way, it's easier to see what fails in the test log.

I'm going to come up with something like that.

> I'm opposing to take advantage of str(Domain) == Domain.name. It is implicit and may change in the future.

It is not implicit (https://github.com/desec-io/desec-stack/blob/master/api/desecapi/models.py#L393). As usual, things break if the API changes, but I don't see any reason to assume that the API is going to change.

> If you still use the whole domain object as argument to the client method, consider renaming it from domain_name to domain.

I'm not following; the argument name is already `domain`?

peterthomassen:
Ready for review.

Let's discuss the question about using `str(Domain) == Domain.name`.

nils-wisiol:
> It is not implicit (https://github.com/desec-io/desec-stack/blob/master/api/desecapi/models.py#L393). As usual, things break if the API changes, but I don't see any reason to assume that the API is going to change.

It's implicit where you use it. To be precise, I suggest changing `self.reverse('v1:rrsets', name=domain)` to `self.reverse('v1:rrsets', name=domain.name)`. That way, the `__str__` method can be adapted to include more information about the domain, similar to what bbb3d59 does, to assist in debugging (PyCharm's debugger uses `__str__` a lot to produce output.) Also, it seems sloppy to rely on ["the 'informal' or nicely printable string representation of an object"](https://docs.python.org/3/reference/datamodel.html#object.__str__) to identify it.

> I'm not following; the argument name is already `domain`?

You're right, it is. I was confused!

peterthomassen:
> It's implicit where you use it. To be precise, I suggest changing `self.reverse('v1:rrsets', name=domain)` to `self.reverse('v1:rrsets', name=domain.name)`.

We can do that, but it breaks https://github.com/desec-io/desec-stack/blob/20190430_bulk_test/api/desecapi/tests/test_dyndns12update.py#L49, so we would need to introduce a case distinction. For consistency, we should do that in all 9 `_rr_set*` functions that access that endpoint.

Please read the below considerations, and let me know what you think. If you still feel like the change should be made, I'll do it.

> That way, the `__str__` method can be adapted to include more information about the domain

No, we can't adapt `__str__` because it determines the input to our `RRsetSerializer` where the `domain` field is a `StringRelatedField`, which relies on `__str__`: https://www.django-rest-framework.org/api-guide/relations/#stringrelatedfield We already are committed to this.

> to assist in debugging (PyCharm's debugger uses `__str__` a lot to produce output.)

Are you sure it doesn't use `__repr__` (which defaults to `__str__` if undefined)?

> Also, it seems sloppy to rely on ["the 'informal' or nicely printable string representation of an object"](https://docs.python.org/3/reference/datamodel.html#object.__str__) to identify it.

It's the way it is intended in our `RRsetSerializer` (see above), and where I'm using it here in this PR is in the exact same context (namely, in figuring out the URL for the endpoint that's tied to this serializer).

peterthomassen:
Now that we have tests for bulk functionality, I added a commit for bumping DRF vom 3.7 to 3.9. Apparently, this does not conflict with the bulk plugin (which we thought earlier that it would).

nils-wisiol:
> Please read the below considerations, and let me know what you think. If you still feel like the change should be made, I'll do it.

What's the benefit of changing the interface of `client.get_rr_sets` (and the other client methods) from the string domain name to domain object?

> > That way, the `__str__` method can be adapted to include more information about the domain
> 
> No, we can't adapt `__str__` because it determines the input to our `RRsetSerializer` where the `domain` field is a `StringRelatedField`, which relies on `__str__`: https://www.django-rest-framework.org/api-guide/relations/#stringrelatedfield We already are committed to this.

Same argument applies also here: I believe it is beneficial to change the relation from `StringRelatedField` to [`SlugRelatedField`](https://www.django-rest-framework.org/api-guide/relations/#slugrelatedfield) to remove the implicit dependency on the `__str__` method. There is no reason to reserve a general method like `__str__` that can play a role in many different contexts to such a specific use-case.

> > to assist in debugging (PyCharm's debugger uses `__str__` a lot to produce output.)
> 
> Are you sure it doesn't use `__repr__` (which defaults to `__str__` if undefined)?

It does use `__str__`:
Screenshot

nils-wisiol:
> Now that we have tests for bulk functionality, I added a commit for bumping DRF vom 3.7 to 3.9. Apparently, this does not conflict with the bulk plugin (which we thought earlier that it would).

I unfortunately [did not leave a note](https://github.com/desec-io/desec-stack/commit/e1db767cf063f0e352293aa7e0ee4df4dae073a0) as to why it was blocking, so all I can do right now is guessing: either it was specific to DRF 3.8, or any other requirement at that time, or I considered it blocking by mistake. Either way, it's not because DRF-bulk was updated, as the last change is still from 2015 🙈. So while it's certainly good we can upgrade DRF, I still think we should find a replacement for DRF-bulk. I stumbled across DRF's [ListSerializer](https://www.django-rest-framework.org/api-guide/serializers/#listserializer), maybe that a good starting point.

peterthomassen:
I reverted the API change in `client.get_rr_sets` etc, and also switched `RRset.domain` from `StringRelatedField` to `SlugRelatedField`.

I am not aware of any open action items regarding this PR.

nils-wisiol:
Lgtm, awesome stuff

PR #180, opened 2019-05-01
Code Style Improvements

nils-wisiol:
This commit updates the code style of the API such that the pycharm code inspection is happy on scope

    file[desec]:api//*.py&&!file[desec]:api/venv//*&&!file:api/manage.py&&!file:api/api/wsgi.py&&!file[desec]:api/desecapi/migrations//*

with the exception of 50ish stupid "typos" that are mostly test data, names defined in our dependencies or hashes, cryptographic keys, etc.

In more detail, this PR does the following.

- While fixing all pycharm warnings, I found to test bugs, fixed in e1bb69e and 154545e.
- Most of the easy work is done in a82f3ad. 
- 23dffd3 removes a future import that has no effect in python3
- In 64ac60f, an `import *` is converted in explicit imports
- In 571980c, an override of `perform_create` was made compatible with the parent's signature
- Lastly, 23feb3f makes the dependency of models on the user model `User` explicit on our own implementation. [This **breaks portability** of our Django app into other Django projects.](https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#reusable-apps-and-auth-user-model) However, it enables type guessing in pycharm, and since we are not planning to make our app reusable and and this change is easy to reverse, I thought this is a reasonable choice.

I'd like to integrate the inspection results (or something similar) in our PR checks, but couldn't find a way to do so. Any ideas?

### DIFF
--- a/api/api/settings_quick_test.py
+++ b/api/api/settings_quick_test.py
@@ -1,5 +1,6 @@
 from api.settings import *
 
+# noinspection PyUnresolvedReferences
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
--- a/api/desecapi/authentication.py
+++ b/api/desecapi/authentication.py
@@ -48,7 +48,7 @@ class BasicTokenAuthentication(BaseAuthentication):
         try:
             user, key = base64.b64decode(basic).decode(HTTP_HEADER_ENCODING).split(':')
             token = self.model.objects.get(key=key)
-        except:
+        except Exception:
             raise exceptions.AuthenticationFailed(invalid_token_message)
 
         if not token.user.is_active:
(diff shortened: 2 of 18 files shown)


nils-wisiol:
We should make a decision on this soon, because it touches so many files. Otherwise we'll end up in merge conflict hell 😅

peterthomassen:
I have a slight preference for keeping it the general way, and not charming the IDE too much. I also think that while our project most likely will not be required to integrate into other Django environments, it's still a good idea to stick to best practices of the framework. On the other hand, I'm not sure if this is at all a best practice issue.

Just my 2ct... I'll be ok with any approach you take. (also, this branch needs to be rebased)

peterthomassen:
I added two commits which split `models.py` and `views.py` into more manageable pieces. I found that this leads to a circular import between the `User` and `Token` classes since `Token` now explicitly depends on `User`, which was not the case before 23feb3fd480c5f6f377d71fce0b5c892160c973a.

We can use this as a justification to get the dependency abstract, or we can merge `User` and `Token` in the same file (which seems a bit unclean to me).

nils-wisiol:
> I have a slight preference for keeping it the general way, and not charming the IDE too much. I also think that while our project most likely will not be required to integrate into other Django environments, it's still a good idea to stick to best practices of the framework. On the other hand, I'm not sure if this is at all a best practice issue.

> Just my 2ct... I'll be ok with any approach you take. (also, this branch needs to be rebased)

I believe since we already added extra fields to the user without considering portability, we can make the dependency explicit. Any replacement for our user model would break things anyway. We can revert this change later easily, as it is only a code style issue.

> I added two commits which split models.py and views.py into more manageable pieces.

Let's move the split-commits to a branch concerned with the user management views and then reconsider splitting up, so we can finalize this PR. In preparation for merge, I rebase and remove those commits, here they are for later reference: a284380, 2c40a5d

PR #188, opened 2019-05-09
chore(api): removes code copied from parent class

nils-wisiol:
During debugging, I stumbles across this copied code and removed it.


nils-wisiol:
The method that was overridden is

- desecapi's `UserCreateView.create`, which overrode
- djoser's `UserCreateView.create`, which is inherited from 
- DRF's generic `CreateAPIView.create`, which is inherited from
- [DRF's mixin `CreateModelMixin.create`](https://github.com/encode/django-rest-framework/blob/master/rest_framework/mixins.py#L16), which is identical to the removed implementation.

PR #189, opened 2019-05-11
fix(api): give 401 for dyn if username is neither email nor domain name

peterthomassen:


api/desecapi/tests/test_authentication.py
```diff
@@ -16,15 +16,15 @@ def assertDynDNS12Status(self, status=HTTP_200_OK, authorization=None):
         self.assertStatus(self._get_dyndns12(), status)
 
     def assertDynDNS12AuthenticationStatus(self, username, token, status):
-        # Note that this overwrites self.client's credentials, which may be unexpected
```
nils-wisiol: This note was important as it is likely to change test behavior when running tests together/separately, see also   comment below

api/desecapi/tests/test_dyndns12update.py
```diff
@@ -142,6 +142,13 @@ def test_identification_by_token(self):
         self.assertEqual(response.data, 'good')
         self.assertIP(ipv4='10.5.5.6')
 
+    def test_identification_by_email(self):
+        self.client.set_credentials_basic_auth(self.owner.email, self.token.key)
```
nils-wisiol: This may break tests if executed in a different test set or different order. I propose to overwrite the client's credentials in the `setUp()` method

api/desecapi/authentication.py
```diff
@@ -48,6 +48,11 @@ def authenticate_credentials(self, basic):
         try:
             user, key = base64.b64decode(basic).decode(HTTP_HEADER_ENCODING).split(':')
             token = self.model.objects.get(key=key)
+            if '@' in user:
```
nils-wisiol: We could make the check here more restrictive and more explicit by checking that it is one of a list of allowed values. Allowed values could be: any of the user's domain names, empty string, the user's email address.

    allowed_values = ['', token.user.email] + [domain.name for domain in token.user.domains.all()]
    if user not in allowed_values:

api/desecapi/tests/test_authentication.py
peterthomassen: Ah. I only considered the comment in the context of the following line, and I did not understand why the overwriting would be "unexpected", as the next line does that very explicitly. I therefore thought the comment was leftover from the past.

I now understand that it is unexpected because it happens inside an assertion function, which has a side effect. That's unexpected. `assertDynDNS12AuthenticationStatus` is not used elsewhere, so I'll change it into a local helper function defined inside the test where it's used. This way, all assertions are free of side-effects. Tests are not, but we have the same problem with other tests, so I think as first step that's ok. (We should also make tests free of side-effects.)

api/desecapi/tests/test_dyndns12update.py
peterthomassen: As far as I can see, the `set_credentials_basic_auth` [is actually run on the client inside `setUp`](https://github.com/desec-io/desec-stack/blob/aee5777d960ee71176a05268d42a6c8ae36a501f/api/desecapi/tests/base.py#L748) of a parent class. So we're good -- or am I missing some aspect here?

api/desecapi/authentication.py
peterthomassen: neat! For domain names, we need to consider the lowercase username, so I made small adaption.

peterthomassen:
I pushed a commit with the fixes I made. Can you review again? If ok to go, I'll squash.

api/desecapi/tests/test_dyndns12update.py
nils-wisiol: ah, woops, I didn't catch that

peterthomassen:
The local part is case-sensitive, and the domain part of the email is not. https://stackoverflow.com/a/9808332/6867099 I guess we're ok as is for the moment.

PR #199, opened 2019-05-30
Getting Started Documentation for Developers

nils-wisiol:
Work in progress; this should be double-checked by someone else and possibly extended with more info.


nils-wisiol:
Known issue that needs to be fixed: when emails can't be sent, user creation will return code 503. However, the user was created and is able to login.

README.md
```diff
@@ -118,3 +113,221 @@ This stack is IPv6-capable. Caveats:
...
+        For this guide, we assume `example.dedyn.io`. Register it with:
+
+            http POST https://desec.io/api/v1/domains/ Authorization:"Token $TOKEN" name:='"example.dedyn.io"'
+
+        If you receive an answer that is different from status code 201, chances are that the name you chose is already taken by someone else.
+        In that case, repeat the last step with a new name.
+        To setup the necessary IP address records, we create a couple of A and AAAA records that point to localhost.
+        As preparation, create a JSON file `dns.json` with the following content defining the DNS setup for desec-stack:
+
+            [
```
autodiscover: get.desec subname is missing. I think it would be easier and more future-proof to just set a CNAME on `*.desec` pointing `desec.$DOMAIN.`.

autodiscover:
cool stuff!

autodiscover:
let me know when this is ready for final review

README.md
autodiscover: CNAME is not straightforward as this is the content of a file and we'd somehow have to deal with substituting `$DOMAIN` etc. So, let's keep the A/AAAA records. I still cleaned it up a bit, by using wildcards.

README.md
```diff
@@ -118,3 +113,210 @@ This stack is IPv6-capable. Caveats:
...
+       http GET https://desec.${DOMAIN}/api/v1/
+
+    Congratulations, you have desec-stack up and running.
+    Of course, as this setup is only on your local machine, DNS information will not be published into the public DNS.
+    However, the desec-stack nameserver is available on localhost port 5321.
+    To check if desec-stack is working as expected, you can query the desec-stack nameserver locally for any information that you saved using your API.
+
+       EMAIL=john@example.com
+       PASSWORD=insecure
+       http POST https://desec.${DOMAIN}/api/v1/auth/users/ email:=\"${EMAIL}\" password:=\"${PASSWORD}\"
```
autodiscover: This does not work if you don't set the email environment variables. Hm. Not sure what the best solution is?

README.md
```diff
@@ -118,3 +113,210 @@ This stack is IPv6-capable. Caveats:
...
+        First, the environment variables as defined in the `.env` file need to be made available in the shell.
+        This can be done with
+
+           set -a && source ../.env && set +a
+
+        Second, the API needs to be configured to use a local database instead of the dbapi host.
+        (dbapi, of course, is unavailable outside the docker-compose application.)
+        To use a different database, use the `settings_quick_test.py` instead of the default `settings.py`:
+
+           DJANGO_SETTINGS_MODULE=api.settings_quick_test
```
autodiscover: Where do I have to enter this?

README.md
```diff
@@ -118,3 +113,210 @@ This stack is IPv6-capable. Caveats:
...
+
+        Finally, you can manage Django using the `manage.py` CLI.
+        As an example, to run the tests, use
+
+           python3 manage.py test
+
+        The tests should take about 10 seconds.
+
+    1. Open PyCharm in the project root directory and open the settings.
```
autodiscover: I read the following but did not try it as my previous comment may be a blocker.

README.md
nils-wisiol: fixed by b75a0579fdea9cd168d8f2b0b028308841e0019b

README.md
nils-wisiol: the environment variable `DJANGO_SETTINGS_MODULE` needs to be set to this value. Not sure how to phrase that?

README.md
nils-wisiol: pycharm needs to be configured to have the appropriate environment variables, just like the shell in the preceding example

README.md
autodiscover: ah! I thought this was Python code, but it's Shell code...

PR #204, opened 2019-06-07
feat(): use init system inside Docker containers, fixes #113

### LINKED ISSUE #113: Use init system inside Docker containers
Rationale: https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html

`docker-compose` now provides an option for this: https://docs.docker.com/compose/compose-file/compose-file-v2/#init

autodiscover:
Rationale: https://engineeringblog.yelp.com/2016/01/dumb-init-an-init-for-docker.html

### DIFF
--- a/api/entrypoint.sh
+++ b/api/entrypoint.sh
@@ -5,10 +5,13 @@ echo "waiting for dependencies ..."
 ./wait
 
 # start cron
-/root/cronhook/start-cron.sh &
+# Start child process that starts grand-child process.
+# After the child process's death, the grand-child will be adopted by init.
+# See https://stackoverflow.com/a/20338327
+( /root/cronhook/start-cron.sh & )
 
 # migrate database
 python manage.py migrate || exit 1
 
 echo Finished migrations, starting API server ...
-uwsgi --ini uwsgi.ini
+exec uwsgi --ini uwsgi.ini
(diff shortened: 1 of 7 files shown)


nils-wisiol:
Should we also change our `entrypoint.sh` scripts to `exec` stuff instead of just running it as a sub-process? (nginx, pdns comes to mind.) That's recommended in the blog post that you reffered to.

autodiscover:
- I added `exec` for `nsmaster` and `nslord`.
- On `www`, we can't replace the `entrypoint.sh` process as it is still required to stick around in order to coordinate reloading certificates once they change.
- On `api`, adding `exec` to the `uwsgi` command does work. However, before we call `uwsgi`, we also run a helper script (`start-cron.sh`) to start the cron daemon. When using `exec` on `uwsgi`, we end up with a process tree like this:
```
init─┬─cron
     └─uwsgi─┬─start-cron.sh───tail
             ├─uwsgi
             ├─63*[uwsgi───3*[{uwsgi}]]
             └─3*[{uwsgi}]
```
That is, `start-cron.sh` becomes a child of `uwsgi` (which is the process replacing the script process that called `start-cron.sh`). I'm not sure what the implications of this are, so I'd rather not do it.

nils-wisiol:
I propose a way to have uwsgi, cron, and cron-start.sh be all direct children of init in api. Otherwise, lgtm.

PR #212, opened 2019-06-21
fix(nslord): disable master mode (brought hopeless NOTIFY attempts only)

peterthomassen:
Our logs have lots of unsuccessful NOTIFY attempts because master mode was enabled on nslord. We don't actually use NOTIFY anymore, so master mode can be disabled. (Allowing AXFR queries from nsmaster is completely independent and continues to work.)


nils-wisiol:
lgtm

PR #218, opened 2019-06-29
Collect Information to Debug Failing Dyndns12-Updates

nils-wisiol:
1. Adds nicer string representations to change tracker objects.
2. Lets the change tracker raise a ValueError when it encounters a DoesNotExist exception for RRset. The view is prepared to add all input information for the serializer to the output and raises another ValueError. Final output will look somewhat like this:

		Error message: For changes ["Update RRsets of 88737777.dedyn.io: additions=[], modifications=[('A', '')], deletions=[]"], could not find RRset when applying Update RRsets of 88737777.dedyn.io: additions=[], modifications=[('A', '')], deletions=[]
		RRset objects: [<RRset: <RRSet domain=88737777.dedyn.io type=A subname=>>]
		RR objects: {'A': [<RR: <RR 172.16.0.1>>]}
		Data: [{'type': 'A', 'subname': '', 'ttl': 60, 'records': ['1.1.1.1']}, {'type': 'AAAA', 'subname': '', 'ttl': 60, 'records': []}]

PR #219, opened 2019-07-01
Do Not Run API Multi-Threaded

nils-wisiol:
**TL;DR: the change tracker is not thread-safe, so don't run it multi-threaded.**

The change tracker relies on Django signals to track model changes that need to be propagated to pdns. Django signals are shared across multi-threading, however not across multi-processing.

The state before this PR is to spawn 64 python processes with four threads each. The state after will be to spawn 64 python processes. This has the additional benefit that we can't exhaust the dbapi connection limit anymore, which is currently at it's [default of 151 connections](https://img-9gag-fun.9cache.com/photo/aVYz0yw_700bwp.webp).

As each worker only receives one request at a time, this fixes the bug caused by multi-threading. When using exclusively multi-processing, signals are separated by request.

(Rationale to prefer this over separation by, say, user: sometimes we want the change tracker to track across user boundaries, e.g. when syncing back from pdns to the API. It's also conceivable to let users administer several domains in parallel, e.g. in an automated setting for ISPs.)

The effectiveness of this PR was verified with the deSEC aquarium, a testing script that can run multiple requests in parallel, that will likely be content of a future PR.

This PR does not fix some other uncaught exceptions that occur when the API is massively queried in parallel, but as far as I can tell, this PR fixes all issues caused by massively parallel requests *of different users*.

(This 2-bit fix will likely my smallest PR for a while.)

### DIFF
--- a/api/uwsgi.ini
+++ b/api/uwsgi.ini
@@ -3,7 +3,7 @@ socket = [::]:3031
 chdir = /usr/src/app/
 wsgi-file = api/wsgi.py
 processes = 64
-threads = 4
+threads = 1
 #stats = 127.0.0.1:9191
 
 #master = true


peterthomassen:
- So, #218 is obsolete?
- Maybe add a comment in the change tracker code, referring to the commit of this PR and the thread-unsafety?
- It may still be good to catch the 1040 MySQL error from https://github.com/desec-io/desec-stack/pull/218/commits/ac9a96b5e7a039c378e148ee0e7b95a8ac09be58, to be future-proof.
- Cherry-pick https://github.com/desec-io/desec-stack/pull/218/commits/4fa6b93d0e3230f877c1f4d2bc10687e6c0c4c6a?

nils-wisiol:
- Yes, #218 is obsolete, but parts of it may occur in a future PR that treats concurrency issues.
- I added a comment that will eventually make it into some PR.
- Same for ac9a96b
- Same for 4fa6b93

PR #222, opened 2019-07-02
Introduce deSEC User Management

peterthomassen:


peterthomassen:
@nils-wisiol This is ready for review. While not mature enough to be merged, there is nothing of substance that I am planning to change.

Open items:

- maybe move signing code around, does not appear to be in the best locations yet
- solve failing e2e tests (need to find a way to work around accout activation)
- devise and implement invitation code scheme

nils-wisiol:
I tried something new today and used pycharm for my code review instead of GitHub. It makes code navigation much easier, but now my comments are all in the code, not in GitHub comments. If that's hard to work with on your end, I may just revert to the old style of review annotations.

My annotations are roughly three-fold:

1. The signing/verification does not take full advantage of the Django framework. In particular, the complex object signing and signature expiry are both already implemented in Django. I couldn't find a reason to not use them, so if there is none, let's move to the Django interface.
In any case, there may be some weaknesses in the complex object singing code that should be discussed if you decide to keep `get_dict_repr`.
2. IMHO, separation of concerns is in some places violated. It's the reoccurring question of where we draw the line in between validation and authentication/permission. I think a good rule of thumb may be to see if an admin action should be able to do the same action: anything that's in*valid*, and admin should not be able to do. Anything we deny based on permission, an admin should be able to do. Therefore, I believe the password check currently conducted in the serializers should be moved to an authentication class.
3. Finally, I think some code duplication and some code cleanliness can be achieved by introducing some inheritance structure in the views and serializers. While the benefit towards code duplication is obvious with inheritance, I think it could also increase cleanliness and readability by removing a couple of the larger if-elif-statements that differentiate treatment by validation action.

I did not yet review changes to Django test and e2e. 

Overall, I think this is a big step in the right direction for our user management. Thanks!

api/desecapi/models.py
```diff
@@ -369,7 +372,7 @@ def __str__(self):
         return '<RR %s>' % self.content
 
 
-def authenticated_action_default_timestamp():
+def timestamp_now():
```
peterthomassen: remove in favor of `django.utils.timezone.now`?

api/desecapi/serializers.py
```diff
@@ -19,6 +20,23 @@
 psl = psl_dns.PSL(resolver=settings.PSL_RESOLVER)
 
 
+class ImageCaptchaSerializer(serializers.ModelSerializer):
+    id = ModelField(  # make id field avail. in validated_data
+        model_field=models.ImageCaptcha()._meta.get_field('id'),
```
peterthomassen: Does it not suffice to put `id` in `Meta.fields`?

api/desecapi/serializers.py
```diff
@@ -19,6 +20,23 @@
...
+        required=False,
+    )
+    challenge = serializers.SerializerMethodField()
+    solution = serializers.CharField(write_only=True, required=False)
```
peterthomassen: length check missing

api/desecapi/tests/test_captcha.py
```diff
@@ -0,0 +1,86 @@
+from io import BytesIO
+
+from PIL import Image
```
peterthomassen: Is this the Python Image Library from 2009, without support for Python 3? :-o

api/desecapi/serializers.py
```diff
@@ -19,6 +20,23 @@
 psl = psl_dns.PSL(resolver=settings.PSL_RESOLVER)
 
 
+class ImageCaptchaSerializer(serializers.ModelSerializer):
```
peterthomassen: breaks (produces 500) if `POST` has `solution` payload

api/desecapi/serializers.py
peterthomassen: refactored

api/desecapi/serializers.py
peterthomassen: future-proof -- why not

api/desecapi/serializers.py
peterthomassen: fixed by refactoring

api/desecapi/models.py
```diff
@@ -313,3 +369,205 @@ class RR(models.Model):
...
+    class Meta:
+        managed = False
+
+    def __init__(self, *args, **kwargs):
+        # silently ignore any value supplied for the mac value, that makes it easier to use with DRF serializers
+        kwargs.pop('mac', None)
+        super().__init__(*args, **kwargs)
+
+    @property
+    def name(self):
```
peterthomassen: Why is this a property and not an attribute?

Also, it would shadow any payload of the same name, right? (Should it be .Meta.name?)

api/desecapi/serializers.py
```diff
@@ -458,25 +472,23 @@ def validate_name(self, value):
...
-        assert is_public_suffix or value == private_domains[0]
+        assert is_public_suffix or domain_name == private_domains[0]
 
         # Deny registration for non-local public suffixes and for domains covered by other users' zones
-        owner = self.context['request'].user
-        queryset = models.Domain.objects.filter(Q(name__in=private_domains) & ~Q(owner=owner))
-        if is_restricted_suffix or queryset.exists():
-            msg = 'This domain name is unavailable.'
-            raise serializers.ValidationError(msg, code='name_unavailable')
+        user = user if not isinstance(user, AnonymousUser) else None
```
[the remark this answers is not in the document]
peterthomassen: my remark is not true

api/desecapi/views.py
```diff
@@ -214,17 +214,21 @@ def perform_update(self, serializer):
...
     def get(self, request, *_):
-        # TODO update
         if self.request.user.is_authenticated:
             routes = {
-                'account': reverse('account', request=request),  # TODO update this in particular
+                'account': {
+                    'show': reverse('account', request=request),
+                    'delete': reverse('account-delete', request=request),
+                    'change-email': reverse('account-change-email', request=request),
+                },
```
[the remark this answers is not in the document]
nils-wisiol: yes

api/desecapi/migrations/0006_auto_20190921_1139.py
```diff
@@ -0,0 +1,83 @@
...
+            ],
+            options={
+                'managed': False,
+            },
+        ),
+        migrations.AlterField(
+            model_name='domain',
+            name='minimum_ttl',
+            field=models.PositiveIntegerField(default=1),
+        ),
```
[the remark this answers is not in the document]
nils-wisiol: yes

api/desecapi/models.py
```diff
@@ -77,7 +76,6 @@ class User(AbstractBaseUser):
     )
     is_active = models.BooleanField(default=True)
     is_admin = models.BooleanField(default=False)
-    registration_remote_ip = models.CharField(max_length=1024, blank=True)
```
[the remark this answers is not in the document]
nils-wisiol: it's unused information

api/desecapi/models.py
```diff
@@ -313,3 +363,178 @@ class RR(models.Model):
...
+
+    def __init__(self, *args, **kwargs):
+        # silently ignore any value supplied for the mac value, that makes it easier to use with DRF serializers
+        kwargs.pop('mac', None)
+        super().__init__(*args, **kwargs)
+
+    @property
+    def name(self):
+        """
+        Returns a human-readable string containing the name of this action class that uniquely identifies this action.
```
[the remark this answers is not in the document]
peterthomassen: switch to class path

api/desecapi/models.py
peterthomassen: It's a property so that it can be an "abstract attribute". Also, yes, shadows payload of the same name, but that's not a problem now.

api/desecapi/models.py
```diff
@@ -313,3 +363,178 @@ class RR(models.Model):
...
+
+
+class AuthenticatedUserAction(AuthenticatedAction):
+    """
+    Abstract AuthenticatedAction involving an user instance, incorporating the user's id, email, password, and
+    is_active flag into the Message Authentication Code state.
+    """
+    user = models.ForeignKey(User, on_delete=models.DO_NOTHING)
+
+    class Meta:
```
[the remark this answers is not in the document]
peterthomassen: yes

api/desecapi/tests/test_captcha.py
peterthomassen: fine for test. Captcha library uses it also, but no external input involved

peterthomassen:
lgtm now!

api/desecapi/serializers.py
```diff
@@ -490,16 +490,22 @@ def raise_if_domain_unavailable(domain_name, user):
...
-            # public suffix and setting both public_suffix and is_public_suffix accordingly).
-            # However, in order to allow to investigate the situation, it's better not catch
-            # this exception. Our error handler turns it into a 503 error and makes sure
-            # admins are notified.
-            raise e
+        # noinspection PyCallByClass
+        _, parent = models.Domain.partition_name(domain_name)
+        if parent in settings.LOCAL_PUBLIC_SUFFIXES:
+            public_suffix = parent
+            is_public_suffix = False
```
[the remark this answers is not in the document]
peterthomassen: 71149be037c359532a74a93ef6edcb5effab7987 has a proposal to cover this case

nils-wisiol:
There is a minor gap in the solution proposed in 71149be that I discovered when I added tests for the PSL policy: https://github.com/desec-io/desec-stack/pull/222/commits/6240c5c4c16c21d5df7f083f9ea6e923a5a32e1f#diff-408d3949827d55c0862420667d31ddfaR100-R101

Any thoughts how to fix this? (Should it be fixed?)

peterthomassen:
@nils-wisiol fixed in https://github.com/desec-io/desec-stack/pull/222/commits/75677ec34656dfc9239359826a7a5594d833a481

nils-wisiol:
lgtm!

nils-wisiol:
Let's clean the history a bit and then merge 😍

peterthomassen:
done. As I started the PR, you have to approve it before we can merge. Can you do so?

PR #231, opened 2019-07-07
Fix validation mistakes

peterthomassen:


docs/rrsets.rst
```diff
@@ -272,8 +272,9 @@ Modifying an RRset
...
 specific RRset.  The API allows changing the values of ``records`` and
 ``ttl``.  When using the ``PATCH`` method, only fields you would like to modify
-need to be provided, where the ``PUT`` method requires specification of both
-fields.  Examples::
+need to be provided.  In contrast, if you use ``PUT``, the full resource must
+be specified (that is, all fields, including ``subname`` and ``type``).
+Examples::
 
     http PUT \
         https://desec.io/api/v1/domains/:name/rrsets/:subname/:type/ \
```
nils-wisiol: example needs updating.

Also: what if the two subnames (types) provided do not agree?

api/desecapi/tests/test_rrsets_bulk.py
```diff
@@ -166,7 +166,7 @@ def test_bulk_patch_fresh_rrsets_need_type(self):
 
     def test_bulk_patch_does_not_accept_single_objects(self):
         response = self.client.bulk_patch_rr_sets(domain_name=self.my_empty_domain.name, payload=self.data[0])
-        self.assertContains(response, 'expected a list of RRsets.', status_code=status.HTTP_400_BAD_REQUEST)
+        self.assertContains(response, 'Expected a list of items but got type ', status_code=status.HTTP_400_BAD_REQUEST)
```
nils-wisiol: avoid the word `type`, as it could be confused with the resource record type

docs/rrsets.rst
peterthomassen: Example updated.

If subnames or types disagree, `400` is returned. Test for that is `test_update_essential_properties`.

api/desecapi/tests/test_rrsets_bulk.py
peterthomassen: fixed

PR #236, opened 2019-07-19
20190718 minimum ttl

peterthomassen:


.env.default
```diff
@@ -22,6 +22,7 @@ DESECSTACK_API_EMAIL_PORT=
 DESECSTACK_API_SECRETKEY=
 DESECSTACK_API_PSL_RESOLVER=
 DESECSTACK_DBAPI_PASSWORD_desec=
+DESECSTACK_MINIMUM_TTL_DEFAULT=
```
nils-wisiol: a default default value would be appropriate here, maybe 300?

peterthomassen:
both done!

PR #237, opened 2019-07-19
20190719 dyndns check script

peterthomassen:


www/conf/sites-available/05-dedyn.conf.var
```diff
@@ -27,6 +27,11 @@ server {
 		add_header Strict-Transport-Security "max-age=31536000" always;
 		return 301 https://desec.$DESECSTACK_DOMAIN$request_uri;
 	}
+
+        location =/check {
+		root /usr/share/nginx/html;
```
nils-wisiol: STS header also here?

nils-wisiol: doesn't have the first matching location preference?

peterthomassen: STS: good point! fixed.

`location` order: The "longest" one is used, according to the docs. Not sure if that's the one with the most path components, or the one with the most bytes, but in any case, it works here as it should.

PR #240, opened 2019-07-19
20190719 rrsets pagination

peterthomassen:
Functionality is there. What's missing:

- tests
- docs

### DIFF
--- a/api/desecapi/serializers.py
+++ b/api/desecapi/serializers.py
@@ -204,7 +204,7 @@ class RRsetSerializer(ConditionalExistenceModelSerializer):
 
     class Meta:
         model = models.RRset
-        fields = ('domain', 'subname', 'name', 'records', 'ttl', 'type',)
+        fields = ('created', 'domain', 'subname', 'name', 'records', 'ttl', 'type',)
         extra_kwargs = {
             'subname': {'required': False, 'default': NonBulkOnlyDefault('')}
         }
--- a/api/desecapi/tests/test_rrsets.py
+++ b/api/desecapi/tests/test_rrsets.py
@@ -172,7 +172,7 @@ class AuthenticatedRRSetTestCase(AuthenticatedRRSetBaseTestCase):
         response = self.client.get_rr_set(self.my_rr_set_domain.name, subname='', type_='A')
         self.assertStatus(response, status.HTTP_200_OK)
         self.assertEqual(response.data['records'][0], '1.2.3.4')
-        self.assertEqual(response.data['ttl'], 120)
+        self.assertEqual(response.data['ttl'], 3620)
 
     def test_retrieve_my_rr_sets_restricted_types(self):
         for type_ in self.RESTRICTED_TYPES:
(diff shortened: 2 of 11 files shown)


peterthomassen:
Tests and documentation added.

To finalize this, I would like to discuss the appropriate response status code for requests that would require pagination, but don't use a `cursor` parameter. Currently, we produce `400` (and instruct how to use pagination). I wonder if there is a more specific code, such as

- `300 Multiple Choices`, but that's intended for varied versions of the same thing (e.g. languages)
- `403 Forbidden`, but very misleading
- `411 Length Required`, but refers to `Content-Length`
- `412 Precondition Failed` -- something with header conditions, but I don't fully understand
- `413 Request Entity Too Large`, but actually refers to the request payload, not the response
- `416 Requested Range Not Satisfiable`, but actually refers to `Range` headers

... any ideas?

peterthomassen:
Agreed with @nils-wisiol that status 400 is good enough.

No more open issues, ready for review.

api/desecapi/tests/test_rrsets.py
```diff
@@ -42,15 +44,90 @@ def test_retrieve_my_rr_sets(self):
...
     def test_retrieve_other_rr_sets(self):
         self.assertStatus(self.client.get_rr_sets(self.other_domain.name), status.HTTP_404_NOT_FOUND)
         self.assertStatus(self.client.get_rr_sets(self.other_domain.name, subname='test'), status.HTTP_404_NOT_FOUND)
         self.assertStatus(self.client.get_rr_sets(self.other_domain.name, type='A'), status.HTTP_404_NOT_FOUND)
 
     def test_retrieve_my_rr_sets_filter(self):
-        response = self.client.get_rr_sets(self.my_rr_set_domain.name)
+        response = self.client.get_rr_sets(self.my_rr_set_domain.name, query='?cursor=')
         self.assertStatus(response, status.HTTP_200_OK)
-        self.assertEqual(len(response.data), len(self._test_rr_sets()))
```
nils-wisiol: can we assert that length is `min(len(self...), page_size_setting)`?

api/desecapi/tests/test_rrsets.py
```diff
@@ -42,15 +44,90 @@ def test_retrieve_my_rr_sets(self):
...
+                mapping[label] = url
+            return mapping
+
+        # Prepare extra records so that we get three pages (total: n + 1)
+        n = int(settings.REST_FRAMEWORK['PAGE_SIZE'] * 2.5)
+        RRset.objects.bulk_create(
+            [RRset(domain=self.my_domain, subname=str(i), ttl=123, type='A') for i in range(n)]
+        )
+
+        # No pagination
```
nils-wisiol: The following five blocks of code should be refactored into some sub-routine to make maintenance easier.

api/desecapi/tests/test_rrsets.py
```diff
@@ -344,6 +344,16 @@ def test_update_essential_properties(self):
...
 
+        # Changing "created" is no-op
+        response = self.client.get(url)
+        data = response.data
+        created = data['created']
+        data['created'] = '2019-07-19T17:22:49.575717Z'
+        response = self.client.patch(url, data)
+        self.assertStatus(response, status.HTTP_200_OK)
+        response = self.client.patch(url, data)
+        self.assertStatus(response, status.HTTP_200_OK)
```
nils-wisiol: Why patching twice?

api/desecapi/serializers.py
```diff
@@ -172,7 +172,7 @@ class RRsetSerializer(ConditionalExistenceModelSerializer):
 
     class Meta:
         model = RRset
-        fields = ('domain', 'subname', 'name', 'records', 'ttl', 'type',)
+        fields = ('created', 'domain', 'subname', 'name', 'records', 'ttl', 'type',)
```
nils-wisiol: This should also be added to the e2e schema test in `test/e2e/schemas.js`

docs/domains.rst
```diff
@@ -141,10 +141,15 @@ The ``/api/v1/domains/`` endpoint reponds to ``GET`` requests with an array of
     curl -X GET https://desec.io/api/v1/domains/ \
         --header "Authorization: Token {token}"
 
-to retrieve an overview of the domains you own.
+to retrieve an overview of the domains you own.  Domains are returned in
+reverse chronological order of their creation.
```
nils-wisiol: Does our code actually guarantee this? I am not aware of any sorting.

docs/rrsets.rst
```diff
@@ -198,11 +198,32 @@ command::
     curl -X GET https://desec.io/api/v1/domains/:name/rrsets/ \
         --header "Authorization: Token {token}"
 
-to retrieve the contents of a zone that you own.
+to retrieve the contents of a zone that you own.  RRsets are returned in
+reverse chronological order of their creation.
```
nils-wisiol: Same as above

docs/domains.rst
peterthomassen: This is a new side-effect feature: The paginator paginates based on the created timestamp.

docs/rrsets.rst
peterthomassen: This is a new side-effect feature: The paginator paginates based on the created timestamp.

api/desecapi/tests/test_rrsets.py
peterthomassen: I guess one should have been PUT, shall change

api/desecapi/serializers.py
peterthomassen: good catch! done

api/desecapi/tests/test_rrsets.py
peterthomassen: sure! good solution, done

api/desecapi/tests/test_rrsets.py
peterthomassen: done! looks much better now, thx for the suggestion

peterthomassen:
> I am not sure what's the intention of [6e42035](https://github.com/desec-io/desec-stack/commit/6e42035ea01cc287e8b10aa40c0c44832399735e):
> 
> * it moves the default value of the `minimum_ttl` field, but as far as I can tell, this does not have any effect -- it's now just defined elsewhere. Could you enlighten me? blush

The intention was to solve the ghost migration issue that arouse when the value of `settings.MINIMUM_TTL_DEFAULT` changed. We solved it in the meantime with 0f1ed0ed523d87ad167cae831bb5cc7dc0f49485 (merged already), so I removed the fix from this PR.

> * it changes the test to `Domain.get` instead of `Domain.get_or_create`, which I confirmed is unnecessary to pass the tests.

That's true, but I figured the tests be as specific as possible. As I noticed that all tests create their domains before using them, I thought `get` would be the more clear route.

> Why does [6a849ea](https://github.com/desec-io/desec-stack/commit/6a849eaa1a96212d7835d3e52b0e47a36a2af677) increase the TTL?

Because the tests failed in my test environment which had the minimum TTL set to 3600 instead of 60. While I don't think that 3600 in particular needs to be supported, I figured that there's no point in overly low TTL values, just to decrease the likelihood of running into this kind of problem again in the future. (Think of someone else working on this, using a minimum TTL of 900, for example). That's not a perfect approach, of course, but I thought it would help a bit with test resilience.

I took all code comments into account, fixed things up, rebased, and I think we're good to go. Can you pls take another look? :blush:

PR #258, opened 2019-10-30
feat(api): migrate User ID datatype from sequential int to UUID

peterthomassen:


peterthomassen:
@nils-wisiol I now managed to write and committed a migration that switches `User.id` from an auto-incrementing integer to a UUID, and also repopulates that field with new random UUID values for all users, cascading the change down to the `Domain.owner` and `Token.user` foreign keys. Writing it took forever, which was mostly due to the fact that an intermediate field type is necessary, and also because the order of migration operations matters even if you wouldn't think it does. But now it is done!

The following issues need to be fixed / decided:
- Creation of authenticated action codes fails because UUID is not serializable. I think I can work that out.
- Unit tests fail because the migration doesn't run properly in sqlite. I'll look into it, but I may need some help.
- Do we want the migration to be "as much reversible as possible"? It cannot be completely reversible (because casting the UUID database column back to an int column will always fail, unless we come up with even more code to assign sequential IDs again). In my opinion, that ruins reversibility completely, and we should not include reverse routines for any of the steps whatsoever (e.g. for altering foreign key). They are currently in the commit, but I'd like to propose to drop those. What do oyu think?

peterthomassen:
This now runs fine on top of a database that has recent migrations applied, but fails on a fresh database. Workaround: Run migrations twice (second run will finish the missing one(s)). We suspect that this is a Django bug and reported it here: https://code.djangoproject.com/ticket/30966

We're now going to wait for feedback on the bug report before proceeding with this PR.

peterthomassen:
I think we can continue with this. The Django people have confirmed that the migration issue is their bug and only occurs when running certain migrations together. Rerunning will fix the issue.

Also, in our prod environment, the problem does not occur because the last migration is already applied.

All in all, I don't think we need to wait for the Django bugfix (but opinions may differ). What do you think?

PR #263, opened 2019-11-08
20191108 encrypt confirmation codes

peterthomassen:
Pretty straightforward. Three questions:

- Should we put `SECRET_KEY_SIGN` and `SECRET_KEY_ENCRYPT` in a central place? They are currently in the files where they're needed.
- Should the key derivation be done differently? There's [serious stuff](https://cryptography.io/en/latest/hazmat/primitives/key-derivation-functions/) out there, but it looks complicated. Let's figure out what's an acceptable approach.
- What's the right way to test this (e.g. that different keys are really being used)?

### DIFF
--- a/api/desecapi/crypto.py
+++ b/api/desecapi/crypto.py
@@ -1,0 +1,37 @@
+from base64 import urlsafe_b64encode
+from enum import Enum
+
+from cryptography.fernet import Fernet, InvalidToken
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.kdf.kbkdf import CounterLocation, KBKDFHMAC, Mode
+from cryptography.hazmat.backends import default_backend
+from django.conf import settings
+from django.utils.encoding import force_bytes
+
+
+def _derive_urlsafe_key(*, label, context):
+    backend = default_backend()
+    kdf = KBKDFHMAC(algorithm=hashes.SHA256(), mode=Mode.CounterMode, length=32, rlen=4, llen=4,
+                    location=CounterLocation.BeforeFixed, label=label, context=context, fixed=None, backend=backend)
+    key = kdf.derive(settings.SECRET_KEY.encode())
+    return urlsafe_b64encode(key)
+
+
+def retrieve_key(*, label, context):
+    # Keeping this function separate from key derivation gives us freedom to implement look-ups later, e.g. from cache
+    label = force_bytes(label, strings_only=True)
+    context = force_bytes(context, strings_only=True)
+    return _derive_urlsafe_key(label=label, context=context)
+
+
+def encrypt(data, *, context):
+    key = retrieve_key(label=b'crypt', context=context)
+    return Fernet(key=key).encrypt(data)
+
+
+def decrypt(token, *, context, ttl=None):
+    key = retrieve_key(label=b'crypt', context=context)
+    try:
+        return Fernet(key=key).decrypt(token, ttl=ttl)
+    except InvalidToken:
+        raise ValueError
--- a/api/desecapi/models.py
+++ b/api/desecapi/models.py
@@ -8,7 +8,8 @@ import string
 import time
 import uuid
 from base64 import urlsafe_b64encode
-from datetime import datetime, timedelta
+from datetime import timedelta
+from hashlib import sha256
 from os import urandom
 
 import psl_dns
@@ -18,13 +19,11 @@ from django.contrib.auth.hashers import make_password
 from django.contrib.auth.models import BaseUserManager, AbstractBaseUser, AnonymousUser
 from django.core.exceptions import ValidationError
 from django.core.mail import EmailMessage, get_connection
-from django.core.signing import Signer
 from django.core.validators import RegexValidator
 from django.db import models
 from django.db.models import Manager, Q
 from django.template.loader import get_template
 from django.utils import timezone
-from django.utils.crypto import constant_time_compare
 from rest_framework.exceptions import APIException
 
 from desecapi import pdns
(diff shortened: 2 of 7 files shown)


api/desecapi/models.py
```diff
@@ -182,7 +182,7 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)  # Call the "real" save() method.
 
     def generate_key(self):
-        return b64encode(urandom(21)).decode('utf-8').replace('/', '-').replace('=', '_').replace('+', '.')
+        return urlsafe_b64encode(urandom(21)).decode()
```
nils-wisiol: check that docs are still valid

api/desecapi/serializers.py
```diff
@@ -16,6 +18,10 @@
 from desecapi import models
 
 
+SECRET_KEY_ENCRYPT = urlsafe_b64encode(sha256(('encrypt-' + settings.SECRET_KEY).encode()).digest())  # DO NOT MODIFY
```
nils-wisiol: duplicate code

nils-wisiol: use serious kdf, https://csrc.nist.gov/publications/detail/sp/800-108/final

peterthomassen: I read SP 800-108 and then worked out two ways to do this, see https://github.com/desec-io/desec-stack/pull/263/commits/6bd669eb48a0bc7ebd07fcebb2da665f523a6678.

As the next step, I'd like to discuss them with you in person and point out some questions I have.

api/desecapi/crypto.py
```diff
@@ -1,10 +1,24 @@
...
+
+def encrypt(data):
+    return fernet.encrypt(data)
+
+
+def decrypt(token):
+    try:
+        return fernet.decrypt(token)
+    except InvalidToken:
+        raise ValueError
```
nils-wisiol: remove try/except: it doesn't do anything but changing the exception type

peterthomassen: The intention was that the code that uses this and wants to catch the exception (serializers.py) does not need to import `from cryptography.fernet import InvalidToken` in order to return a sensible error. Does that make sense?

peterthomassen:
> how do we deal with the existing tokens? they will not work any longer, interrupting the sign up process. (Oh, I see.)

Verification codes expire after 12 hours. As we currently issue them only sporadically, I think it would be easiest to deploy this at a point in time where we have not issued tokens for the last 12 hours (generally some time before noon or something like that).

peterthomassen:
Worked out the agreed-upon KDF mechanism. Does this look good to you? If so, let me add some tests and then we should be ready to go

peterthomassen:
Added tests in 78cfd4374da746a8040b7a8171ffc5861d9d0415 and 4f9479c39193d7a30d13b335c8785d425315f011.

Also fixed an unrelated shadowed test in 4532b96d7f5286f6a129b8256ec62d640a3f065d

From my perspective, this is ready to ship

peterthomassen:
Refactored things so that we now do encrypt-then-authenticate. Both is done by Python cryptography. It turns out that this simplifies a bunch of our code.

ready from my side!

nils-wisiol:
this lgtm!

@franziskuskiefer, you opinion on `crypto.py`, `test_crypto.py` as introduced in this PR is highly appreciated :)

api/desecapi/crypto.py
```diff
@@ -0,0 +1,37 @@
...
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.kdf.kbkdf import CounterLocation, KBKDFHMAC, Mode
+from cryptography.hazmat.backends import default_backend
+from django.conf import settings
+from django.utils.encoding import force_bytes
+
+
+def _derive_urlsafe_key(*, label, context):
+    backend = default_backend()
+    kdf = KBKDFHMAC(algorithm=hashes.SHA256(), mode=Mode.CounterMode, length=32, rlen=4, llen=4,
```
franziskuskiefer: From the docs I see that this can fail already. Probably won't happen with all the defaults used here, but if it does, how do you handle that?

franziskuskiefer: Hard-coding things like key-length might lead to trouble later on if you want to change it. Why do you need 32 byte? Could this change?

api/desecapi/tests/test_crypto.py
```diff
@@ -0,0 +1,56 @@
...
+class CryptoTestCase(TestCase):
+    context = 'desecapi.tests.test_crypto'
+
+    def test_retrieved_key_is_reproducible(self):
+        keys = (crypto.retrieve_key(label='test', context=self.context) for _ in range(2))
+        self.assertEqual(*keys)
+
+    def test_retrieved_key_depends_on_secret(self):
+        keys = []
+        for secret in ['abcdefgh', 'hgfedcba']:
```
franziskuskiefer: Is the real secret you use longer than this? 😄  HKDF probably shouldn't be used for values this small.

api/desecapi/crypto.py
peterthomassen: Currently, uncaught exceptions are turned into 500 errors which in turn leads to an email getting sent to us with details, so that we can track down what happened. So far this has served us well in finding unforeseen edge cases.

If this would happen here, it would certainly be a server-side error (5xx), and we would certainly want to know about it (and not silence it), so I figured this approach is fine. However, I understand that it does not appear to be very mindful. Do you have another suggestion (also, generically for other such cases)?

peterthomassen: Currently, we use the key exclusively with the [Fernet crypto implementation](https://cryptography.io/en/latest/fernet/#cryptography.fernet.Fernet) which requires a URL-safe base64-encoded 32-byte key. If it ever changes, we can certainly add another kwarg here.

Also, things produced with this key are not persisted at the moment -- the use case is verification codes in things like password reset emails. Those links have a validity window of 12 windows. If all else fails, we could stop issuing such links for 12 hours and perform whatever migration we like (maybe use a shorter expiration window temporarily).

api/desecapi/tests/test_crypto.py
peterthomassen: It is provided to the Docker container as an environment variable, so this code does not really care about it. But yes, our secret is much longer! :)

peterthomassen:
@franziskuskiefer Thanks a lot for the review. I had one remark regarding your error handling question, but overall I'm glad that you didn't find any obvious mess-ups!

api/desecapi/crypto.py
franziskuskiefer: Server errors seem reasonable 🙂 As long as you know that it can fail and know what happens everything's good 👍

PR #264, opened 2019-11-11
20191111 token refactoring

peterthomassen:
This PR changes token handling such that token values are stored in hashed format. For now, it is PBKDF2-HMAC-SHA256 with one iteration and a static salt.

Rationale:

- We should not store secrets in the clear, including tokens. What we don't have in plaintext, we can't loose. Potential loss vectors can be: a) backup compromise, b) API compromise (e.g. through code injection vulnerability), c) dbapi container compromise via MariaDB vulnerability, d) coding mistake in view authentication so that we return tokens we don't want to return, e) deSEC staff accessing tokens directly in the database, f) probably more things I forgot.
- It's also a good practice to not show secrets to the user multiple times, so that somebody accessing the user's account cannot quickly list API tokens. (Amazon AWS takes this approach, for example.) However, if we show secrets only once, there's no reason for being able to access them. [Btw, this makes me think that maybe we should send out a notification email when a token is created?]
- There is no practical timing penalty for this implementation: A quick survey using `timeit` showed that about 250,000 tokens can be hashed on my laptop per second, sequentially. Most likely, each API request comes with other much more expensive operations (e.g. connecting to the database). In another test, running our Django test suite on master vs. this branch three times each took (9.032s, 8.941s, 9.063s) vs. (9.173s, 9.014s, 8.963s).

Implementation choices:
- Why one iteration? - Our tokens have high entropy (168 bits), so an attacker who is in possession of a database copy can only use brute force. The size of the combinatorial space makes that impractical. As a consequence, there's no point in adding a time penalty (unlike with user passwords!). However, we don't want to spend unnecessary time in verifying tokens, so keeping the number of iterations low is a good thing. If we think our tokens should be slower to break, we should increase entropy.
- Why static salt? - The purpose of salt is to make rainbow table generation infeasible, by enlarging the space using some extra entropy (separate from the secret, usually a password). However, as our tokens have high entropy, rainbow tables are not feasible in the first place. Another way to look at it: One could consider the salt's entropy to be contained in the token itself. On the other hand, if the salt was not static, then it would not be clear which row to look up from the database.
- Why PBKDF2-HMAC-SHA256? - It's what NIST recommends. The advantage over using plain SHA256 is that length attacks "by adding to the digest" are not possible. (That's more of an attack scenario for integrity checks than for secret storage, but whatever.) I chose Django's password-management tooling ("hasher configuration") so that we can more easily change the algorithm and its parameters without invalidating existing tokens, in case we should want to do that in the future.

I also noticed that the `tokens/` list endpoint is lacking comprehensive tests. I'll add those in another commit. I'll also think about the token management endpoints once more (with future functionality like domain-scoping in mind), and in case I come up with something useful, I'll put it in.


api/desecapi/views.py
```diff
@@ -581,7 +581,7 @@ def finalize(self):
             PDNSChangeTracker.track(lambda: DomainList.auto_delegate(domain))
             token = models.Token.objects.create(user=action.user, name='dyndns')
             return Response({
-                'detail': 'Success! Here is the password ("auth_token") to configure your router (or any other dynDNS '
+                'detail': 'Success! Here is the password ("token") to configure your router (or any other dynDNS '
                           'client). This password is different from your account password for security reasons.',
```
nils-wisiol: I would like to make the distinction between token and account password more clear. One attempt:

Success! Here is the authentication token to configure your router, or any other DNS client. It must be used as the password for your client. Never store your account password anywhere in plain text, including in configuration files or router devices.

api/desecapi/tests/base.py
```diff
@@ -689,7 +689,7 @@ def has_local_suffix(cls, domain_name: str):
     def create_token(cls, user):
         token = Token.objects.create(user=user)
         token.save()
-        return token
+        return token.key
```
nils-wisiol: that will make it harder to delete tokens

api/desecapi/models.py
```diff
@@ -169,18 +178,26 @@ class Token(rest_framework.authtoken.models.Token):
     name = models.CharField("Name", max_length=64, default="")
     user_specific_id = models.BigIntegerField("User-Specific ID")
 
+    class Meta:
+        abstract = False
```
nils-wisiol: no-op?

api/desecapi/models.py
```diff
@@ -169,18 +178,26 @@ class Token(rest_framework.authtoken.models.Token):
...
+        unique_together = (('user', 'user_specific_id'),)
+
+    def generate_key(self):
+        plain = b64encode(urandom(21)).decode('utf-8').replace('/', '-').replace('=', '_').replace('+', '.')
```
nils-wisiol: use urlsave b64enc? Maybe this branch should be rebased on top of 20191108_encrypt_confirmation_codes or cherry-pick 6460c4e867844702dfdd111237b70ad37c695cb4

api/desecapi/models.py
```diff
@@ -159,8 +160,16 @@ def send_email(self, reason, context=None, recipient=None):
         email.send()
 
 
+class TokenHashWithPlaintext(str):
```
nils-wisiol: why attach the plain value on the hashed string rather then just attach it to the Token class?

api/desecapi/models.py
```diff
@@ -169,18 +178,26 @@ class Token(rest_framework.authtoken.models.Token):
...
+        unique_together = (('user', 'user_specific_id'),)
+
+    def generate_key(self):
+        plain = b64encode(urandom(21)).decode('utf-8').replace('/', '-').replace('=', '_').replace('+', '.')
+        return TokenHashWithPlaintext(plain)
+
+    @property
+    def plain(self):
+        try:
+            return self.key.plain
```
nils-wisiol: seems unnecessarily complicated: just `return self.plain`?

api/desecapi/tests/base.py
```diff
@@ -545,6 +546,12 @@ def assertResponse(self, response, code=None, body=None):
         if body:
             self.assertJSONEqual(response.content, body)
 
+    def assertToken(self, plain, user=None):
+        user = user or self.owner
+        hashed = make_password(plain, salt='static', hasher='pbkdf2_sha256_iter1')
+        self.assertEqual(len(hashed.split('$')), 4)
+        self.assertIn(hashed, Token.objects.filter(user=user).values_list('key', flat=True))
```
nils-wisiol: can we use Django's check mechanism here?

docs/authentication.rst
```diff
@@ -442,18 +442,17 @@ any action. We may implement specialized tokens in the future.
...
-have any padding at the end because the string length is a multiple of 4.)
+Any token is generated from 168 bits of randomness at the server and stored in
+hashed format (PBKDF2-HMAC-SHA256). Guessing the token correctly or reversing
+the hash is hence practically impossible.
 
-As all tokens are stored in plain text on the server, the user may not choose
-the token value individually to prevent re-using passwords as tokens at deSEC.
+The token value is represented by 28 characters using a URL-safe variant of
+base64 encoding. It comprises only the characters ``A-Z``, ``a-z``, ``-``, and
+``.``. (Base64 padding is not needed as the string length is a multiple of 4.)
```
nils-wisiol: again, this should probably cherry-pick 6460c4e867844702dfdd111237b70ad37c695cb4

api/desecapi/tests/base.py
peterthomassen: That's right. The intention was to later replace it with `token.plain` without having to change all the other tests which use the tokens. Otherwise, they would need to know the difference between the plain and the encrypted version.

Within out tests, we currently don't do deletion on the Token object itself (if we want to test deletion, we'd use the API). If we need to, we can do `Tokens.objects.get(key=TokenHashWithPlaintext(token)).delete()`*. (Looks like we should override the getter and allow it to accept `plain=...`.)

*: of course, `TokenHashWithPlaintext` implementation is also up for discussion

api/desecapi/models.py
peterthomassen: that was just reordering because we usually have the `Meta` class before the method declarations. I can undo the change, or put it into a separate commit. Opinions?

api/desecapi/models.py
peterthomassen: My plan was to rebase it on top of `20191108_encrypt_confirmation_codes` once the discussions there are settled

api/desecapi/models.py
peterthomassen: Having `plain` as a direct attribute of `Token` seems indeed more natural, but I somehow had gotten stuck on that route. I don't recall the exact reason, but I'll try to reproduce it and then we can see how to proceed.

api/desecapi/models.py
peterthomassen: will go away if the `TokenHashWithPlaintext` discussion is settled

docs/authentication.rst
peterthomassen: planning to rebase on top of other branch

api/desecapi/views.py
peterthomassen: changed to:

Success! Here is the secret token required for updating your domain's DNS information. When configuring a router (or other DNS client), place it into the password field of the configuration. Do not confuse the secret token with your account password! Your password is not needed for DNS configuration, and you should not store it anywhere in plain text.

api/desecapi/tests/base.py
peterthomassen: deletion code would now be `Tokens.objects.get(key=Token.make_hash(token)).delete()`. We can write a manager that allows doing `Tokens.objects.get(plain=token).delete()`.

api/desecapi/models.py
peterthomassen: moved assignment of `self.plain` and `self.key` directly to `generate_key`

api/desecapi/models.py
peterthomassen: gone

peterthomassen:
> Nice work. One question that came to my mind is if we would be able to increase the number of iterations (currently one) without user interaction, if ever needed?

Yes, we can do that. That's why I chose `make_password`, because it encodes the number of iterations etc. in the output string. If we increase the number of iterations in the hasher (https://github.com/desec-io/desec-stack/blob/20191111_token_refactoring/api/desecapi/authentication.py#L123), we can use [`check_password()`](https://docs.djangoproject.com/en/2.2/_modules/django/contrib/auth/hashers/#check_password) to force an update in the database.

api/desecapi/tests/base.py
peterthomassen: good catch! See fix in https://github.com/desec-io/desec-stack/commit/9774af98ee8dab78e5df2a3a35169fdb18ea276c

Downsides:
- It's now not so easy anymore to check that hashed tokens in fact are in hashed format, with `$` in it etc, as we're not dealing with a hashed token directly anymore
- The hash is computed in `check_password()` once per candidate, so the same thing is computed repeatedly (but it's very fast, so no problem)

peterthomassen:
> > Nice work. One question that came to my mind is if we would be able to increase the number of iterations (currently one) without user interaction, if ever needed?
> 
> Yes, we can do that. That's why I chose `make_password`, because it encodes the number of iterations etc. in the output string. If we increase the number of iterations in the hasher (https://github.com/desec-io/desec-stack/blob/20191111_token_refactoring/api/desecapi/authentication.py#L123), we can use [`check_password()`](https://docs.djangoproject.com/en/2.2/_modules/django/contrib/auth/hashers/#check_password) to force an update in the database.

I take it back, that was wrong. PBKDF2 does not "chain along iterations".

peterthomassen:
Worked out the agreed-upon KDF mechanism. Does this look good to you? If so, let me add some tests and then we should be ready to go

peterthomassen:
rebased on top of 20191108_encrypt_confirmation_codes

peterthomassen:
switched Token.id to UUID and fixed up the deletion endpoint at `auth/tokens/:uuid/`.

Now need to add tests for the TokenViewSet, and then this is ready to roll!

peterthomassen:
I finally got around to add token tests. I think we're now all covered! Ready to ship? @nils-wisiol

api/desecapi/models.py
nils-wisiol: I'd put it in a separate commit, but I won't be dogmatic about it

peterthomassen:
> Otherwise, this looks good to me.

thx

> A side note on the tests: I prefer the tests to be added directly with the feature, to easy debugging and bug hunting when going through git history.

I agree. However, we were completely lacking tests for the endpoints under `tokens/` until now, but the functionality has been there for a long time. So this time it had to go to a separate commit ...

peterthomassen:
> [6e6dafa](https://github.com/desec-io/desec-stack/commit/6e6dafa71c6bd1890140df9292401f21aab98b98) breaks the email we currently sent upon dyn domain create (via `signals.py`). This should be fixed, however also note that in #269, we are planning to remove this email message entirely.

Why does it break that? That commit only hides token values from the views when the token is not fresh. The model still has the plaintext value (no hashing applied in this commit yet), and the email can be sent as usual.
The other commit which introduces hashing also changes the signal handler code to use the `.plain` attribute of the fresh token, so that should also work.

If something breaks here, then I have made a logical error. Can you explain?

peterthomassen:
clarified in discussion; issue was that GitHub sorts by commits timestamp and not by tree order

PR #266, opened 2019-11-18
fix(api): forward only selected pdns error statuses, closes #265

### LINKED ISSUE #265: Don't forward PDNSException status codes to API user
https://github.com/desec-io/desec-stack/blob/master/api/desecapi/exceptions.py#L10

e.g. if Domain.keys() raises PDNSException, the user will get HTTP 404. Should be 500.

peterthomassen:


peterthomassen:
I reworked the whole thing. If pdns returns an error that is not 422, we will now get an email with extra information from the pdns response, but the client is told `503 Service Unavailable`. This uses the same mechanism as our database exception handling.

Looks better now?

api/desecapi/exceptions.py
```diff
@@ -1,17 +1,28 @@
...
 
-class PDNSException(APIException):
+class RequestEntityTooLarge(APIException):
+    status_code = status.HTTP_413_REQUEST_ENTITY_TOO_LARGE
+    default_detail = 'Payload too large.'
+    default_code = 'too_large'
+
+
+class PDNSValidationError(ValidationError):
+    status_code = 422
```
nils-wisiol: should we change this code to 400? Our own validation will also return 400, so why differ here by where the error came from?

api/desecapi/exceptions.py
```diff
@@ -1,17 +1,28 @@
...
+
+    def __init__(self, response=None):
+        try:
+            detail = json.loads(response.text)['error']
+        except JSONDecodeError:
```
nils-wisiol: add KeyError

peterthomassen: I did not add KeyError because we expect it to always be present. If it is not, we should have this turn into a 500 error so that we can investigate what's going on. Ok for you?

nils-wisiol: alright, fair point

peterthomassen:
changed status code 422 to 400 in a separate commit. can you take a look before I squash? thanks!

nils-wisiol:
lgtm. Maybe keep it separately, as it qualifies as a "feat", not a "fix".

peterthomassen:
done

PR #267, opened 2019-11-19
feat(api): add fast / slow lane depending on email type, closes #242

### LINKED ISSUE #242: api: Differentiate between urgent and rate-limited emails

peterthomassen:
Todos:

- add some more logging at sending time to see which track is being sent from (Celery)
- think about making it nicer, somehow (`mail_backends.py` has two dicts, a class, a function, and a loop ...)
- add some tests. @nils-wisiol , what exactly do you think we should test?

### DIFF
--- a/api/api/celery.py
+++ b/api/api/celery.py
@@ -4,7 +4,7 @@ from celery import Celery
 
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'api.settings')
 
-app = Celery('api')
+app = Celery('api', include='desecapi.mail_backends')
 app.config_from_object('django.conf:settings', namespace='CELERY')
 app.autodiscover_tasks()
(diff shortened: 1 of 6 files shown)


peterthomassen:
I made the code a bit nicer, moving config to settings and encapsulating some more stuff in the `CustomCeleryEmailBackend` class. I also added some logging.

I think this is ready for review.

What's still missing is the tests. I am unsure how to test the email backend, as the tests use a dummy backend. As a result, the test coverage for `mail_backends.py` is 0%.

peterthomassen:
Renamed to `MultiLaneEmailBackend`, and added tests in b8b43513ff114c3b8a1f0a6f7ba45ecfb6fbbf3f. Can you take quick look before I squash?

PR #278, opened 2019-12-09
rate limiting

peterthomassen:


nils-wisiol:
Regarding the removal of the IP based rate limit for static assets, I'm not sure if the test that you conducted is valid. As you're using only one IP address, you do not consider the scenario where a single remote IP blocks so many of nginx's resources that other requests won't come through. From that point of view, I am not yet convinced that a rate limit is completely pointless. On the other hand, I am aware that we also do not have proof that a rate limit /is/ useful.

peterthomassen:
Here are my thoughts on this (but I'm also uncertain):

If the client just does many requests (acknowledging the response), most of the time will be spent on establishing the TLS connection (which is before rate limiting anyways), and only little time will be spent in answering the request. If instead a 429 response is sent, little time is saved.

Another case is when a malicious client tries to occupy resources by receiving the response artificially slowly, so that the connection remains open for a long time while the response is being sent. In this case, rate limiting may help, because the request has already been fully received.

However, the connection will remain open also in this case if the client does not TCP ACK the arrival of the 429 response, or when the client sends the request artificially slowly, e.g. by sending one header byte every 30 seconds. This kind of attack can't be handled by nginx rate limiting in general.

So, there may be some edge cases in which rate limiting helps, and if the attacker chooses those, then ok. The questions remains though what a reasonable number of requests per second per IP would be, then? -- Again, the question is not about what a legit use case would be, but starting from which number of requests we would actually see a benefit from sending 429 instead of 200 along with our (rather small) static payload.

I don't mind leaving the rate limit in if we have some rough estimate of what would be a reasonable number. I just figured that given that the advantage is small for static files, and that there are ways around it, maybe it's better to stop thinking about it ...?

peterthomassen:
There is now a commit that introduces API-level rate limiting. I decided to distinguish five different cases, see limit definitions in `settings.py`.

There are a few open questions:
- The `basic` limit is a workaround for AuthenticatedAction views. Those should actually be covered by `account_management_active`, but when doing so, the throttling mechanism triggers the authenticator to run (because the throttling is user-specific). However, so far, we only ran authentication once the authenticated action is actually executed, for a good reason: Authentication uses the serializer (so that the action code can be verified); in doing so, the serializer will complain about missing fields (such as `new_password`). As a consequence, rate-limited calls to `AuthenticatedResetPasswordUserActionView` will return `400 Bad Request` if `new_password` is missing, with an error message indicating that. That's not the intended effect of the rate limit.
   A solution would be to always use the `AuthenticatedUserActionSerializer` within the authenticator. Sides may require a minor refactoring in another place, so I wanted to ask for an opinion before I do it.
- I think the rates I configured are reasonable, but we should discuss them.
- Currently, rate limiting is turned off in tests (by setting all rates to `None`). We probably should test the `ScopedRatesThrottle` somehow. Should we also test that each view has a throttle configured?
- Rate limiting requires noting down timestamps of each user's request so that a decision can be made on the next request. This uses Django's cache framework, which by default uses the in-memory backend. This is process-specific. As a consequence, the rate limit is really unreliable (we have 64 processes or so). The solution suggested in the Django docs is using memcache. So one remaining TODO is adding a memcache container and adding that to the Django configuration. Do you have any objections to that? (I somehow dislike the overhead, but this seems unavoidable in a multi-process setup.)

peterthomassen:
Forgot to mention: I compared running the test suite multiple times on master as well as on this branch with all rate limits in effect (but with high allowed rates). It turns out that the performance penalty is negligible (definitely sub-percent).

peterthomassen:
The approach used in dbc02470bdc4ece05a6dc4564a5a4db9136323a0 does not seem the most elegant one, but it was simple because the `envsubst` script is already there. (Note that there is no direct way of accessing environment variables within the nginx configuration.)

Another approach would be to outsource the `limit_req` directive into a separate configuration file which gets included in the configuration, and make sure that the entrypoint scripts deletes or truncates that file if it detects that it's running within a test. Any preferences?

peterthomassen:
Added memcached. Still need to do some testing about how the various types of throttles interact.

peterthomassen:
We use nginx' `limit_req` for throttling requests. One can also rate limit connections per IP using `limit_conn`.

Should we? The implications seem non-trivial for me, e.g., the nginx docs say "In HTTP/2 and SPDY, each concurrent request is considered a separate connection." -- now thing of a browser fetching our static assets; in HTTP/1, this would be one or two connections only, but now they get counted as 20. So with HTTP/2, for this purpose, connection is roughly equivalent to request. Why are there two different limit directives?

peterthomassen:
From my perspective, this is ready now. Please review!

Two open thoughts:
- The documentation does not mention rate limiting. I don't think we need to reproduce all the the different rates for the different views and HTTP methods in the docs. I think for the moment it's even fine to not mention it at all. (We also did not use to mention the nginx rate limit that we've always had.)
- Tests currently test the throttling class, but not the fact that all routes' views have a throttling scope configured. Does this need testing? How? (I did not find a way to iterate over views.)

api/api/settings.py
```diff
@@ -97,6 +97,21 @@
     'EXCEPTION_HANDLER': 'desecapi.exception_handlers.exception_handler',
     'DEFAULT_VERSIONING_CLASS': 'rest_framework.versioning.NamespaceVersioning',
     'ALLOWED_VERSIONS': ['v1', 'v2'],
+    'DEFAULT_THROTTLE_CLASSES': [
+        'desecapi.throttling.ScopedRatesThrottle',
+        'rest_framework.throttling.UserRateThrottle',
+    ],
+    'DEFAULT_THROTTLE_RATES': {
+        # ScopedRatesThrottle
+        'account_management_active': ['3/min'],  # things that send email or create domains on pdns
```
nils-wisiol: creating domains on pdns should go into the `dns_api_write` group, no?

api/api/settings.py
```diff
@@ -97,6 +97,21 @@
...
+        'account_management_passive': ['10/min'],  # things like viewing your account or creating/deleting tokens
+        'dyndns': ['1/min'],  # dynDNS updates; anything above 1/min is a client misconfiguration
+        'dns_api_read': ['5/s', '50/min'],  # DNS API requests that do not involve pdns
+        'dns_api_write': ['3/s', '50/min', '200/h'],  # DNS API requests that do involve pdns
+        # UserRateThrottle
+        'user': '1000/d',  # hard limit on requests by a) an authenticated user, b) an unauthenticated IP address
+    },
+    'NUM_PROXIES': 0,  # Do not use X-Forwarded-For header when determining IP for throttling
```
nils-wisiol: then all requests will be considered coming from nginx's IP address?

api/desecapi/throttling.py
```diff
@@ -0,0 +1,58 @@
+from rest_framework import throttling
+from rest_framework.settings import api_settings
+
+
+class ScopedRatesThrottle(throttling.ScopedRateThrottle):
+    """
+    Like DRF's ScopedRateThrottle, but supports several rates per scope, e.g. for burst vs. sustained limit.
+    """
+    def parse_rate(self, rates):
+        return [super(ScopedRatesThrottle, self).parse_rate(rate) for rate in rates]
```
nils-wisiol: I think just `super().parse_rate` does the job here

www/conf/nginx.conf
```diff
@@ -34,9 +34,6 @@ http {
     # if rate limits are exceeded, clients will receive a 503 answer (try again later)
     # rate limits are exceeded when the 'leaky bucket' is full
 
-    # set up one bucket per remote ip for general purpose
-    limit_req_zone $binary_remote_addr zone=perip-general:100m rate=30r/s;
```
nils-wisiol: it appears that this bucket was not used anyways, at least `perip-general` does not appear anywhere in current nginx configuration

www/conf/nginx.conf
```diff
@@ -36,6 +36,7 @@ http {
 
     # set up one bucket per remote ip for (costly) API access
```
nils-wisiol: adjust comment

www/conf/sites-available/90-desec.api.location.var
```diff
@@ -2,9 +2,7 @@
 # Pass API requests to an API server
 ######
 location /api/ {
-    # max .5r/s to the API per IP, but 10 at once is okay. This limit may need to be increased once
-    # client applications become more powerful
-    ${PROD_ONLY} limit_req zone=perip-api burst=10 nodelay;
+    ${PROD_ONLY} limit_req zone=perip-api;
```
nils-wisiol: I believe `PROD_ONLY` variable is only introduced in the next commit

api/api/settings.py
peterthomassen: Well, yes. What's meant here is creating a dyn domain when clicking the account activation link. This can happen with the same rate as the activation email can be sent. Does it make sense? Maybe I need to improve the comment?

api/api/settings.py
peterthomassen: Doesn't the API see the client IP directly? We use it [here](https://github.com/desec-io/desec-stack/blob/4136abc/api/desecapi/views.py#L294).

I wonder why that is. Does [`uwsgi_pass`](https://github.com/desec-io/desec-stack/blob/c07c4ea/www/conf/sites-available/90-desec.api.location#L13) do the magic?

api/desecapi/throttling.py
peterthomassen: [nope](https://stackoverflow.com/a/31895448/6867099), as `self` is undefined within list comprehensions (it also did not make sense to me intuitively)

www/conf/sites-available/90-desec.api.location.var
peterthomassen: no, it is introduced in the same commit (in `envreplace.sh`)

www/conf/nginx.conf
peterthomassen: done

peterthomassen:
> I don't fully understand the changes of 0861dba, let's discuss that in offline.

I'm happy to explain it here. Authentication in authenticated action views happens via `AuthenticatedActionAuthentication`. Before 0861dba, [this used the view's serializer](https://github.com/desec-io/desec-stack/commit/0861dbaa96af966b5db42076d9898e44ffb9cac4#diff-9113ed2b3def50a931c704932d7edb84L132) (which always inherited from `AuthenticatedUserActionSerializer`). As a result, authentication would fail when validation of one of the other (not user-related) serializer fields failed. That's technically wrong, because authentication should only check the user. This subtle "imperfection" turned into a problem now, as follows.

Rate limiting is per user (if authenticated), otherwise per IP. This means that rate limiting has to run authentication to extract the user. For the rate limit on authentication action endpoints, this means decoding the action code. This would now 1) trigger validation of fields inside the code that are not needed to make a throttling decision (i.e. waste of time), 2) yield premature validation results that lead to unintended results. I'll explain what I mean by the second case.

Let's say you create a verification link for changing your email. To perform the action, you click on the link (GET request), the API sends a redirect to our frontend, and the frontend then sends a POST request to the API endpoint (same URL as in email, but different method). What happens if somebody created a new (not yet activated) account with that email address before you click your change-email verification link? You would expect that the same workflow repeats, and that the frontend's POST request to the API yields [a uniqueness error](https://github.com/desec-io/desec-stack/blob/0861dba/api/desecapi/serializers.py#L674). Before 0861dba, this was actually what happened, because we delayed authentication until `request.auth` was used (see the `perform_authentication()` method on the view).

But now, with rate limiting, the throttling mechanism does authentication in order to make its throttling decision, which in turn runs the serializer validation! That is, before the view's main work gets done, the new email address is validated, even on the GET request that is supposed to always return a redirect (or maybe a rate limit error), but never supposed to return a payload validation error!

In other words, we must not validate anything besides the user herself when doing authentication. That's why I changed the `authenticate()` method such that it always uses `AuthenticatedUserActionSerializer` explicitly, regardless of the circumstances. This 1) saves time* when rejecting throttled requests and 2) avoids the above unintended side results. As a result, I had to introduce the `action` attribute on the view.

*: The validation is repeated with the specialized serializer when the `post()` method is run. In fact, that causes extra time to be spent. Getting rid of this does not seem straightforward to me. However, this happens only once the request was accepted by the throttling mechanism. When rejecting requests, we save time.

> Also, what does memcache cache?

It caches the last _n_ request timestamps per authenticated user or otherwise IP. The Django in-memory cache backend is inappropriate as it is per process, but we need something that covers the whole application.

api/api/settings.py
nils-wisiol: yup let's make clear here "create" here only refers to creation of domains as part of user registration process

api/api/settings.py
nils-wisiol: Indeed, it's a combination of our [uwsgi configuration](https://uwsgi-docs.readthedocs.io/en/latest/Nginx.html#what-is-the-uwsgi-params-file) and nginx's [remote_addr variable](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_remote_addr)

api/api/settings.py
peterthomassen: fixed comment

nils-wisiol:
This looks good. Let's squash and merge

PR #281, opened 2019-12-16
Ho! Ho! Ho! Santa's here 🎅

nils-wisiol:
Adds a [Santa hat](https://desec.sandbox.dedyn.io/) to the logo.

Jokes aside, this is also to test if our image deployment works correctly. On my development machine, it works /sometimes/. 🤐

Changes in this PR can easily be reverted.


peterthomassen:
haha cute! But shouldn't the hat be hanging on the top tip of the d, instead of the bulge? (I looked at https://desec.sandbox.dedyn.io/img/logo.3bdcf119.svg assuming it's the same)

peterthomassen:
ah but I understand now that that would require changing the image dimensions at the top. Ok, fine! lgtm

nils-wisiol:
It should, but ... yeah exactly

PR #282, opened 2019-12-19
add openvpn-server

peterthomassen:
Last thing to fix that I am aware of is a potential IP subnet collision on stack and slave. That's not currently an issue, but it could become one if the slave has 7 containers (currently 4).

### DIFF
--- a/openvpn-server/conf/server.conf
+++ b/openvpn-server/conf/server.conf
@@ -1,0 +1,168 @@
+port 1194
+
+proto udp
+
+dev tun
+
+ca /etc/openvpn/secrets/ca.crt
+cert /etc/openvpn/secrets/server.crt
+key /etc/openvpn/secrets/server.key  # This file should be kept secret
+
+tls-version-min 1.3 or-highest
+tls-cipher TLS-ECDHE-RSA-WITH-AES-256-GCM-SHA384
+ncp-ciphers AES-256-GCM
+auth SHA256
+dh none
+ecdh-curve secp384r1
+
+remote-cert-tls client
+
+# Network topology
+# Should be subnet (addressing via IP)
+# unless Windows clients v2.0.9 and lower have to
+# be supported (then net30, i.e. a /30 per client)
+# Defaults to net30 (not recommended)
+topology subnet
+
+# Configure server mode and supply a VPN subnet
+# for OpenVPN to draw client addresses from.
+# The server will take 10.8.0.1 for itself,
+# the rest will be made available to clients.
+# Each client will be able to reach the server
+# on 10.8.0.1. Comment this line out if you are
+# ethernet bridging. See the man page for more info.
+server 10.8.0.0 255.255.255.0
+
+# Maintain a record of client <-> virtual IP address
+# associations in this file.  If OpenVPN goes down or
+# is restarted, reconnecting clients can be assigned
+# the same virtual IP address from the pool that was
+# previously assigned.
+ifconfig-pool-persist /var/log/openvpn/ipp.txt
+
+# Configure server mode for ethernet bridging.
+# You must first use your OS's bridging capability
+# to bridge the TAP interface with the ethernet
+# NIC interface.  Then you must manually set the
+# IP/netmask on the bridge interface, here we
+# assume 10.8.0.4/255.255.255.0.  Finally we
+# must set aside an IP range in this subnet
+# (start=10.8.0.50 end=10.8.0.100) to allocate
+# to connecting clients.  Leave this line commented
+# out unless you are ethernet bridging.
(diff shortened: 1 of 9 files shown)


openvpn-server/conf/server.conf
```diff
@@ -0,0 +1,170 @@
...
+ecdh-curve secp384r1
+
+remote-cert-tls client
+
+# Network topology
+# Should be subnet (addressing via IP)
+# unless Windows clients v2.0.9 and lower have to
+# be supported (then net30, i.e. a /30 per client)
+# Defaults to net30 (not recommended)
+;topology subnet
```
nils-wisiol: comment above says this should be `subnet`

openvpn-server/conf/server.conf
```diff
@@ -0,0 +1,170 @@
...
+# Certain Windows-specific network settings
+# can be pushed to clients, such as DNS
+# or WINS server addresses.  CAVEAT:
+# http://openvpn.net/faq.html#dhcpcaveats
+# The addresses below refer to the public
+# DNS servers provided by opendns.com.
+;push "dhcp-option DNS 208.67.222.222"
+;push "dhcp-option DNS 208.67.220.220"
+
+client-to-client
```
nils-wisiol: do we need client-to-client configuration?

openvpn-server/conf/server.conf
peterthomassen: good catch, fixed

openvpn-server/conf/server.conf
peterthomassen: no, fixed. thx!

peterthomassen:
Regarding my initial comment:

> Last thing to fix that I am aware of is a potential IP subnet collision on stack and slave. That's not currently an issue, but it could become one if the slave has 7 containers (currently 4).

The thing is that the slave has access to the master-side replication network, which is of the form `x.y.7.z`. For example, nsmaster is `x.y.7.3`. `x.y` is defined in the stack deployment configuration (via the `.env` file).

In the slave configuration, internal networks have a hardcoded prefix of `172.16`. Currently, we use `172.16.[0-3].z`. If we add a slave-side /24 network prefixed with `172.16.7`, and if the stack deployment configuration has the prefix `x.y` configured to be `172.16`, then we have a collision.

it is unlikely that the slave will ever have that many networks, so it is probably a non-issue. In case it will be an issue, we can work around it by not skipping `172.16.7` on the slave (using `.8` instead), or by making the slave prefix configurable. Migration would be a non-brainer as slaves are designed to be of the throw-away type.

I therefore suggest to no deal with this now.

If you agree, then I think this is ready for merge.

peterthomassen:
for the records, security-related OpenVPN configuration is inspired by https://www.privacy-handbuch.de/handbuch_97a.htm and https://www.securai.de/veroeffentlichungen/blog/sichere-konfiguration-eines-vpn-mit-openvpn/

peterthomassen:
Howto for certificate provisioning: https://community.openvpn.net/openvpn/wiki/EasyRSA3-OpenVPN-Howto

nils-wisiol:
Looks good. Let's exchange the curve to a better one as soon as OpenVPN supports it. (Or change to wireguard altogether as soon as they are production-ready.)

Also, add some documentation how to get the secrets ready for usage.

peterthomassen:
added secrets provisioning information to README, and also updated other aspects of README in a separate commit.

peterthomassen:
incorporating in #284

andreasnuesslein:
:+1:  on wireguard being used instead of openvpn - altough it's really not complicated per se, I can offer assistance since I've set it up a bunch of times already

nils-wisiol:
This would be highly appreciated, as OpenVPN has a couple of limitations and Ubuntu 20.04 ships wireguard!

andreasnuesslein:
yes indeed :)

on that note: how _do_ you deploy on your frontend machines?  just `apt install docker && `docker-compose up` or do you run the different services natively on the frontend machines?

peterthomassen:
We run this docker-compose application on pretty much vanilla VMs: https://github.com/desec-io/desec-slave

There is some other stuff running (BGP daemon for anycast etc.), but that's not part of the frontend stack.

nils-wisiol:
There is [some work in progress on automation of deployment,](https://github.com/desec-io/automation/pull/1) with OpenVPN being a major complication.

andreasnuesslein:
Since this is kind of an omnibus installation I highly recommend to do something similar to what gitlab is doing for example and use some actual configuration management for deployment. I personally would probably put this into saltstack. And depending on whether you want to have one central master-node from where you can configure all your servers or if they should be self-sufficient (more or less) you can set it up masterless or not.

Issue #220, opened 2019-07-02
API: Cannot Replace AAAA with CNAME in Single Request


domenukk:
Removing AAAA and adding a CNAME in the same request fails:
```python
PUT @ route /domains/enowars.com/rrsets/ [{'type': 'CNAME', 'records': ['elk.enowars.com.'], 'subname': 'kibana', 'ttl': 600}, {'type': 'AAAA', 'records': [], 'subname': 'kibana', 'ttl': 1}]
Request failed!
Unexpected error!
{"detail":"RRset kibana.enowars.com. IN CNAME: Conflicts with pre-existing non-CNAME RRset"}
```

The current entries were:
```bash
hnj@prokrastinator:~/Projects/ENOWARS2019/enonetwork/code/ansible$ dig ANY kibana.enowars.com

; <<>> DiG 9.10.3-P4-Ubuntu <<>> ANY kibana.enowars.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 43922
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;kibana.enowars.com.		IN	ANY

;; ANSWER SECTION:
kibana.enowars.com.	589	IN	RRSIG	AAAA 13 3 600 20190711000000 20190620000000 5301 enowars.com. Tf2LVZjOnrYj9myLmGgzLoz1+oM5HK2+33PM3lKUbEzv7SgS/4CfogIu 7AU1SaH1s8clbSlMT+ocV2vqLHBBRw==
kibana.enowars.com.	589	IN	AAAA	fd00:1337:0:1:e61f:13ff:fecb:3c5a

;; Query time: 4 msec
;; SERVER: 127.0.1.1#53(127.0.1.1)
;; WHEN: Tue Jul 02 17:50:31 CEST 2019
;; MSG SIZE  rcvd: 182
```

nils-wisiol:
My guess is that the change tracker should either list the deletions first, or make to requests to pdns to avoid the conflict. Currently, [deletions are last in the list](https://github.com/desec-io/desec-stack/blob/f0d288b205e82eb5555e96f913424b8223ead827/api/desecapi/pdns_change_tracker.py#L127-L158).

peterthomassen:
[...]

PR #284, opened 2019-12-21
add catalog zone for new replication method

peterthomassen:
https://tools.ietf.org/html/draft-muks-dnsop-dns-catalog-zones-04

### DIFF
--- a/api/desecapi/management/commands/check-slaves.py
+++ b/api/desecapi/management/commands/check-slaves.py
@@ -1,0 +1,95 @@
+from datetime import timedelta
+from socket import gethostbyname
+from time import sleep
+
+from django.conf import settings
+from django.core.mail import get_connection, mail_admins
+from django.core.management import BaseCommand
+from django.utils import timezone
+import dns.message, dns.query, dns.rdatatype
+
+from desecapi import pdns
+from desecapi.models import Domain
+
+
+def query_serial(zone, server):
+    query = dns.message.make_query(zone, 'SOA')
+    response = dns.query.tcp(query, server)
+
+    for rrset in response.answer:
+        if rrset.rdtype == dns.rdatatype.SOA:
+            return int(rrset[0].serial)
+    return None
+
+
+class Command(BaseCommand):
+    help = 'Check slaves for consistency with nsmaster.'
+
+    def __init__(self, *args, **kwargs):
+        self.servers = {gethostbyname(server): server for server in settings.WATCHDOG_SLAVES}
+        super().__init__(*args, **kwargs)
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain-name', nargs='*',
+                            help='Domain name to check. If omitted, will check all recently published domains.')
+        parser.add_argument('--delay', type=int, default=60, help='Delay SOA checks to allow pending AXFRs to finish.')
+        parser.add_argument('--window', type=int, default=settings.WATCHDOG_WINDOW_SEC,
+                            help='Check domains that were published no longer than this many seconds ago.')
+
+    def find_outdated_zones(self, zone, local_serial):
+        """
+        Returns a dict, the key being the outdated slave name, and the value being the slave's current zone serial.
+        """
+        outdated = {}
+        for server in self.servers:
+            remote_serial = query_serial(zone, server)
+            if not remote_serial or remote_serial < local_serial:
+                outdated[self.servers[server]] = remote_serial
+
+        return outdated
+
+    def handle(self, *args, **options):
+        threshold = timezone.now() - timedelta(seconds=options['window'])
+        recent_domain_names = Domain.objects.filter(published__gt=threshold).values_list('name', flat=True)
+        serials = {zone: s for zone, s in pdns.get_serials().items() if zone.rstrip('.') in recent_domain_names}
+
+        if options['domain-name']:
+            serials = {zone: serial for zone, serial in serials.items() if zone.rstrip('.') in options['domain-name']}
+
+        print('Sleeping for {} seconds before checking {} domains ...'.format(options['delay'], len(serials)))
+        sleep(options['delay'])
+
+        outdated_zone_count = 0
+        outdated_slaves = set()
+
+        output = []
+        for zone, local_serial in serials.items():
+            outdated_serials = self.find_outdated_servers(zone, local_serial)
+            outdated_slaves.update(outdated_serials.keys())
+
+            if outdated_serials:
+                output.append(f'{zone} ({local_serial}) is outdated on {outdated_serials}')
+                print(output[-1])
+                outdated_zone_count += 1
+            else:
+                print(f'{zone} ok')
+
+        output.append(f'Checked {len(serials)} domains, {outdated_zone_count} were outdated.')
+        print(output[-1])
+
+        self.report(outdated_slaves, output)
+
+    def report(self, outdated_slaves, output):
+        if not outdated_slaves:
+            return
+
+        subject = f'ALERT {len(outdated_slaves)} slaves out of sync'
+        message = f'The following {len(outdated_slaves)} slaves are out of sync:\n'
+        for outdated_slave in outdated_slaves:
+            message += f'* {outdated_slave}\n'
+        message += '\n'
+        message += f'Current slave IPs: {self.servers}'
+        message += '\n'
+        message += '\n'.join(output)
+
+        mail_admins(subject, message, connection=get_connection('django.core.mail.backends.smtp.EmailBackend'))
--- a/api/desecapi/management/commands/sync-to-pdns.py
+++ b/api/desecapi/management/commands/sync-to-pdns.py
@@ -1,0 +1,72 @@
+from django.core.management import BaseCommand, CommandError, call_command
+from django.db import transaction
+
+from desecapi import pdns
+from desecapi.exceptions import PDNSException
+from desecapi.models import Domain
+from desecapi.pdns_change_tracker import PDNSChangeTracker
+
+
+class Command(BaseCommand):
+    help = 'Sync RRsets from local API database to pdns.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain-name', nargs='*',
+                            help='Domain name to sync. If omitted, will import all API domains.')
+
+    def handle(self, *args, **options):
+        domains = Domain.objects.all()
+
+        if options['domain-name']:
+            domains = domains.filter(name__in=options['domain-name'])
+            domain_names = domains.values_list('name', flat=True)
+
+            for domain_name in options['domain-name']:
+                if domain_name not in domain_names:
+                    raise CommandError('{} is not a known domain'.format(domain_name))
+
+        catalog_alignment = False
+        for domain in domains:
+            self.stdout.write('%s ...' % domain.name, ending='')
+            try:
+                created = self._sync_domain(domain)
+                if created:
+                    self.stdout.write(f' created (was missing) ...', ending='')
+                    catalog_alignment = True
+                self.stdout.write(' synced')
+            except Exception as e:
+                self.stdout.write(' failed')
+                msg = 'Error while processing {}: {}'.format(domain.name, e)
+                raise CommandError(msg)
+
+        if catalog_alignment:
+            call_command('align-catalog-zone')
+
+    @staticmethod
+    @transaction.atomic
+    def _sync_domain(domain):
+        created = False
+
+        # Create domain on pdns if it does not exist
+        try:
+            PDNSChangeTracker.CreateDomain(domain_name=domain.name).pdns_do()
+        except PDNSException as e:
+            # Domain already exists
+            if e.response.status_code == 409:
+                pass
+            else:
+                raise e
+        else:
+            created = True
+            PDNSChangeTracker.PDNSChange(domain_name=domain.name).update_catalog()
+
+        # modifications actually merged with additions in CreateUpdateDeleteRRSets
+        modifications = {(rrset.type, rrset.subname) for rrset in domain.rrset_set.all()}
+        deletions = {(rrset['type'], rrset['subname']) for rrset in pdns.get_rrset_datas(domain)} - modifications
+        deletions.discard(('SOA', ''))  # do not remove SOA record
+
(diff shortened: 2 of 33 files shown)


peterthomassen:
Ready for review. If the purpose of this in unclear, I can explain in a few words.

peterthomassen:
Open TODO: SOA changes due to DNSSEC resigning on Thursday nights are currently not reflected in the catalog.

nils-wisiol:
let's remove the serials and then merge this :+1:

peterthomassen:
I removed the serials from the catalog, and prepared the networking setup as discussed. I'll add final polishing for multi-slave replication and then we can merge.

api/desecapi/models.py
```diff
@@ -389,6 +394,16 @@ class Meta:
     def construct_name(subname, domain_name):
         return '.'.join(filter(None, [subname, domain_name])) + '.'
 
+    @classmethod
+    def construct_catalog_rrset(cls, subname, qtype, content):
```
nils-wisiol: This code is pdns specific and shouldn't live in models.py

api/desecapi/models.py
```diff
@@ -280,6 +280,11 @@ def _partitioned_name(self):
         subname, _, parent_name = self.name.partition('.')
         return subname, parent_name or None
 
+    def get_catalog_rrset(self, delete=False):
```
nils-wisiol: This code may be pdns specific and shouldn't live in models.py

api/desecapi/views.py
```diff
@@ -118,6 +120,19 @@ def perform_destroy(self, instance: models.Domain):
                 parent_domain.update_delegation(instance)
 
 
+class SerialList(generics.ListAPIView):
+    permission_classes = (IsVPNClient,)
+    throttle_classes = []  # don't break slaves when they ask too often (our cached responses are cheap)
+
+    def list(self, request):
```
nils-wisiol: this method contains a lot of boilerplate caching code, is there a better way to achieve this?

api/desecapi/management/commands/check-slaves.py
```diff
@@ -0,0 +1,80 @@
...
+
+from desecapi import pdns
+from desecapi.models import Domain
+
+
+def query_serial(zone, server):
+    query = dns.message.make_query(zone, 'SOA')
+    response = dns.query.tcp(query, server)
+
+    for rrset in response.answer:
```
nils-wisiol: this loop can be replaced by `response.find_rrset(dns.message.ANSWER, zone, dns.rdataclass.IN, dns.rdatatype.SOA)`; it may be necessary to update zone using `dns.name.from_text`

api/desecapi/management/commands/check-slaves.py
```diff
@@ -0,0 +1,80 @@
...
+        parser.add_argument('domain-name', nargs='*',
+                            help='Domain name to check. If omitted, will check all recently published domains.')
+        parser.add_argument('--delay', type=int, default=60, help='Delay SOA checks to allow pending AXFRs to finish.')
+        parser.add_argument('--window', type=int, default=settings.WATCHDOG_WINDOW_SEC,
+                            help='Check domains that were published no longer than this many seconds ago.')
+
+    def handle(self, *args, **options):
+        threshold = timezone.now() - timedelta(seconds=options['window'])
+        recent_domain_names = Domain.objects.filter(published__gt=threshold).values_list('name', flat=True)
+        serials = {zone: serial for zone, serial in get_serials().items() if zone.rstrip('.') in recent_domain_names}
```
nils-wisiol: only do this if `not options['domain-name']`

api/desecapi/management/commands/check-slaves.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if options['domain-name']:
```
nils-wisiol: possible `KeyError`?

api/desecapi/management/commands/check-slaves.py
```diff
@@ -0,0 +1,80 @@
+from datetime import timedelta
+from socket import gethostbyname
+from time import sleep
+
+from django.conf import settings
+from django.core.mail import get_connection, mail_admins
+from django.core.management import BaseCommand
+from django.utils import timezone
+import dns.message, dns.query, dns.rdatatype
```
nils-wisiol: missing requirement?

api/desecapi/management/commands/check-slaves.py
```diff
@@ -0,0 +1,80 @@
...
+                print(output[-1])
+                outdated_count += 1
+            else:
+                print(f'{zone} ok')
+
+        output.append(f'Checked {len(serials)} domains, {outdated_count} were outdated.')
+        print(output[-1])
+
+        if outdated_slaves:
+            subject = f'ALERT {len(outdated_slaves)} slaves out of sync'
```
nils-wisiol: use a template for the email body?

api/desecapi/management/commands/check-slaves.py
```diff
@@ -0,0 +1,80 @@
...
+            message = f'The following {len(outdated_slaves)} slaves are out of sync:\n'
+            for server in outdated_slaves:
+                message += f'* {servers[server]} ({server})\n'
+            message += '\n'
+            message += '\n'.join(output)
+
+            mail_admins(subject, message, connection=get_connection('django.core.mail.backends.smtp.EmailBackend'))
```
nils-wisiol: fast lane?

api/desecapi/management/commands/check-slaves.py
```diff
@@ -0,0 +1,80 @@
...
+
+        print('Sleeping for {} seconds before checking {} domains ...'.format(options['delay'], len(serials)))
+        sleep(options['delay'])
+
+        outdated_count = 0
+        outdated_slaves = set()
+        servers = {gethostbyname(server): server for server in settings.WATCHDOG_SLAVES}
+        output = []
+        for zone, local_serial in serials.items():
+            outdated = {}
```
nils-wisiol: consider moving the checking logic for one zone into a separate method

api/desecapi/models.py
peterthomassen: fix in 2f649f29870a7008e78608be28d0997ba20c0a98

api/desecapi/models.py
peterthomassen: fix in 2f649f29870a7008e78608be28d0997ba20c0a98

api/desecapi/pdns.py
```diff
@@ -104,3 +104,7 @@ def get_rrset_datas(domain):
              'records': [record['content'] for record in rrset['records']],
              'ttl': rrset['ttl']}
             for rrset in get_zone(domain)['rrsets']]
+
+
+def get_serials():
+    return {zone['name']: zone['edited_serial'] for zone in _pdns_get(NSMASTER, '/zones').json()}
```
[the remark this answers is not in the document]
peterthomassen: No, the `rrsets` field is not returned on `GET zones` anyways. Documentation was unclear, but I submitted a PR to pdns: https://github.com/PowerDNS/pdns/pull/8750

api/desecapi/permissions.py
```diff
@@ -1,3 +1,5 @@
+from ipaddress import IPv4Address, IPv4Network
```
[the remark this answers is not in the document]
peterthomassen: no, standard library: https://docs.python.org/3/library/ipaddress.html

api/desecapi/views.py
peterthomassen: no

api/desecapi/management/commands/check-slaves.py
peterthomassen: no, it's the argument parser sets it to an empty list if the argument is not given

api/desecapi/management/commands/check-slaves.py
peterthomassen: fix in ac92d3661201870f5069af1ee21518ebb8396e84

api/desecapi/management/commands/check-slaves.py
peterthomassen: agreed not to

api/desecapi/management/commands/check-slaves.py
peterthomassen: it's direct SMTP, no Celery queueing

api/desecapi/management/commands/check-slaves.py
peterthomassen: fixed in 0daa221da566747d7ca40e402ba5b76e93436827

api/desecapi/management/commands/sync-to-pdns.py
```diff
@@ -0,0 +1,92 @@
...
+                self._sync_domain(domain)
+                self.stdout.write(' synced')
+            except Exception as e:
+                self.stdout.write(' failed')
+                msg = 'Error while processing {}: {}'.format(domain.name, e)
+                raise CommandError(msg)
+
+    @staticmethod
+    @transaction.atomic
+    def _sync_domain(domain):
```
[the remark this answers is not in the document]
peterthomassen: We could; I can think of two ways:
- Extract the relevant code from the change tracker (move it somewhere to `pdns.py`)
- Keep it where it is, but give the pdns change tracker the option to only run `pdns_do()`, but not `api_do()`.

I'd rather do both of these at some later point, in combination with the Merkle tree replication catalog. That will require some refactoring in that area anyways.

api/desecapi/tests/test_domains.py
```diff
@@ -105,6 +105,11 @@ def test_can_register_public_suffixes_under_private_domains(self):
             self.assertNotRegistrable('b.private.public.suffix', user_c)
             self.assertRegistrable('b.private.public.suffix', user_b)
 
+    def test_cant_register_internal(self):
+        self.assertNotRegistrable('internal')
+        self.assertNotRegistrable('catalog.internal')
+        self.assertNotRegistrable('some.other.internal')
```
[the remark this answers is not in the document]
peterthomassen: fixed in c7e4b4a98b0f86ee3fc3ed8be1d68c833666a296

peterthomassen:
took care of all changes. Anything else before I squash and we can merge?

api/desecapi/tests/test_domains.py
```diff
@@ -109,6 +109,7 @@ def test_cant_register_internal(self):
         self.assertNotRegistrable('internal')
         self.assertNotRegistrable('catalog.internal')
         self.assertNotRegistrable('some.other.internal')
+        self.assertNotRegistrable('ending.in.a.dot.internal')
```
nils-wisiol: the trailing dot's missing :)

api/desecapi/management/commands/sync-to-pdns.py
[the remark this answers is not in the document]
nils-wisiol: Actually, that's what the PDNSChange classes are for. The whole `rrsets_meaty`, `rrsets_empty` construction can be replaced with instances of `PDNSChange` classes, for which then in a loop can be `pdns_do` be called. That also saves additional caveat of using "REPLACE" instead of "DELETE" for deletion (#220).

api/desecapi/pdns.py
```diff
@@ -103,3 +105,31 @@ def get_rrset_datas(domain):
              'records': [record['content'] for record in rrset['records']],
              'ttl': rrset['ttl']}
             for rrset in get_zone(domain)['rrsets']]
+
+
+def construct_catalog_rrset(zone=None, delete=False, subname=None, qtype='PTR', rdata=None):
+    # subname can be generated from zone for convenience; exactly one needs to be given
+    assert (zone is None or subname is None)
+    assert not (zone is None and subname is None)
```
nils-wisiol: alternative: assert (a is None) ^ (b is None)

api/desecapi/management/commands/check-slaves.py
```diff
@@ -47,34 +63,37 @@ def handle(self, *args, **options):
...
+        message = f'The following {len(outdated_slaves)} slaves are out of sync:\n'
+        for outdated_slave in outdated_slaves:
+            message += f'* {outdated_slave}\n'
+        message += '\n'
+        message += f'Current slave IPs: {self.servers}'
+        message += '\n'
+        message += '\n'.join(output)
 
-            mail_admins(subject, message, connection=get_connection('django.core.mail.backends.smtp.EmailBackend'))
+        mail_admins(subject, message, connection=get_connection('django.core.mail.backends.smtp.EmailBackend'))
```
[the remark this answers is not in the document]
nils-wisiol: found it

nils-wisiol:
I found some conversations that neither have changes, nor replies, nor are marked as resolved. Supposedly, they have been lost somewhere in the process. On my GitHub, they have been hidden in the middle of this page.

For the computation of JSON sent to the pdns API, the PDNSChange classes should be used. They not only save the duplication of the data structure definition, but also the hacky fix for #220. Otherwise, I only added minor annotations here and there.

api/desecapi/tests/test_domains.py
peterthomassen: `Domain.is_registrable()` assumes there is no trailing dot at the end. This is because the function is called from within `DomainSerializer.validate_name()` which runs only after the name regex has been applied. Outside of tests, it cannot happen that the argument of `Domain.is_registrable()` has a dot at the end.

If it happens in a test, `Domain.is_registrable()` throws an uncaught exception. We may or may not want to fix this, but in any case it has nothing to do with the commits in this PR. If you think it's worth fixing, let's open an issue.

api/desecapi/pdns.py
peterthomassen: nice! fixed

api/desecapi/management/commands/sync-to-pdns.py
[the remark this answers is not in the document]
peterthomassen: fixed! Now it is much much nicer, thank you for the suggestion.

peterthomassen:
I finalized everything and squashed accordingly. Can you please take a final look at 5a7225145896c0e6f763fb1a95794e445e582e3d? thx!

nils-wisiol:
oh, don't forget to squash :)

PR #286, opened 2020-01-10
Improve Error Handling

nils-wisiol:
This PR improves error handling in various places, and catches exceptions that are caused when concurrent requests cause double insertions. Such special cases were not handled correctly as time-of-check and time-of-action differ in our treatment of RR set data.

Errors caused in such situations will now be coerced in 429 Too Many Requests, potentially hiding bugs that the change tracker may have.


test/e2e/spec/api_spec.js
```diff
@@ -379,6 +379,38 @@ describe("API v1", function () {
...
+                                }
+                            )).to.have.status(400);
+                        });
+
+                        it("still does not accept a semantic duplicates", function () {
+                            return expect(chakram.post(
+                                '/domains/' + domain + '/rrsets/',
+                                {
+                                    'subname': 'duplicate-contents', 'type': 'AAAA',
+                                    'records': ['::1', '::0001'], 'ttl': 60
```
peterthomassen: This should be rejected regardless of `::1` being created in `before()`, because duplicate content in an RRset is not allowed even when fresh. Are we testing that as well?

api/desecapi/pdns_change_tracker.py
```diff
@@ -225,11 +236,14 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
                     axfr_required.add(change.domain_name)
-            except Exception as e:
-                # TODO gather as much info as possible
-                #  see if pdns and api are possibly in an inconsistent state
+            except PDNSValidationError as e:
                 self.transaction.__exit__(type(e), e, e.__traceback__)
                 raise e
+            except Exception as e:
+                self.transaction.__exit__(type(e), e, e.__traceback__)
+                exc = ValueError('For changes %s, %s occured when applying %s' %
```
peterthomassen: Should be `RuntimeError`

api/desecapi/tests/test_pdns_change_tracker.py
```diff
@@ -24,6 +24,13 @@ def assertPdnsZoneUpdate(self, name, rr_sets):
             self.request_pdns_zone_axfr(name),
         ])
 
+    def test_rrset_does_not_exist_exception(self):
+        tracker = PDNSChangeTracker()
+        tracker.__enter__()
+        tracker._rr_set_updated(RRset(domain=self.empty_domain, subname='', type='A'))
```
peterthomassen: nice mock!

api/desecapi/renderers.py
```diff
@@ -1,4 +1,5 @@
 from rest_framework import renderers
+from rest_framework.utils.serializer_helpers import ReturnList
```
peterthomassen: Unused line of code

api/desecapi/exceptions.py
```diff
@@ -25,3 +25,9 @@ def __init__(self, response=None):
 class PDNSException(APIException):
     def __init__(self, response=None):
         return super().__init__(f'pdns response code: {response.status_code}, pdns response body: {response.text}')
+
+
+class ConcurrencyException(APIException):
+    status_code = status.HTTP_429_TOO_MANY_REQUESTS
+    default_detail = 'Too many concurrent requests.'
+    default_code = 'too_many_requests'
```
peterthomassen: Better: `too_many_concurrent_requests`; the other value sounds like it's a result of rate limiting.

The reason is actually not that it is two many requests in the sense that could configure a limit other than 2 concurrent requests. The reason really is that a concurrency inconsistency / race condition is happening. So, you about `default_code = 'concurrency_race'`, `'concurrency_conflict`', `'concurrency_inconsistency'` or similar?

api/desecapi/views.py
```diff
@@ -323,6 +324,16 @@ def get(self, request, *_):
...
+            if isinstance(e.detail, list) and \
+                    any([
+                        any([
+                            getattr(non_field_error, 'code', '') == 'unique'
+                            for non_field_error
+                            in err.get('non_field_errors', [])
+                        ])
+                        for err in e.detail
+                    ]):
+                raise ConcurrencyException from e
```
peterthomassen: I suppose this happens when a dynDNS update triggers creation of an RRset that was already created by another concurrent request (and the change tracker did not see it yet when it calculated its plan of action). I think a comment explaining that would help here.

However, shouldn't this be part of `serializer.is_valid()`? Why is it on view level?

api/desecapi/renderers.py
```diff
@@ -11,11 +12,23 @@ def render(self, data, media_type=None, renderer_context=None):
...
-            # TODO enable the renderer to handle responses where reponse.data is a list
-            if not isinstance(data, dict) or data.get('detail', None) is None:
-                raise ValueError('Expected response.data to be a dict with error details in response.data[\'detail\'], '
-                                 'but got %s:\n\n%s' % (type(response.data), response.data))
             response['Content-Type'] = 'text/plain'
-            return data['detail']
+
+            try:
+                return data['detail']
+            except KeyError:
```
peterthomassen: If `data` is a list of dicts, this raises `TypeError` which is not caught here. This would immediately raise and not run the following code. That is probably not intended. Is my understanding correct?

api/desecapi/renderers.py
```diff
@@ -11,11 +12,23 @@ def render(self, data, media_type=None, renderer_context=None):
...
+                pass
+
+            try:
+                details = [el['detail'] for el in data if getattr(el, 'detail', None) is not None]
```
peterthomassen: redundant, as the `'detail'` is looked up twice from `el` if `not None`. Also, `detail` is not an attribute of `el` (it is a dictionary key), so you probably mean `el.get('detail')` instead of `getattr`. How about:

`details = list(filter(None, [el.get('detail') for el in data]))`

See the note in https://docs.python.org/3/library/functions.html#filter

(PS: obfuscation contest: `list(filter(None, list(map(lambda x: x.get('detail'), data))))`)

api/desecapi/renderers.py
```diff
@@ -11,11 +12,23 @@ def render(self, data, media_type=None, renderer_context=None):
...
+                if details:
+                    return ', '.join(details)
+            except TypeError:
```
peterthomassen: If we assume the structure of the elements in the list case to be the same as the structure of the single case, then we should also be checking for `KeyError` here (because we do so above)

api/desecapi/pdns_change_tracker.py
```diff
@@ -31,6 +31,10 @@ class PDNSChangeTracker:
     `modifications` and `deletions` are guaranteed to be disjoint.
     - If an item is in the set of deletions while being modified, an exception is raised.
     - If an item is in the set of modifications while being deleted, it is removed from `rr_set_modifications`.
+
+    Note every change tracker object will track all changes to the model across threading.
+    To avoid side-effects, it is recommended to run only one change tracker at a time, i.e.
+    do not use them in parallel (in a multi-threading scenario), do not use them nested.
```
peterthomassen: Is the problem at hand really related to threading? If so, I don't understand what's happening: Our [uwsgi configuration](https://github.com/desec-io/desec-stack/blob/2c20d11/api/uwsgi.ini#L6) has `threads = 1`.

api/desecapi/views.py
nils-wisiol: The serializer can't handle this by itself as the `instances` argument -- given to it by the view -- can be outdated due to a concurrent request.

api/desecapi/exceptions.py
nils-wisiol: I like `concurrency_conflict`

test/e2e/spec/api_spec.js
nils-wisiol: yes, in lines 368 to 381

api/desecapi/renderers.py
nils-wisiol: Yes, fixed

api/desecapi/pdns_change_tracker.py
nils-wisiol: Multiple tests fail when this is `RuntimeError` (see commit history)

api/desecapi/pdns_change_tracker.py
nils-wisiol: It's not, but the note is still true

nils-wisiol:
I rebased this on top of master, please have another look at the merge I did in 0063b2f2ba4ed648c49dca00a2e839c2864ef0b9

PR #289, opened 2020-01-21
fix(nslord,nsmaster): make ACLs more explicit

peterthomassen:


peterthomassen:
incorporating in #284

PR #295, opened 2020-02-04
feat(www): compress /api/v1/serials/ endpoint

peterthomassen:
Tests have shown that size after compression is around 23%.
Python "requests" automatically asks for gzip, so no need for
change on the slave.

This branch is sandbox-tested.


peterthomassen:
Good point! Array variable definitions (like `uwsgi_params`, `proxy_params`) are inherited; `uwsgi_pass` and `proxy_pass` are not. ([Source](https://stackoverflow.com/a/32126596/6867099)) I also checked that it works as intended.

PR #297, opened 2020-02-06
Add monitoring

peterthomassen:
This adds a Prometheus container which can be accessed through `www` (with basic auth). To keep the PR small, monitoring currently only includes Prometheus itself, and generic metrics from Django.

Next step is metrics for nginx and OpenVPN, but I guess this PR can be reviewed/merged independently while I work on these.


peterthomassen:
I added basic nginx monitoring.

docker-compose.yml
```diff
@@ -281,10 +284,39 @@ services:
         tag: "desec/openvpn-server"
     restart: unless-stopped
 
+  prometheus:
+    image: prom/prometheus:latest
+    init: true
+    volumes:
+    - ./prometheus/conf:/etc/prometheus:ro
+    - prometheus:/prometheus
```
nils-wisiol: is there a difference in `./prometheus/` (two lines above) and `prometheus/` (line above)?

nils-wisiol:
I confirmed it working on my machine

docker-compose.yml
peterthomassen: Yes! `prometheus` is the volume defined further down in the file, and `./prometheus` is a directory relative to the current directory. The volume carries persistent data, the other one is for mounting configuration.

peterthomassen:
> Is there any data that needs to be deleted eventually? Otherwise, this lgtm

The [default retention time is 15 days](https://prometheus.io/docs/prometheus/latest/storage/#operational-aspects). We can adjust it later as we see fit.

PR #303, opened 2020-02-07
feat(openvpn-server): add Prometheus monitoring

peterthomassen:


peterthomassen:
tested on sandbox, works correctly.

PR #306, opened 2020-02-09
fix(api): correctly handle POST/PUT/PATCH with empty payload

peterthomassen:
This fixes the bug reported to our support mailbox today at 09:02.

### DIFF
--- a/api/desecapi/serializers.py
+++ b/api/desecapi/serializers.py
@@ -227,6 +227,8 @@ class RRsetSerializer(ConditionalExistenceModelSerializer):
     @classmethod
     def many_init(cls, *args, **kwargs):
         domain = kwargs.pop('domain')
+        # Note: We are not yet deciding the value of the child's "partial" attribute, as its value depends on whether
+        # the RRSet is created (never partial) or not (partial if PATCH), for each given item (RRset) individually.
         kwargs['child'] = cls(domain=domain)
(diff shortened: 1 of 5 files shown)


nils-wisiol:
I confirmed that the error here was the faulty setting of the `many` parameter. According to our docs, PUT and PATCH are exclusively for bulk operations; for POST, we accept bulk or single payload.

Note that the `many` parameter by virtue of the `__new__` method decides the serializer class that is instantiated: if `many` is True, `RRsetListSerializer` is used; otherwise, `RRsetSerializer` is instantiated. That was the source of the problem: `RRsetSerializer` defines the `UniqueTogetherSerializer`, which cannot handle a `QuerySet` as the serializer's instance, and then throws an error.

I could not figure out why the old version insists that `data` is truly. Even assuming that this was meant to read `data is not None`, I don't see how it could make sense.

I extended the tests to cover empty lists and simplified the handling of the request in the view, leaving it up to the serializer to provide a correct error message when payload was faulty. (I confirmed GET requests set many=True; other methods are undefined on the view and will be rejected.)

The remaining difference now is that this new version overrides the many flag for POST, PATCH, and PUT methods and ignores the many parameter to `get_serializer` for those methods. In the old version, this was not the case. However, this property was dead in the sense that none of the methods invoking `get_serializer` actually provided a `many` value for these methods.

PR #307, opened 2020-02-12
20200212 restore security headers

peterthomassen:


peterthomassen:
thx. - I modified them slightly, removing Mailchimp and YouTube external references as we don't use them anymore.

PR #311, opened 2020-02-25
feat(www): disable TLS < 1.2, enable 1.3, improve ciphers, closes #261

### LINKED ISSUE #261: Enforce more strict TLS
Preliminary TLS log analysis shows that most users TLSv1.2:

```
$ grep desec/www | grep "\.dedyn\.io" | egrep -o TLS.* | sort | uniq -c | sort -n
      3 TLSv1.2/ECDHE-RSA-AES256-SHA384
      6 TLSv1/AES256-SHA
     21 TLSv1.2/ECDHE-RSA-AES128-SHA256
     64 TLSv1/ECDHE-RSA-AES256-SHA
   2145 TLSv1.2/ECDHE-RSA-AES256-GCM-SHA384
```

Ths TLSv1 requests (about 3%) have the following user agent breakdown:

```
$ grep TLSv1/ | awk '{print $18}' | sort | uniq -c
     34 "-"
     38 "ddclient/3.8.1"
      2 "Wget/1.10.2"
      2 "Wget/1.10+devel"
      2 "Wget/1.11.4"
```

These are really old version that we can easily stop supporting (after informing users).

We should also take the chance and strictly require TLS on update6.dedyn.io. (There is now ~1 non-TLS request per day on this domain.)

It is more difficult for update.dedyn.io which has 10-15% non-TLS requests.

peterthomassen:
We currently are capped to B+ by SSL Labs, due to the support for TLS < 1.2.


peterthomassen:
(tested on sandbox, SSL Labs is very satisfied with this configuration)

peterthomassen:
Regarding 89414527b56fa6ec0fa3cddbb4c17c20d4bee84a, let's give it a try and revert if it breaks things.

I'd actually prefer disabling port 80 also for `update.dedyn`, but I guess that would break too many setups: 8.3% of all requests on that host are still via port 80. According to #261, it used to be 10-15% just a few months ago. Hopefully we can do it later this year, then.

PR #314, opened 2020-03-17
feat(webapp): adds Terms of Use page

nils-wisiol:
Terms of Use need some discussion I suppose.

As of now, the sandbox has a [rendered version](https://desec.sandbox.dedyn.io/terms).


api/api/settings.py
```diff
@@ -195,6 +195,14 @@
 }
 
 # user management
+AUTH_PASSWORD_VALIDATORS = [
+    {
+        'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
+        'OPTIONS': {'min_length': 8}
+    },
+    {'NAME': 'django.contrib.auth.password_validation.CommonPasswordValidator'},
+    {'NAME': 'django.contrib.auth.password_validation.NumericPasswordValidator'},
```
nils-wisiol: Doesn't this introduce more password rules than just the min length?

api/desecapi/management/commands/check-slaves.py
```diff
@@ -63,11 +66,17 @@ def handle(self, *args, **options):
         outdated_slaves = set()
 
         output = []
+        timeouts = {}
         for zone, local_serial in serials.items():
             outdated_serials = self.find_outdated_servers(zone, local_serial)
-            outdated_slaves.update(outdated_serials.keys())
+            for server, serial in outdated_serials.items():
+                if not serial and serial is not None:  # serial is False (error condition, instead of "not found"),
```
nils-wisiol: how about "if serial is False"?

api/api/settings.py
peterthomassen: yes - I figured that doesn't hurt, although is non-critical functionality, so I didn't write a test. I'm fine dropping it. What do you think?

api/desecapi/management/commands/check-slaves.py
peterthomassen: I thought so too, but then I found https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations:

> Don't compare boolean values to True or False using ==.
> 
> ```
> Yes:   if greeting:
> No:    if greeting == True:
> Worse: if greeting is True:
> ```

mh.

api/api/settings.py
nils-wisiol: It should be either dropped or documented

api/desecapi/management/commands/check-slaves.py
nils-wisiol: The rule does not apply, as it is not (strictly) a Boolean value. No?

nils-wisiol:
I made the commits of your last push ready to merge by squashing them and adjusting commit messages. I then added a couple fixup commits that require review by you:

- 73f441559791faede84288e7759d180aeb5d4ea2 and 573e8a4f6ffb5011cca6da70188368d2cff546af use a [more "vuetify way"](https://vuetifyjs.com/en/styles/flex/#flex-align-self) of making v-card full-height
- 04743bb260a59dbf5ae8ea7c238443c61278aee8 generalizes the retention time to all "known" certificates. This probably needs relaxation, as this time can be quite long, what I have in mind is a warning to the user that registers the domain. But for now, this makes sense to me as it shows awareness on our side while at the same time, we do not have any implementation of deletion policy.
- 352e6143447e9b294f7a5929d424802706c1d5af and 7e6afdb008e8cfec45a9292bd67973cc1d0aed70 remove dead code (hopefully without merge conflict!)
- e99a85df49e988c8ea07ad3be5ca42ffa92b764f adds a note what to do when the activation link has expired

I moved the "smarten signup" commit to a new branch, not because I don't like but, but so this is not blocked anymore. Please open a separate PR for that one; two thoughts:

1. bug: redirect URL to the welcome page contains the wrong domain name for "managed DNS"
1. optics: let's checkout if [tabs](https://vuetifyjs.com/en/components/tabs/) could look better than radio buttons

api/api/settings.py
peterthomassen: dropped

api/desecapi/management/commands/check-slaves.py
peterthomassen: fair enough, changed according to your suggestion

PR #321, opened 2020-03-28
20200328 improve signup

peterthomassen:
From your [earlier feedback](https://github.com/desec-io/desec-stack/pull/314#issuecomment-605428852):

> bug: redirect URL to the welcome page contains the wrong domain name for "managed DNS"

Fixed

> optics: let's checkout if tabs could look better than radio buttons

I do not think this is a good solution for two reasons:
a) We can't have tabs on the main page where you can already preselect the account type, before opening the modal. I think though that the UI element to specify the account type should be consistent whenever it appears, i.e. if we have a radio button on the main page, it should also be a radio button in the modal.
b) Different tabs usually are for things that are not mutually exclusive, but instead are not of interest at all times (such as a typical options dialog). Tabs therefore usually have independent content. In our case, the tab contents would be nearly identical. Overall, I'd say they are not associated with "making a choice", but with "dealing with another aspect".

If you still would like to see the comparison, I can of course implement it.

-- 

I also fixed a bug, which had caused scrolling to the top when changing the account type selection on the main page. The new behavior is that when navigation occurs which leads to a destination that only differs in the query string, scrolling is not done. (When the query string is removed, that's an exception -- like when clicking on the deSEC logo, you want to see things scroll to the top.)

### DIFF
--- a/webapp/src/App.vue
+++ b/webapp/src/App.vue
@@ -47,7 +47,7 @@
           <v-icon :color="item.post_icon_color" class="ml-1 text--darken-1" small v-if="item.post_icon">{{item.post_icon}}</v-icon>
         </span>
       </div>
-      <v-btn class="mx-4 mr-0" color="primary" depressed :to="{name: 'signup'}">Create Account</v-btn>
+      <v-btn class="mx-4 mr-0" color="primary" depressed :to="{name: 'signup', query: $route.query}">Create Account</v-btn>
       <v-app-bar-nav-icon class="d-md-none" @click.stop="drawer = !drawer" />
     </v-app-bar>
 
--- a/webapp/src/router/index.js
+++ b/webapp/src/router/index.js
@@ -73,7 +73,11 @@ const routes = [
 const router = new VueRouter({
   mode: 'history',
   base: process.env.BASE_URL,
-  scrollBehavior () {
+  scrollBehavior (to, from) {
+    // Skip if destination full path has query parameters and differs in no other way from previous
+    if (from && Object.keys(to.query).length) {
+      if (to.fullPath.split('?')[0] == from.fullPath.split('?')[0]) return;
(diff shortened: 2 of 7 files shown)


nils-wisiol:
I adjusted the text with two purposes in mind:

1. make it easier to understand if you don't know what managed DNS is
2. also communicate that *.dedyn.io domains can be used for managed DNS and non-dedyn.io domains can be used for dynDNS, too.

Both points could use further improvement.

Also adapted the form to choose the "dynDNS" radio button by default, if no parameter in the URL is given. Otherwise, sending the form without choosing has the same effect as choosing none. However, choosing none as default seemed odd.

Finally, a technical note, the code is very diverse in terms of variable/function name style, even before this PR. vuejs [seems to prefer camelCase](https://vuejs.org/v2/style-guide/) for JS (ctrl-f camelCase), though no explicit statement is made -- should we follow this in the future?

peterthomassen:
> I adjusted the text with two purposes in mind:

Settled as discussed.

> Also adapted the form to choose the "dynDNS" radio button by default, if no parameter in the URL is given. Otherwise, sending the form without choosing has the same effect as choosing none. However, choosing none as default seemed odd.

:+1:

> Finally, a technical note, the code is very diverse in terms of variable/function name style, even before this PR. vuejs [seems to prefer camelCase](https://vuejs.org/v2/style-guide/) for JS (ctrl-f camelCase), though no explicit statement is made -- should we follow this in the future?

Yes, let's do that.

I squashed, and I think this ready for merge. lgty?

In #328, I will add a commit that advertises talk.desec.io in the confirmation views, and on that occasion I'll add the display of DNSSEC keys for custom domains after signup (so this is not part of this PR).

nils-wisiol:
ah, don't forget to squash once more

PR #323, opened 2020-03-29
feat(): adds GitHub CI integration

nils-wisiol:


nils-wisiol:
The fixup needs to be squashed and the two middle commits need to be removed before merging.

Running the tests on GitHub takes currently about 7 minutes.

What's missing are the logs. For some reason docker-compose is using the syslog.

nils-wisiol:
After a couple of unsuccessful tries, I removed the logs files in the interest of time. Once we have a problem that can't be locally reproduced, we can worry about how to extract more debugging info.

I confirmed that failures in API and/or e2e tests are recognized by the GitHub Action.

PR #326, opened 2020-03-31
20200331 api multiprocess prometheus

peterthomassen:
I verified in my local deployment that the metrics numbers now indeed are the sum across workers.

### DIFF
--- a/api/api/settings.py
+++ b/api/api/settings.py
@@ -214,6 +214,18 @@ CAPTCHA_VALIDITY_PERIOD = timedelta(hours=24)
 WATCHDOG_SLAVES = os.environ.get('DESECSTACK_WATCHDOG_SLAVES', '').split()
 WATCHDOG_WINDOW_SEC = 600
 
+# Prometheus (see https://github.com/korfuri/django-prometheus/blob/master/documentation/exports.md)
+#  TODO Switch to PROMETHEUS_METRICS_EXPORT_PORT_RANGE instead of this workaround, which currently necessary to due
+#  https://github.com/korfuri/django-prometheus/issues/215
+try:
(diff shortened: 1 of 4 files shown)


api/api/settings.py
```diff
@@ -214,6 +214,17 @@
 WATCHDOG_SLAVES = os.environ.get('DESECSTACK_WATCHDOG_SLAVES', '').split()
 WATCHDOG_WINDOW_SEC = 600
 
+# Prometheus
```
nils-wisiol, line 217: Add "TODO" so that this can hopefully be resolved one day

api/api/settings.py
```diff
@@ -214,6 +214,17 @@
...
+# As per https://github.com/korfuri/django-prometheus/blob/master/documentation/exports.md while we can't use
+# PROMETHEUS_METRICS_EXPORT_PORT_RANGE due to https://github.com/korfuri/django-prometheus/issues/215
+try:
+    import prometheus_client
+    import uwsgi
+    prometheus_client.values.ValueClass = prometheus_client.values.MultiProcessValue(
+        _pidFunc=uwsgi.worker_id)
+except ImportError:
+    pass  # not running in uwsgi
```
nils-wisiol, line 226: should this be narrower to avoid false silence of errors?

api/api/settings.py
peterthomassen: done

api/api/settings.py
peterthomassen: tried to do something smarter in the fixup commit - ok now?

nils-wisiol: right, the clause was too large. lgtm!

PR #327, opened 2020-03-31
20200331 ratelimit checkip

peterthomassen:


peterthomassen:
HSTS header is still present, it comes from the inclusion of global.conf. Previously, it need to be re-stated explicitly as we were also setting another header, which invalidates headers set in other scopes. As we're not setting any other headers in that location anymore, headers from global.conf remain in effect.

PR #328, opened 2020-04-02
20200402 talk.desec.io

peterthomassen:


nils-wisiol:
Two quick thoughts here

1. we're hardcoding desec.io, which is bad, but not hardcoding it also does not seem to make sense
2. there should be advertisement on `DynSetup.vue` and possible on `Confirmation.vue`

peterthomassen:
> 1. we're hardcoding desec.io, which is bad, but not hardcoding it also does not seem to make sense

I think hardcoding is fine in this case. - At another time, I was thinking that the templates could be made part of a volume, so that they don't strictly belong to the codebase anymore. Do you think it's the time to make this change now?

> 2. there should be advertisement on `DynSetup.vue` and possibly on `Confirmation.vue`

Good point, Will add it.

peterthomassen:
I added advertisement of talk.desec.io in the confirmation views, plus a separate confirmation view for sign-up with a non-dyn domain which displays DS records.

nils-wisiol:
I squashed your changes and added a small fixup. Please review that guy, afterwards this lgtm. Let's not worry too much about the grey color ... :woman_shrugging:

PR #331, opened 2020-04-05
20200405 record noop dyn update

peterthomassen:


peterthomassen:
We need "touched" for determining how long a dedyn.io domain has not been touched. (This is what "updated" in this PR implements.)

Do we need "changed" for anything? It is what "updated" did before this PR, but it was undocumented. If we don't have a use case, should we really implement it? I'm wondering about the general "why", and then there's also the usual "data austerity" argument.

nils-wisiol:
I don't think data austerity should apply here. Touched vs. updated can provide both us and the user with debugging information; but I cannot imagine it's revealing some information that the user rather does not want to reveal. Just as an example, maybe it can help us measuring the frontend response times, eventually?

peterthomassen:
I renamed `.updated` to `.touched`, but did not implemented `.changed` yet. Do you think we should do it in this PR?

api/desecapi/models.py
```diff
@@ -281,6 +281,16 @@ def keys(self):
             self._keys = pdns.get_keys(self)
         return self._keys
 
+    @property
+    def touched(self):
+        try:
+            rrset_touched = max(updated for updated in self.rrset_set.values_list('touched', flat=True))
+            # If the domain has not been published yet, self.published is None and max() would fail
+            return rrset_touched if not self.published else max(rrset_touched, self.published)
+        except ValueError:
```
nils-wisiol, line 290: Where does this ValueError come from? I understood comment above such that `max(None, some_timestamp)` would raise, but that is covered by the branching. Any other source of ValueErrors here?

peterthomassen: The `ValueError` comes from the `rrset_touched` assignment, if the `max()` argument iterable has no elements (i.e. there are no records). In this case, `max()` raises, we catch the exception and return `self.published` (which may or may not be `None` as well).

PR #342, opened 2020-04-14
20200414 fix update6 tls

peterthomassen:
no comment -__-


peterthomassen:
It's not there because update6 is IPv6 only. I did it the same way as further down in the same file for port 443

PR #354, opened 2020-04-23
Update Randomness Sources

nils-wisiol:
This changes the source of randomness for

1. NSEC3 param salt (for new domains), from `random` to `secrets`,
2. authentication tokens, from `urandom` to `secrets`.

Rationale:

1. Creating domains is expensive anyhow, we can afford a little more time spend on the salt generation. 
2. We abstract from the OS and use Python's interface for generating secrets instead of having to worry about the OS interface (like `/dev/random` vs `/dev/urandom`).

### DIFF
--- a/api/desecapi/models.py
+++ b/api/desecapi/models.py
@@ -9,7 +9,6 @@ import uuid
 from base64 import urlsafe_b64encode
 from datetime import timedelta
 from hashlib import sha256
-from os import urandom
 
 import psl_dns
 import rest_framework.authtoken.models
@@ -183,7 +182,7 @@ class Token(ExportModelOperationsMixin('Token'), rest_framework.authtoken.models
     plain = None
 
     def generate_key(self):
-        self.plain = urlsafe_b64encode(urandom(21)).decode()
+        self.plain = secrets.token_urlsafe(21)
         self.key = Token.make_hash(self.plain)
         return self.key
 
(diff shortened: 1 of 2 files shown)


api/desecapi/models.py
```diff
@@ -179,7 +178,7 @@ class Token(ExportModelOperationsMixin('Token'), rest_framework.authtoken.models
     plain = None
 
     def generate_key(self):
-        self.plain = urlsafe_b64encode(urandom(21)).decode()
+        self.plain = urlsafe_b64encode(secrets.token_bytes(21)).decode()
```
peterthomassen, line 181: If you use `secrets.token_urlsafe()`, you can save the `urlsafe_b64encode()` and the `decode()` call

api/desecapi/pdns_change_tracker.py
```diff
@@ -82,7 +82,7 @@ def axfr_required(self):
             return True
 
         def pdns_do(self):
-            salt = '%016x' % random.randrange(16 ** 16)
+            salt = '%016x' % secrets.randbelow(16 ** 16)
```
peterthomassen, line 85: should be equivalent to the more clear `secrets.token_hex(8)  # 16 hex chars`

api/desecapi/models.py
nils-wisiol: done

api/desecapi/pdns_change_tracker.py
nils-wisiol: done

PR #366, opened 2020-04-27
Enable OCSP Stapling

Knight1:
The ssl_trusted_certificate is the intermediate from the used Root Certificate Authority. 
I guess that most people would use just one CA instead of different ones. 

The other solution would be to introduce a new type of file ex. $DOMAIN_intermediate.crt in https://github.com/desec-io/desec-stack/blob/master/www/entrypoint.sh and move the ssl_trusted_certificate to the server {} block.

IN the case of LE the file is https://letsencrypt.org/certs/lets-encrypt-x3-cross-signed.pem


nils-wisiol:
Thanks for this contribution. [Let's Encrypt recommends OCSP stapling](https://letsencrypt.org/docs/integration-guide/#implement-ocsp-stapling), not sure why we don't have it yet.

Your changes broke the CI tests, and unfortunately, our CI is not configured to post logs in case of failure (cf #367). I pulled your changes and started locally, the log says

    nginx: [emerg] SSL_CTX_load_verify_locations("/etc/nginx/${CERT_PATH}issuer_intermediate.crt") failed (SSL: error:02001002:system library:fopen:No such file or directory:fopen('/etc/nginx/${CERT_PATH}issuer_intermediate.crt','r') error:2006D080:BIO routines:BIO_new_file:no such file error:0B084002:x509 certificate routines:X509_load_cert_crl_file:system lib)

The reasons seems to be that `www/sites-available/envreplace.sh` magic is not applied to `ssl.conf`. So either, the path name needs to constant, or we need to extend envreplace magic.

nils-wisiol:
Also, please make the commit message something like "feat(www): enable OCSP stapling". We shall [add a contributions file](https://github.com/desec-io/desec-stack/issues/368) shortly explaining those policies.

peterthomassen:
I considered OCSP stapling in the past and found that it is broken beyond practical use, see these articles:

- https://blog.hboeck.de/archives/886-The-Problem-with-OCSP-Stapling-and-Must-Staple-and-why-Certificate-Revocation-is-still-broken.html
- https://medium.com/@alexeysamoshkin/how-ssl-certificate-revocation-is-broken-in-practice-af3b63b9cb3

Now it's possible that the issues described in the articles are no longer relevant (maybe servers are fixed by now), or overstated. @Knight1, do you have more recent insights on the topic?

Darkspirit:
https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&guideline=5.4 recommends OCSP stapling with reason.
If you don't enable OCSP stapling, browsers connect directly to the CA's OCSP server, which is a privacy problem. Yes, nginx still needs https://trac.nginx.org/nginx/ticket/812 to be perfect.

peterthomassen:
Frankly, I don't really think that it's our responsibility when browser make requests that impair privacy (same with DoH or generally DNS resolution). But fair enough, why not use OCSP Stapling if it's working well enough. I updated myself on the state of OCSP Stapling in nginx, and it's actually in better shape that I remember (it was Apache that had more issues). -- So, I'm fine with including this.

We should move `ssl.conf` to `ssl.conf.var` and include that file in the `envreplace.sh` step. This script is currently under `conf/sites-available/`; we probably should move it to `conf/` and replace the globbing in the `for` statement with a search into subdirectories, e.g. using `find`. Finally, we should update the README to make people aware that the issuer intermediate certificate has to be included in the certs directory.

@Darkspirit, would you like to improve the PR with these changes if you find a bit of time?

peterthomassen:
@Knight1 Are you still interested in finishing this? If not, please let us know and we'll pick it up later.

Knight1:
Yeah, sure! But i am a bit shy to break something and i am waiting for https://github.com/desec-io/desec-stack/issues/368 but if you are fine at this moment i would rewrite the code and push it to the branch.

peterthomassen:
Thanks for getting back! No need to wait for the code contribution guidelines. In essence, we please make sure that your PR can be cleanly applied to master (i.e., when your work is done, please rebase), and use semantic commit messages with a commit title that describes the meaning of the change (not just "change this or that file"). To see the format we use, take a look at other commits in master (especially in the www subdirectory, which is where you're contributing).

peterthomassen:
@Knight1 ping

Knight1:
I am on it :)

Had a few questions in mind before i go to sleep. 

For OCSP to work we need the LE Chain back to the Intermediate Certificate. This is currently the R3. I could not figure out if you include it in the .cer File, but we would need it. We could do that manually but this would be a real hassle if LE switches there Intermediate, would not be automated .. :s

Let's see what Travis has to report.

nils-wisiol:
I think you can safely assume that the certificate file that nginx receives contains the certificate chain

www/conf/envreplace.sh
```diff
@@ -0,0 +1,15 @@
...
+then
+    export PROD_ONLY='#'
+fi
+
+for file in www/conf/sites-available/*.var www/conf/donf.d/*.var; do
+    # we only replace occurances of the variables specified below as first argument
+    (envsubst "$DESECSTACK_IPV4_REAR_PREFIX16" |
+    envsubst "$DESECSTACK_DOMAIN" |
+    envsubst "$CERT_PATH" |
+    envsubst "$PROD_ONLY" ) < "$file" > "$DIR"/$(basename "$file" .var)
```
peterthomassen, line 14: `$DIR` is undefined

www/conf/envreplace.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+export PROD_ONLY=
+if [[ $DESECSTACK_E2E_TEST = "TRUE" ]]
+then
+    export PROD_ONLY='#'
+fi
+
+for file in www/conf/sites-available/*.var www/conf/donf.d/*.var; do
```
peterthomassen, line 9: typo

peterthomassen (commented):
If you move `envreplace.sh`, you also have to adjust where it's called (in `www/entrypoint.sh`)

Knight1:
The Path is based on the Dockerfile and fixed to /etc/nginx/. I do not think that someone would actually change that so I used the path hardcoded.

www/conf/envreplace.sh
```diff
@@ -0,0 +1,15 @@
...
+then
+    export PROD_ONLY='#'
+fi
+
+for file in /etc/nginx/sites-available/*.var /etc/nginx/conf.d/*.var; do
+    # we only replace occurances of the variables specified below as first argument
+    (envsubst "$DESECSTACK_IPV4_REAR_PREFIX16" |
+    envsubst "$DESECSTACK_DOMAIN" |
+    envsubst "$CERT_PATH" |
+    envsubst "$PROD_ONLY" ) < "$file" > /etc/nginx/$(basename "$file" .var)
```
peterthomassen, line 14: All files now get written to /etc/nginx/, even when the .var file was in a subdirectory.

peterthomassen:
True, nobody will change the path :) I liked the $DIR approach because the script is more easily reusable then. We have the same logic e.g. in dbmaster/initdb.d/00-init.sh (similarly in dblord).

peterthomassen:

nils-wisiol (approved):

PR #379, opened 2020-05-04
BREAKING feat(nslord,nsmaster): upgrade to pdns 4.3

peterthomassen:
This comes with a mandatory database migration on dblord and dbmaster,
https://github.com/PowerDNS/pdns/blob/rel/auth-4.3.x/modules/gmysqlbackend/4.2.0_to_4.3.0_schema.mysql.sql

The commit also contains two small fixes in the database schema that
are not mandatory to apply.

### DIFF
--- a/dblord/initdb.d/10-pdns-lord.sql
+++ b/dblord/initdb.d/10-pdns-lord.sql
@@ -6,7 +6,7 @@ CREATE TABLE domains (
   master                VARCHAR(128) DEFAULT NULL,
   last_check            INT DEFAULT NULL,
   type                  VARCHAR(6) NOT NULL,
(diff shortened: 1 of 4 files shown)


peterthomassen:
fyi, I already applied the mandatory migration on nslord and nsmaster (added `cryptokeys.published`).

I wonder whether we should also apply the optional ones (changing `domains.notified_serial`, dropping `records.change_date`). It's optional, but maybe cleaner to do it. What do you think?

nils-wisiol:
Let's apply the optional changes, too

peterthomassen:
> Let's apply the optional changes, too

done

PR #380, opened 2020-05-12
First stab at GUI

peterthomassen:
GUI with
- Log in
- Token Management (add, delete, change name)
- Simple Domain Management (add, delete, display NS/DS/IPs)
- Change email
- Delete account

:tada:


webapp/src/views/CrudDomainList.vue
```diff
@@ -53,7 +53,7 @@ export default {
         actions: [
           {
             key: 'info',
-            go: d => (alert(d.name)),
+            go: d => this.postcreate(d),
```
nils-wisiol: change to `this.showDomainInfo(d)`

webapp/src/views/Console/DomainDetailsDialog.vue
```diff
@@ -105,18 +113,22 @@ export default {
       type: Array,
       required: true,
     },
+    ns: {
+      type: Array,
+      default: () => ['ns1.desec.io', 'ns2.desec.org'],
```
nils-wisiol, line 130: read these values from env

webapp/src/views/CrudDomainList.vue
```diff
@@ -76,7 +76,9 @@ export default {
               store.commit('working', false);
             }
           }
-          this.extraComponentBind = {'name': d.name, 'ds': d.keys.map(key => key.ds)};
+          let ds = d.keys.map(key => key.ds);
+          ds = ds.concat.apply([], ds)
```
nils-wisiol: what does this do?

webapp/src/views/DomainList.vue
```diff
@@ -23,6 +23,7 @@ export default {
           destroy: 'Domain Deletion',
         },
         texts: {
+          banner: () => ('You can create and delete domains here. We will soon extend our GUI to offer DNS record management. In the meantime, please use the API to manage records.'),
```
nils-wisiol, line 26: add link to the docs?

webapp/src/views/CrudList.vue
```diff
@@ -576,9 +575,6 @@ export default {
               typeof value === 'string' &&
               value.toString().toLocaleLowerCase().indexOf(search.toLocaleLowerCase()) !== -1
     },
-    getDatatype(datatype, item) {
```
nils-wisiol: this is important for editing of RRs

webapp/src/views/CrudList.vue
```diff
@@ -204,20 +204,11 @@
...
-              <div
-                v-else
-                class="py-5 text-xs-center"
-              >
+              <div class="py-4 text-xs-center">
                 <h2 class="title">
                   Feels so empty here!
                 </h2>
-                <p>No data yet.</p>
+                <p>No entries yet.</p>
```
nils-wisiol, line 211: This message is misleadingly displayed during loading

webapp/src/views/CrudList.vue
```diff
@@ -371,7 +362,6 @@ export default {
     defaultObject: {},
     // callbacks
     postcreate: () => (undefined),
-    rowclick: () => (undefined),
```
nils-wisiol: This was used to switch to a domain's RR set view

webapp/src/views/DeleteAccount.vue
```diff
@@ -0,0 +1,164 @@
...
+          this.done = true;
+        } catch (error) {
+          if (error.response) {
+            // status is not 2xx
+            if (error.response.status < 500 && typeof error.response.data === 'object') {
+              // 3xx or 4xx
+              this.errors = [error.response.data.detail];
+            } else {
+              // 5xx
+              this.errors = ['Something went wrong at the server, but we currently do not know why. The customer support was already notified.'];
```
nils-wisiol, line 146: customer support => support

webapp/src/views/CrudDomainList.vue
```diff
@@ -1,61 +1,86 @@
...
+        updatable: false,
+        destroyable: true,
+        limit_domains: 0,
+        headlines: {
+          table: 'Domains',
+          create: 'Create New Domain',
+          destroy: 'Domain Deletion',
+        },
+        texts: {
+          create: () => `You have ${self.availableCount} of ${self.limit_domains} domains left.`,
```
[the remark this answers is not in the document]
peterthomassen: `createWarning` appears nowhere else in the codebase, so I guess you're suggestion is missing something?

webapp/src/views/CrudList.vue
```diff
@@ -100,7 +100,10 @@
                       v-bind="c.fieldProps ? c.fieldProps(createDialogItem) : {}"
                       :label="c.textCreate || c.text"
                       :error-messages="c.createErrors"
+                      :required="c.required || false"
+                      :disabled="!available"
```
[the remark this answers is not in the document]
peterthomassen: I don't know, but that would be defined in `RRSetList.vue`; CrudList makes no assumption on what availability means.

webapp/src/views/CrudDomainList.vue
peterthomassen: already done in newer revision

webapp/src/views/CrudDomainList.vue
peterthomassen: it flattens the list. `d.keys` is the `keys` field specified [here](https://desec.readthedocs.io/en/latest/dns/domains.html#domain-field-reference), and the resulting `ds` is a list of strings from `d.keys[*]['ds']`, where the list index `*` is "integrated over"; it does not matter how may entries the `keys` field has.

webapp/src/views/DomainList.vue
peterthomassen: I don't think you can add HTML here

webapp/src/views/CrudList.vue
peterthomassen: but it's currently dead, and I ran into some issue when refactoring the datatable `item.<name>` slots. I decided to fix it when introducing RRset editing.

webapp/src/views/CrudDomainList.vue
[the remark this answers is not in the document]
nils-wisiol: I meant to suggest that this could be introduced, similar to `destroyWarning`

webapp/src/views/CrudList.vue
peterthomassen: I tested this many times, and though I had verified that it displays the standard v-datatable loading message ("fetching data" or something). Can you let me know how to reproduce the issue?

webapp/src/views/CrudList.vue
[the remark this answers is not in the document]
nils-wisiol: It defines the semantics, which is: above-mentioned element is disabled if it is "unavailable"

webapp/src/views/CrudList.vue
peterthomassen: yes, but it's currently dead, and besides the click handling was no longer compatible with the new `item.<name>` slots. We should instead be using the `click:row` event provided by v-datatable. I decided to come up with a new way of handling in consistently with the new slots once we have a use case.

webapp/src/views/DomainList.vue
nils-wisiol: I've seen it elsewhere, maybe it's worth a try

webapp/src/router/index.js
```diff
@@ -94,4 +134,44 @@ const router = new VueRouter({
...
+router.beforeEach((to, from, next) => {
+  // see if there are credentials in the session store that we don't know of
+  let recovered = false;
+  if (sessionStorage.getItem('token') && !store.state.authenticated){
+    HTTP.defaults.headers.common['Authorization'] = 'Token ' + sessionStorage.getItem('token');
+    store.commit('login', sessionStorage.getItem('token'));
+    recovered = true
+  }
+
+  if (to.matched.some(record => 'guest' in record.meta && record.meta.guest === false)) {
```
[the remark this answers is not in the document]
peterthomassen: It's a `RouteRecord` from Vue Router, and they call it `record` in their own examples: https://router.vuejs.org/guide/advanced/meta.html. In my terminology, the route would be `record.path`.

If you feel strongly, I'll change it.

webapp/src/views/TokenList.vue
```diff
@@ -0,0 +1,85 @@
...
+            searchable: false,
+          },
+        },
+        paths: {
+          list: 'auth/tokens/',
+          create: 'auth/tokens/',
+          delete: 'auth/tokens/:{id}/',
+          update: 'auth/tokens/:{id}/',
+        },
+        defaultObject: { name: '' },
```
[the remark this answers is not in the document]
peterthomassen: `defaultObject` is only used in `CrudList.vue` with `this.createDialogItem = Object.assign({}, this.defaultObject);` which copies all enumerable own properties from `defaultObject` into the new object, so there is no issue I think (but the function would introduce unneeded complexity).

I was thinking to add `Object.freeze(this.defaultObject)` to the `created()` lifecycle hook, but it turns out that properties still can be changed then ...

webapp/src/views/CrudDomainList.vue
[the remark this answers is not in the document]
peterthomassen: done!

webapp/src/views/CrudList.vue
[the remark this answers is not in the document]
peterthomassen: renamed to `disabled` with inverse meaning

webapp/src/views/CrudList.vue
```diff
@@ -295,6 +295,12 @@
               </v-form>
             </v-card>
           </v-dialog>
+          <component
+                  :is="extraComponentName"
+                  v-if="true || extraComponentName"
+                  v-bind="extraComponentBind"
+                  @input="() => { this.extraComponentName = ''; }"
+          ></component>
```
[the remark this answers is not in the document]
peterthomassen: that was a no-op bug, I removed the line

webapp/src/views/DomainList.vue
peterthomassen: worked by adding a `span`.

webapp/src/views/DeleteAccount.vue
peterthomassen: done

webapp/src/views/Console/DomainDetailsDialog.vue
peterthomassen: done

peterthomassen:
I took care of all your comments and added a commit with the corresponding fixes. Can you close the conversations you're ok with and comment on the other ones? thx!

webapp/src/router/index.js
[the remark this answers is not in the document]
nils-wisiol: nope, no strong feeling

webapp/src/views/CrudList.vue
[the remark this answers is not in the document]
nils-wisiol: I'm not sure if I am missing something, but in `CrudList.vue`, this value only determines if new entries in the table can be created, no? IMHO, both "available" and "disabled" suggest different semantics. `CrudList.vue` also implements a field called `createable`, maybe those two are related? (Sorry I'm a bit confused about this 😕)

webapp/src/views/DomainList.vue
nils-wisiol: nice :+1:

webapp/src/views/TokenList.vue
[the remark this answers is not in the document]
nils-wisiol: Alright, fine to me. One could argue that the function returning a new object would be more vuejs-ish, as they also do that in important places. No strong feeling though.

webapp/src/views/CrudList.vue
nils-wisiol: Indeed, it looks fine. Not sure anymore when/where I saw that.

nils-wisiol:
Two more unresolved conversations before this is ready to go!

webapp/src/views/TokenList.vue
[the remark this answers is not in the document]
peterthomassen: done! was less complicated than I thought somehow.

webapp/src/views/CrudList.vue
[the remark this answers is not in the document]
peterthomassen: done

peterthomassen:
I addressed the two outstanding conversations. Do you think it's good to go now?

In that case, I'll move unused DomainDetail code out of the way and squash, alirght?

nils-wisiol:
engage

PR #381, opened 2020-05-12
fix(api): do not allow underscores in zone names

nils-wisiol:


api/desecapi/models.py
```diff
@@ -194,8 +194,9 @@ def make_hash(plain):
 validate_domain_name = [
     validate_lower,
     RegexValidator(
-        regex=r'^([a-z0-9_-]{1,63}\.)*[a-z]{1,63}$',
-        message='Invalid value (not a DNS name).',
+        regex=r'^(([a-z0-9][a-z0-9-]{0,61}[a-z0-9]|[a-z0-9])\.)*[a-z]{1,63}$',
+        message='Invalid value: only use hyphens, digits, and letters; labels must not start with a hyphen and not '
+                'exceed 63 byte. TLDs may only have letters.',
```
peterthomassen, line 199: - labels apparently also can't end with a hyphen?
- `Invalid value: ` can go away (other validation error messages don't say that)
- `TLDs` --> `The last label`

nils-wisiol: > labels apparently also can't end with a hyphen?

The rationale was to only allow host names as zone names, plus names that start with digits to accommodate for reverse lookups.

I formatted the message along the lines of the type and subname messages and replaced "TLD" as suggested.

peterthomassen:
lgtm!

Issue #373, opened 2020-04-29
Support AXFR Import via API for Convenient Migration


nils-wisiol:
Maybe via a POST `api/v1/domains/example.com/axfr/` or something. dnspython may be able to do the parsing. Delete everything and replace? Also, ignore the SOA? What if DNSSEC records are present? (Suggestion: silently ignore)

peterthomassen:
introduced via 51f7a33afbe60bbdc2b8c847d4046046e69c2674

Issue #334, opened 2020-04-10
Lift 500 byte length restriction for RR.content


peterthomassen:
Context:
- introduced here https://github.com/desec-io/desec-stack/commit/591e15ec687f349daa0966f5ad7bc21a78639789
- due to https://github.com/PowerDNS/pdns/issues/8012
- users are noticing https://talk.desec.io/t/openpgpkey-record/28

renne:
The 500 byte limit hits DKIM-RRs, too.

My1:
https://github.com/PowerDNS/pdns/issues/8012 -> seems fixed

peterthomassen:
Will be fixed in pdns-auth 4.4.0, but it has not been released yet. We will deploy the fix once it is officially available.

PR #385, opened 2020-05-15
Move RR content validation away from nslord

nils-wisiol:


nils-wisiol:
In relation to #334, this imposes an additional 255 char limit "per field" which we may not want

nils-wisiol:
> In relation to #334, this imposes an additional 255 char limit "per field" which we may not want

Actually, this limit is per "token" and not per record, and was [apparently introduced by BIND.](https://kb.isc.org/docs/aa-00356) It seems to be [widely accepted](https://tools.ietf.org/html/rfc4408#section-3.1.3) although I could [not find any standard](https://tools.ietf.org/html/rfc1035#section-3.3.14) on it. 

With the current status of this PR, we would start to enforce this widely accepted non-standard limitation, because it is [enforced by dnspython](https://github.com/rthalley/dnspython/blob/d85b31d8be47bd7d8cfa67f7119ba833738a8038/dns/rdtypes/txtbase.py#L72). We could make an exception, but this may interfere with [future development](https://github.com/desec-io/desec-stack/issues/373) that we have planned.

I separated the tests for the "255 limit" per token and "500 limit" per record, and adapted the error message for the former appropriately.

andrewtj:
I can clarify the size limits. TXT records are composed of `character-string`s. A [`character-string` is up to 255 arbitrary bytes](https://tools.ietf.org/html/rfc1035#section-3.3), preceded by their length in a single byte.

In ASCII presentation format a `character-string` can be longer than 255 bytes as non printable ASCII bytes are presented in the [`\DDD` format](https://tools.ietf.org/html/rfc1035#section-5.1).

So for example a single zero byte `character-string` would be `"\000"` in ASCII presentation format and on the wire would be the two byte sequence `0x01` for the length and `0x00` for the content.

There is no size limit on the TXT record itself other than imposed by RDLENGTH (16-bits) and the restraints of being in a message (header, RR owner meta, etc).

andrewtj:
I should also mention that PowerDNS doesn't parse TXT records in an entirely standard way. It is [briefly mentioned here](https://doc.powerdns.com/authoritative/appendices/types.html#txt) but I can't find a precise reference to exactly what it does. Possibly its behaviour has lead to some of the confusion here.

peterthomassen:
> I can clarify the size limits. TXT records are composed of `character-string`s. A [`character-string` is up to 255 arbitrary bytes](https://tools.ietf.org/html/rfc1035#section-3.3), preceded by their length in a single byte.
...
> There is no size limit on the TXT record itself other than imposed by RDLENGTH (16-bits) and the restraints of being in a message (header, RR owner meta, etc).

Ah, I also remember that pdns permits TXT records to be longer than 255 bytes "on the admin side" (database, API etc.), and handles the 255-byte chunking as an aspect of client-side serialization.

I think that's a pretty reasonable decision, but it causes the above problem if we introduce a validation layer that works a) on the data that we want to send to the pdns API, but b) is based on "client-side style" serialization/deserialization (using dnspython).

@Habbie, did I summarize this correctly?

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+        """
+        rdtype = rdatatype.from_text(self.type)  # TODO may raise if type is unknown
+
+        def _error(record, msg):
+            raise ValidationError(f'Record content of {self.name}/{self.type} invalid: `{record}`: {msg}')
+
+        records_canonical_presentation_format = []
+        for r in records_presentation_format:
+            try:
+                records_canonical_presentation_format.append(RR.canonical_presentation_format(r, rdtype))
```
peterthomassen, line 466: you can start this as a set right away

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+        Raises ValidationError if violation of DNS specification is found.
+
+        Returns a set of records in canonical presentation format.
+
+        :param records_presentation_format: iterable of records in presentation format
+        """
+        rdtype = rdatatype.from_text(self.type)  # TODO may raise if type is unknown
+
+        def _error(record, msg):
+            raise ValidationError(f'Record content of {self.name}/{self.type} invalid: `{record}`: {msg}')
```
peterthomassen, line 461: Maybe collect errors in a dictionary or something, and in the end do `if any(errors): raise ValidationError(...)`? This way, all errors can be returned at the same time, as we do in other parts of the code

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+        Updates this RR set's resource records, discarding any old values.
+
+        Records are expected in presentation format and are converted to canonical
+        presentation format (e.g., 127.00.0.1 will be converted to 127.0.0.1).
+        Raises if a invalid set of records is provided.
+
+        This method triggers the following database queries:
+        - two SELECT queries for comparison of old with new records
+        - for each removed record, one DELETE query
+        - if at least record was added, a total of one INSERT query
```
peterthomassen, line 502: `if one or more records were added`

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+
+        Changes are saved to the database immediately.
+
+        :param records: list of records in representation format
```
peterthomassen, line 506: `presentation format`

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+        """
+        records = self.clean_records(records)
+
+        # Remove RRs that we didn't see in the new list
+        removed_rrs = self.records.exclude(content__in=records)  # one SELECT
```
peterthomassen, line 511: `self.records.exclude(content__in=records).delete()`?

Also, do we need locking of the RRset to avoid interference with concurrent requests?

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+        for rr in removed_rrs:
+            rr.delete()  # one DELETE query
+
+        # Figure out which entries in records have not changed
+        unchanged_rrs = self.records.filter(content__in=records)  # one SELECT
```
peterthomassen, line 516: This should be equivalent to `self.records` as all records that do not fulfill the filter conditions have just been deleted

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+        unchanged_content = [unchanged_rr.content for unchanged_rr in unchanged_rrs]
+        added_content = filter(lambda c: c not in unchanged_content, records)
+
+        rrs = [RR(rrset=self, content=content) for content in added_content]
```
peterthomassen, line 520:
```
    added_content = self.records.value_list('content', flat=True)
    RR.objects.bulk_create([
        RR(rrset=self, content=content)
        for content in added_content
    ])
```

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+        RR.objects.bulk_create(rrs)  # One INSERT
```
peterthomassen, line 521: Should we return anything reasonable here, e.g. number of deleted and added records? Both are available in a straightforward manner from `QuerySet.delete()` and `QuerySet.bulk_create()`.

api/desecapi/serializers.py
```diff
@@ -243,12 +244,15 @@ def get_validators(self):
...
-            raise serializers.ValidationError(f'The {value} RRset type is currently unsupported.')
-        if value in models.RRset.RESTRICTED_TYPES:
-            raise serializers.ValidationError(f'You cannot tinker with the {value} RRset.')
-        if value.startswith('TYPE'):
-            raise serializers.ValidationError('Generic type format is not supported.')
+        if value not in models.RR_SET_TYPES_MANAGEABLE:
+            # TODO see if SMIMEA can become a manageable type
+            # user cannot manage this type, let's try to tell her the reason
+            if value in models.RR_SET_TYPES_AUTOMATIC:
+                raise serializers.ValidationError(f'The {value} RRset(s) are managed by the API automatically.')
```
peterthomassen, line 251: `f'You cannot tinker with the {value} RRset. It is managed automatically.')`

api/desecapi/tests/base.py
```diff
@@ -984,16 +956,9 @@ def setUp(self):
...
-    # see https://doc.powerdns.com/md/types/
-    PDNS_RR_TYPES = ['A', 'AAAA', 'AFSDB', 'ALIAS', 'CAA', 'CERT', 'CDNSKEY', 'CDS', 'CNAME', 'DNSKEY', 'DNAME', 'DS',
-                     'HINFO', 'KEY', 'LOC', 'MX', 'NAPTR', 'NS', 'NSEC', 'NSEC3', 'NSEC3PARAM', 'OPENPGPKEY', 'PTR',
-                     'RP', 'RRSIG', 'SOA', 'SPF', 'SSHFP', 'SRV', 'TKEY', 'TSIG', 'TLSA', 'SMIMEA', 'TXT', 'URI']
-    ALLOWED_TYPES = ['A', 'AAAA', 'AFSDB', 'CAA', 'CERT', 'CDNSKEY', 'CDS', 'CNAME', 'DS', 'HINFO', 'KEY', 'LOC', 'MX',
-                     'NAPTR', 'NS', 'NSEC', 'NSEC3', 'OPENPGPKEY', 'PTR', 'RP', 'SPF', 'SSHFP', 'SRV', 'TKEY', 'TSIG',
-                     'TLSA', 'SMIMEA', 'TXT', 'URI']
+    UNSUPPORTED_TYPES = RR_SET_TYPES_UNSUPPORTED
+    AUTOMATIC_TYPES = RR_SET_TYPES_AUTOMATIC
+    ALLOWED_TYPES = RR_SET_TYPES_MANAGEABLE
```
peterthomassen, line 961: Shouldn't some be hardcoded here in the tests? If for some reason these sets are empty, the tests would not notice, right?

api/desecapi/tests/test_domains.py
```diff
@@ -296,12 +296,6 @@ def test_create_domain_other_parent(self):
         self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(response.data['name'][0].code, 'name_unavailable')
 
-    def test_create_domain_atomicity(self):
-        name = self.random_domain_name()
-        with self.assertPdnsRequests(self.request_pdns_zone_create_422()):
-            self.client.post(self.reverse('v1:domain-list'), {'name': name})
-            self.assertFalse(Domain.objects.filter(name=name).exists())
-
```
peterthomassen, line 304: why remove this?

api/desecapi/models.py
```diff
@@ -418,6 +444,83 @@ def save(self, *args, **kwargs):
...
+                records_canonical_presentation_format.append(RR.canonical_presentation_format(r, rdtype))
+            except binascii.Error:
+                # e.g., odd-length string
+                _error(r, 'Cannot parse hexadecimal or base64 record contents')
+            except dns.exception.SyntaxError as e:
+                # e.g., A/127.0.0.999
+                if e.args[0] == 'string too long':
+                    excess = len(r) - 255
+                    _error(r, f'Record content {excess} chars too long.')
+                _error(r, 'Record syntax malformed')
```
peterthomassen, line 474: `Record content malformed`

api/desecapi/tests/test_rrsets.py
```diff
@@ -199,16 +199,16 @@ def test_create_other_rr_sets(self):
 
     def test_create_my_rr_sets_too_long_content(self):
         def _create_data(length):
-            content_string = 'A' * (length - 2)  # we have two quotes
-            return {'records': [f'"{content_string}"'], 'ttl': 3600, 'type': 'TXT', 'subname': f'name{length}'}
+            content_string = 'A' * length
+            return {'records': [f'{content_string}'], 'ttl': 3600, 'type': 'TXT', 'subname': f'name{length}'}
```
peterthomassen: Why remove the quotes? They are mandatory in `TXT`

api/desecapi/tests/test_rrsets.py
```diff
@@ -199,16 +199,16 @@ def test_create_other_rr_sets(self):
...
 
         with self.assertPdnsRequests(self.requests_desec_rr_sets_update(self.my_empty_domain.name)):
-            response = self.client.post_rr_set(self.my_empty_domain.name, **_create_data(500))
+            response = self.client.post_rr_set(self.my_empty_domain.name, **_create_data(255))
```
peterthomassen: We should allow longer record contents and somehow work around the chunking issue

api/desecapi/tests/test_rrsets.py
```diff
@@ -228,6 +228,15 @@ def test_create_my_rr_sets_twice(self):
...
         self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
 
+    def test_create_my_rr_sets_duplicate_content(self):
+        for records in [
+            ['127.0.0.1', '127.00.0.1'],
+            # TODO add more examples
+        ]:
+            data = {'records': records, 'ttl': 3660, 'type': 'A'}
+            response = self.client.post_rr_set(self.my_empty_domain.name, **data)
+            self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
```
peterthomassen, line 269: assert error message text `f'Duplicate records in RR set .../...'`

api/desecapi/tests/test_rrsets.py
```diff
@@ -239,14 +248,155 @@ def test_create_my_rr_sets_empty_payload(self):
...
+            ('OPENPGPKEY', ('', '')),
+            ('PTR', ('', '')),
+            ('RP', ('', '')),
+            ('SPF', ('', '')),
+            ('SRV', ('', '')),
+            ('SSHFP', ('', '')),
+            ('TKEY', ('', '')),
+            ('TLSA', ('3 0001 1 000AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA', '3 1 1 000aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')),
+            ('TSIG', ('', '')),
+            ('TXT', ('"foobar"', 'foobar')),
```
peterthomassen, line 322: I'm pretty sure the canonical presentation format does have quotes

api/desecapi/tests/test_rrsets.py
```diff
@@ -239,14 +248,155 @@ def test_create_my_rr_sets_empty_payload(self):
...
 
-    def test_create_my_rr_sets_unknown_type(self):
-        for _type in ['AA', 'ASDF']:
-            with self.assertPdnsRequests(
-                    self.request_pdns_zone_update_unknown_type(name=self.my_domain.name, unknown_types=_type)
-            ):
-                response = self.client.post_rr_set(self.my_domain.name, records=['1234'], ttl=3660, type=_type)
+    def test_create_my_rr_sets_canonical_content(self):
+        # TODO fill in more examples
+        datas = [
```
peterthomassen, line 284: should be a dict

api/desecapi/tests/test_rrsets.py
```diff
@@ -239,14 +248,155 @@ def test_create_my_rr_sets_empty_payload(self):
...
+            ('SRV', ('', '')),
+            ('SSHFP', ('', '')),
+            ('TKEY', ('', '')),
+            ('TLSA', ('3 0001 1 000AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA', '3 1 1 000aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')),
+            ('TSIG', ('', '')),
+            ('TXT', ('"foobar"', 'foobar')),
+            ('URI', ('', '')),
+            ('WKS', ('', '')),
+        ]
+        for t, (record, canonical_record) in datas:
```
peterthomassen, line 326: nice unpacking! Did not know it works like that

api/desecapi/tests/test_rrsets.py
```diff
@@ -239,14 +248,155 @@ def test_create_my_rr_sets_empty_payload(self):
...
+            if not record:
+                continue
+            data = {'records': [record], 'ttl': 3660, 'type': t, 'subname': ''}
+            with self.assertPdnsRequests(self.requests_desec_rr_sets_update(name=self.my_empty_domain.name)):
+                response = self.client.post_rr_set(self.my_empty_domain.name, **data)
+                self.assertContains(response, canonical_record, status_code=status.HTTP_201_CREATED)
+            with self.assertPdnsRequests(self.requests_desec_rr_sets_update(name=self.my_empty_domain.name)):
+                response = self.client.delete_rr_set(self.my_empty_domain.name, subname='', type_=t)
```
peterthomassen, line 334: `type_` could be made `type` (without shadowing anything) if the function definition is changed from

```
    def delete_rr_set(self, domain_name, subname, type_):
        return self.delete(
            self.reverse('v1:rrset@', name=domain_name, subname=subname, type=type_)
        )
```

to

```
    def delete_rr_set(self, domain_name, *args, **kwargs):
        return self.delete(
            self.reverse('v1:rrset@', name=domain_name, *args, **kwargs)
        )
```

But I recognize this has nothing to do with this PR.

api/desecapi/tests/test_rrsets.py
```diff
@@ -239,14 +248,155 @@ def test_create_my_rr_sets_empty_payload(self):
...
+            'DHCID': [],
+            'DLV': [],
+            'DS': ['-34 13 1 aabbccddeeff'],
+            'EUI48': [],
+            'EUI64': [],
+            'HINFO': ['"ARMv8-A"'],
+            'IPSECKEY': [], 'KEY': [], 'KX': [],
+            'LOC': ['23 12 61.000 N 42 22 48.500 W 65.00m 20.00m 10.00m 10.00m', 'foo', '1.1.1.1'],
+            'MAILA': [], 'MAILB': [], 'MINFO': [], 'MR': [],
+            'MX': ['10 example.com', 'example.com.', '-5 asdf.', '65537 asdf.'],
```
peterthomassen, line 405: try adding IP as a hostname

api/desecapi/tests/test_rrsets.py
```diff
@@ -239,14 +248,155 @@ def test_create_my_rr_sets_empty_payload(self):
...
+        for t, records in datas.items():
+            for r in records:
+                data = {'records': [r], 'ttl': 3660, 'type': t, 'subname': ''}
+                response = self.client.post_rr_set(self.my_empty_domain.name, **data)
                 self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
 
+    def test_create_my_rr_sets_unknown_type(self):
+        for _type in ['AA', 'ASDF'] + list(RR_SET_TYPES_AUTOMATIC | RR_SET_TYPES_UNSUPPORTED):
+            response = self.client.post_rr_set(self.my_domain.name, records=['1234'], ttl=3660, type=_type)
+            self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
```
peterthomassen, line 429: let's check a bit more specifically that the type was rejected (and not the record content or something else)

api/desecapi/tests/test_rrsets.py
```diff
@@ -197,18 +197,49 @@ def test_create_other_rr_sets(self):
...
+            **_create_data('"' + 'A' * 256 + '"')
+        )
+        self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
+        self.assertIn('Splitting the data in chunks is required (RFC 4408).', str(response.data))
+
     def test_create_my_rr_sets_too_long_content(self):
-        def _create_data(length):
-            content_string = 'A' * length
-            return {'records': [f'{content_string}'], 'ttl': 3600, 'type': 'TXT', 'subname': f'name{length}'}
+        def _create_data(record):
```
peterthomassen, line 223: code duplication from a few lines up

api/desecapi/tests/test_rrsets.py
```diff
@@ -197,18 +197,49 @@ def test_create_other_rr_sets(self):
...
+                **_create_data('"' + 'A' * 255 + '" "' + 'A' * 240 + '"')
+            )
             self.assertStatus(response, status.HTTP_201_CREATED)
 
-        response = self.client.post_rr_set(self.my_empty_domain.name, **_create_data(256))
+        response = self.client.post_rr_set(
+            self.my_empty_domain.name,
+            # record of length 501 bytes in chunks of max 255 each (RFC 4408)
+            # "AAA...AAA" "AAA...AAA"
+            **_create_data('"' + 'A' * 255 + '" "' + 'A' * 241 + '"')
```
peterthomassen, line 239: Is this all there is to the tokenization? If so, let's our API handle it automatically.

Also, `**_create_data(f'"{"A"*255}" "{"A"*241}"')`

api/desecapi/models.py
```diff
@@ -358,6 +359,34 @@ class Meta:
...
+    'CDNSKEY', 'CDS', 'DNSKEY', 'NSEC', 'NSEC3', 'OPT', 'RRSIG',
+    # automatically managed by the API:
+    'NSEC3PARAM', 'SOA'
+}
+# backend types are types that are the types supported by the backend(s)
+RR_SET_TYPES_BACKEND = pdns.SUPPORTED_RRSET_TYPES
+# validation types are types supported by the validation backend, currently: dnspython
+RR_SET_TYPES_VALIDATION = {
+    # from dns import rdatatype
+    # sorted(rdatatype._by_text.keys())
```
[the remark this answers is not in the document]
nils-wisiol: :shrug: wasn't sure what's worse: hardcoding or accessing private members

api/desecapi/pdns.py
```diff
@@ -1,14 +1,24 @@
...
+from desecapi.exceptions import PDNSException, RequestEntityTooLarge
 
+SUPPORTED_RRSET_TYPES = {
+    # https://doc.powerdns.com/authoritative/appendices/types.html
+    # "major" types
+    'A', 'AAAA', 'AFSDB', 'ALIAS', 'CAA', 'CERT', 'CDNSKEY', 'CDS', 'CNAME', 'DNSKEY', 'DNAME', 'DS', 'HINFO', 'KEY',
+    'LOC', 'MX', 'NAPTR', 'NS', 'NSEC', 'NSEC3', 'NSEC3PARAM', 'OPENPGPKEY', 'PTR', 'RP', 'RRSIG', 'SOA', 'SPF',
+    'SSHFP', 'SRV', 'TKEY', 'TSIG', 'TLSA', 'SMIMEA', 'TXT', 'URI',
+
+    # "additional" types, without obsolete ones
```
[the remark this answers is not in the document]
nils-wisiol: I guessed they would be obsolete for a reason, and usage is discouraged and/or should be phased out. Have to admin, though, that I did not look into the details of what those record types are for.

api/desecapi/models.py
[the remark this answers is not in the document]
nils-wisiol: replaced by `'`

api/desecapi/models.py
nils-wisiol: done

api/desecapi/models.py
nils-wisiol: will do

api/desecapi/tests/test_rrsets.py
nils-wisiol: no, dicts do not allow duplicate keys. (Assuming you meant that the record type is the key.)

api/desecapi/tests/test_rrsets.py
nils-wisiol: done

api/desecapi/models.py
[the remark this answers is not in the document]
peterthomassen: Let's access the private member; we'll at least notice then when it breaks (while we will not notice when the hardcoded list should change)

api/desecapi/pdns.py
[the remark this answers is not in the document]
peterthomassen: If dnspython supports them, do we have to care? :)

Again, they are not deprecated. I think it's like SPF, which today is practically superseded by TXT.

api/desecapi/tests/test_rrsets.py
nils-wisiol: done

api/desecapi/tests/test_rrsets.py
peterthomassen: Well, you don't have duplicates, so I didn't expect that that should be supported :)

Fine with me. -- Anyways, [this `defaultdict` idea](https://stackoverflow.com/a/6813689/6867099) may be instructive (not implying that it should be used here).

api/desecapi/models.py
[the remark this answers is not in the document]
peterthomassen: I would do: `f'Record content of {self.name}/{self.type} invalid: {msg}. Record: {record}'`

This way, you get rid of any surrounding delimiter, which especially in case of TXT records may be confusing to the user ("why does it show `'`? I thought I passed `"`!" etc.) However, I don't feel strongly.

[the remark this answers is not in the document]
nils-wisiol: Leading and trailing whitespace may matter in the validation, which is why I included the delimiters. The TXT record would show `'"my record"'`, so that should be comprehensible.

api/desecapi/tests/test_rrsets.py
nils-wisiol: I prefer multiple lines with the same record type over one line where either the elements in the list have alternating meaning (input, canonical expectation, ...) or is a nested list, like suggested in `defaultdict`.

api/desecapi/pdns.py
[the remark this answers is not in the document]
nils-wisiol: This only affects `A6` and `SIG` records, `A6` is defined by RFC 2874, which has status "historic". `SIG` records are superseded by `RRSIG`. `KEY`, btw, is listed both as "major" and "obsolete" record type [in the pdns docs](# https://doc.powerdns.com/authoritative/appendices/types.html).

Care to support `A6` and/or `SIG`?

api/desecapi/tests/test_rrsets.py
nils-wisiol: There's a whole bunch of functions with similar interface, changing all of them isn't worth the small benefit, just changing this one would make the code style worse. So, I'll opt for being lazy

api/desecapi/tests/test_rrsets.py
nils-wisiol: Please elaborate :) Not sure how/what the API should be doing precisely. If, e.g., a token exceeds the max. length, then what do we do?

api/desecapi/tests/test_rrsets.py
nils-wisiol: fixed

nils-wisiol:
resolved a couple of annotations, added a couple of tests. Of particular interest are the TXT canonicalization tests: we newly accept now `foo bar`, which means `"foo" "bar"`. The semantics of anything in quotation marks didn't change.

The increase of the max token length is still missing.

api/desecapi/models.py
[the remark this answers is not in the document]
nils-wisiol: fair point. done

nils-wisiol:
I (re)introduced non-standard TXT/SPF records with token length up to 256**8-1. The validation is now still based upon dnspython, but involves quite some copy-paste code. Not sure how we want to proceed with this, some options:

- only offer standard records, at the cost of the user's convenience
- don't care about the copy-paste code, as it is likely to last (we wouldn't expect changes to the TXT record format)
- a smarter solution (you indicated in code annotations that you may have an idea, @peterthomassen ?)

Instead of a database migration, I opted to canonicalize record content every time we talk to pdns. We can apply a migration anytime (not saying it shouldn't be now) and then remove this piece of code again.

Some tests for more specific record types are still missing.

nils-wisiol:
Anyways, this extends the TXT record interface to also accept not-quoted record contents. For interpretation details, see the tests. (It follows RFC 1035 with the exception of longer tokens.)

nils-wisiol:
This needs some more e2e testing to make sure pdns accepts our canonicalized input.

api/desecapi/pdns.py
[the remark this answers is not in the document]
peterthomassen: > This only affects `A6` and `SIG` records

How do you mean? `MAILA`, for example, is also labeled obsolete in the pdns docs, while you list it as major.

> `KEY`, btw, is listed both as "major" and "obsolete" record type [in the pdns docs](https://doc.powerdns.com/authoritative/appendices/types.html).

😂

> Care to support `A6` and/or `SIG`?

Nope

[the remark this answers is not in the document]
nils-wisiol: No, by obsolete I meant record types listed in the [pdns docs](https://doc.powerdns.com/authoritative/appendices/types.html) in the section "other types" **and** explicitly labelled "obsolete". I.e., only A6, KEY, SIG, where KEY is also listed as as supported type in the list above.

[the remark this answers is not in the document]
peterthomassen: ah, my bad. So, with regards to my initial question "why not support obsolete ones?", the question is if we want to support `A6` and `SIG`. I'm not sure if there are other uses of `SIG` now which don't correspond to the meaning of `RRSIG`; on the other hand, I don't think that makes it necessary to support it. Either way is fine with me!

api/desecapi/models.py
```diff
@@ -373,18 +374,23 @@ class Meta:
...
-    'CDNSKEY', 'CDS', 'DNSKEY', 'NSEC', 'NSEC3', 'OPT', 'RRSIG',
+    'CDNSKEY', 'CDS', 'DNSKEY', 'KEY', 'NSEC', 'NSEC3', 'OPT', 'RRSIG',
     # automatically managed by the API:
     'NSEC3PARAM', 'SOA'
 }
 # backend types are types that are the types supported by the backend(s)
 RR_SET_TYPES_BACKEND = pdns.SUPPORTED_RRSET_TYPES
 # validation types are types supported by the validation backend, currently: dnspython
-RR_SET_TYPES_VALIDATION = set(rdatatype._by_text.keys())
+RR_SET_TYPES_VALIDATION = set(ANY.__all__) | set(IN.__all__)
```
peterthomassen, line 393: for clarity, better import `rdtypes` and write `rdtypes.ANY` and `rdtypes.IN` here

api/desecapi/tests/test_rrsets.py
```diff
@@ -339,6 +335,7 @@ def test_create_my_rr_sets_canonical_content(self):
             with self.assertPdnsRequests(self.requests_desec_rr_sets_update(name=self.my_empty_domain.name)):
                 response = self.client.delete_rr_set(self.my_empty_domain.name, subname='', type_=t)
                 self.assertStatus(response, status.HTTP_204_NO_CONTENT)
+        self.assertAllSupportedRRSetTypes(set(t for t, _ in datas))
```
peterthomassen, line 338: nice!

api/desecapi/tests/base.py
```diff
@@ -634,6 +634,9 @@ class DesecTestCase(MockPDNSTestCase):
     AUTO_DELEGATION_DOMAINS = settings.LOCAL_PUBLIC_SUFFIXES
     PUBLIC_SUFFIXES = {'de', 'com', 'io', 'gov.cd', 'edu.ec', 'xxx', 'pinb.gov.pl', 'valer.ostfold.no',
                        'kota.aichi.jp', 's3.amazonaws.com', 'wildcard.ck'}
+    SUPPORTED_RR_SET_TYPES = {'A', 'AAAA', 'AFSDB', 'CAA', 'CERT', 'CNAME', 'DHCID', 'DLV', 'DS', 'EUI48', 'EUI64',
+                              'HINFO', 'KX', 'LOC', 'MX', 'NAPTR', 'NS', 'OPENPGPKEY', 'PTR', 'RP', 'SPF', 'SRV',
+                              'SSHFP', 'TLSA', 'TXT', 'URI', 'WKS'}
```
peterthomassen, line 639: shall we somehow test this against what's considered supported in models.py?

api/desecapi/models.py
nils-wisiol: yeah I wish!! But:

```python3
>>> from dns import rdtypes
>>> rdtypes.ANY
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'dns.rdtypes' has no attribute 'ANY'
```

No idea what's going on here. But I decided it's not so important. If you can clarify my confusion here, that would be highly appreciated :)

nils-wisiol: I guess ANY is just not a member of `rdtypes`, but rather a submodule (that does not import ANY itself)

api/desecapi/models.py
nils-wisiol: done

api/desecapi/models.py
nils-wisiol: done

api/desecapi/models.py
nils-wisiol: done, this deserved a separate commit, as it touches the heart of DNS operations. There is no locking, operations follow the [optimistic locking](https://en.wikipedia.org/wiki/Optimistic_concurrency_control) paradigm.

api/desecapi/models.py
nils-wisiol: Changed accordingly. Curiously, we can't use `values_list` here: tests will fail due to `RR`s with identical content. Not sure why that happens -- I'm suspecting some part of `self.records` is cached somewhere

api/desecapi/models.py
nils-wisiol: No: imagine database holds records contents A and B, and records is B, C, then your proposal would have `added_content` containing only B, and C is missing.

nils-wisiol: Additionally, there is some problem with using `values_list` here, see comment above.

api/desecapi/models.py
nils-wisiol: This method is currently only used in (effectively) `RRsetSerializer.update` and `RRsetSerializer.create`, both return the `RRset` in question. I don't see how these methods could use those values.

api/desecapi/serializers.py
nils-wisiol: No other error message addresses the user personally

api/desecapi/tests/base.py
nils-wisiol: I added a separate test for that

api/desecapi/tests/test_domains.py
nils-wisiol: Because pdns errors are not exclusively treated as server errors, i.e. Django's DB transaction will be rolled back, so it seemed silly to keep it. But yeah, why not: I re-added an updated version of it.

api/desecapi/models.py
nils-wisiol: done

api/desecapi/tests/test_rrsets.py
nils-wisiol: While not strictly necessary, we agreed offline to deviate from the standard and make them mandatory

api/desecapi/tests/test_rrsets.py
nils-wisiol: done

api/desecapi/tests/test_rrsets.py
nils-wisiol: The notation in this test is misleading: the [canonical form of the record content](https://tools.ietf.org/html/rfc4034#section-6.2) is in __wire__ format, i.e. no quotes. What is compared here is the presentation format reconstructed by dnspython from canonical wire format. AFAIK, there is no definition of canonical presentation format.

The canonical form of `"foobar"` and `foobar` TXT records is `b'\x06foobar'`. [dnspython uses quotes whenever constructing TXT records from wire data](https://github.com/rthalley/dnspython/blob/5a3df82e3f93c4f480f96f78de8678504282a3e0/dns/rdtypes/txtbase.py#L58), so in above code snipped, there should have been quotes.

api/desecapi/tests/base.py
nils-wisiol: Such tests exist, e.g. in `AuthenticatedRRSetTestCase.test_create_my_rr_sets_type_restriction`

nils-wisiol:
I cleaned up the history. What's still missing:

- add a database migration that canonicalizes all database record contents
- determine what's our canonical TXT and SPF presentation format: chunked or not
  - if not chunked, determine what we send to pdns (chunked or not)
- unresolved conversations from above

api/desecapi/serializers.py
peterthomassen: So? :-)

nils-wisiol:
Also missing may be a check to determine if a CNAME is added to a zone where A/AAAA/TXT records already exist

NewRedsquare:
I don't understand ; this PR will set a new restriction or will fix bug to allow records longer than 500 chars ?

peterthomassen:
Neither. It is an internal change in our validation procedure that should be invisible from the client side (except for some edge cases).

NewRedsquare:
I thought it could solve the issue with OPENPGPKEY records... but it's an problem of PowerDNS not your's but they're incredibly slow to solve it :/

Thank you for this fast answer btw

api/desecapi/models.py
```diff
@@ -474,6 +501,90 @@ def save(self, *args, **kwargs):
...
+        rdtype = rdatatype.from_text(self.type)
+        errors = []
+
+        def _error_msg(record, detail):
+            return f'Record content of {self.type} {self.name} invalid: \'{record}\': {detail}'
+
+        records_canonical_format = set()
+        for r in records_presentation_format:
+            try:
+                r_canonical_format = RR.canonical_presentation_format(r, rdatatype.to_text(rdtype))
```
peterthomassen, line 523: Why call `rdatatype.to_text` and then the inverse inside `RR.canonical_presentation_format`?

api/desecapi/serializers.py
```diff
@@ -245,12 +246,16 @@ def get_validators(self):
...
     @staticmethod
     def validate_type(value):
-        if value in models.RRset.DEAD_TYPES:
-            raise serializers.ValidationError(f'The {value} RRset type is currently unsupported.')
-        if value in models.RRset.RESTRICTED_TYPES:
-            raise serializers.ValidationError(f'You cannot tinker with the {value} RRset.')
-        if value.startswith('TYPE'):
-            raise serializers.ValidationError('Generic type format is not supported.')
+        if value not in models.RR_SET_TYPES_MANAGEABLE:
+            # TODO see if SMIMEA can become a manageable type
```
peterthomassen, line 250: move comment to `models.RR_SET_TYPES_MANAGEABLE` declaration? -- but there, it is listed as supported?

api/desecapi/tests/base.py
```diff
@@ -665,6 +634,9 @@ class DesecTestCase(MockPDNSTestCase):
     AUTO_DELEGATION_DOMAINS = settings.LOCAL_PUBLIC_SUFFIXES
     PUBLIC_SUFFIXES = {'de', 'com', 'io', 'gov.cd', 'edu.ec', 'xxx', 'pinb.gov.pl', 'valer.ostfold.no',
                        'kota.aichi.jp', 's3.amazonaws.com', 'wildcard.ck'}
+    SUPPORTED_RR_SET_TYPES = {'A', 'AAAA', 'AFSDB', 'CAA', 'CERT', 'CNAME', 'DHCID', 'DLV', 'DS', 'EUI48', 'EUI64',
+                              'HINFO', 'KX', 'LOC', 'MX', 'NAPTR', 'NS', 'PTR', 'RP', 'SPF', 'SRV',
+                              'SSHFP', 'TLSA', 'TXT', 'URI'}
```
peterthomassen, line 639: `SMIMEA`?

docs/dns/rrsets.rst
```diff
@@ -87,6 +87,12 @@ Field details:
     The maximum number of array elements is 4091, and the maximum length of
     the array is 64,000 (after JSON encoding).
 
+    Records must be given in presentation format (a.k.a. "BIND" or zone file
+    format). Record values that are not given in canonical format, such as
+    ``127.000.0.1`` will be converted by the API into canonical format, e.g.
+    ``127.0.0.1``. Exact validation rules on for the records depend on the
```
peterthomassen, line 93: `on form`

docs/dns/rrsets.rst
```diff
@@ -87,6 +87,12 @@ Field details:
     The maximum number of array elements is 4091, and the maximum length of
     the array is 64,000 (after JSON encoding).
 
+    Records must be given in presentation format (a.k.a. "BIND" or zone file
+    format). Record values that are not given in canonical format, such as
```
peterthomassen, line 91: `canonical format` --> `canonical form`, see https://tools.ietf.org/html/rfc4034#section-6

test/e2e2/conftest.py
```diff
@@ -27,30 +32,87 @@ def random_email() -> str:
...
+            + ".test"
+        )
+
+    @staticmethod
+    def random_local_public_suffix_domainname():
+        return (
+            "".join(random.choice(string.ascii_lowercase) for _ in range(16))
+            + ".dedyn."
+            + os.environ['DESECSTACK_DOMAIN']
+        )
```
peterthomassen, line 48: Why are these methods of the client (whose instances are usually labeled `api_user`, so you get `api_user.random_domainname()`)?

test/e2e2/spec/test_api_rr_validation.py
```diff
@@ -0,0 +1,176 @@
...
+    ],
+    'SRV': ['0 0 0 .', '100 1 5061 example.com.'],
+    'SSHFP': ['2 2 aabbcceeddff'],
+    'TLSA': ['3 1 1 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',],
+    'TXT': [
+        '"foobar"',
+        '"foo" "bar"',
+        '"foo" "" "bar"',
+        '"" "" "foo" "" "bar"',
+        '"new\\010line"',
```
peterthomassen, line 66: try a NUL byte?

test/e2e2/spec/test_api_rr_validation.py
```diff
@@ -0,0 +1,176 @@
...
+        '"foo\nbar"',
+        '"' + 124 * '🧥' + '==="',  # 501 byte total length
+    ],
+    'URI': ['"1" "2" "3"'],
+}
+INVALID_RECORDS_PARAMS = [(rr_type, value) for rr_type in INVALID_RECORDS.keys() for value in INVALID_RECORDS[rr_type]]
+
+
+def test_soundness():
+    assert INVALID_RECORDS.keys() == VALID_RECORDS_CANONICAL.keys() == VALID_RECORDS_NON_CANONICAL.keys()
```
peterthomassen, line 160: 💛

peterthomassen (approved):
Great stuff! Lgtm, very minor comments

api/desecapi/tests/test_rrsets.py
```diff
@@ -364,9 +364,9 @@ def test_create_my_rr_sets_known_type_benign(self):
             'MX': ['10 example.com.', '20 1.1.1.1.'],
             'NAPTR': ['100  50  "s"  "z3950+I2L+I2C"     ""  _z3950._tcp.gatech.edu.'],
             'NS': ['ns1.example.com.'],
-            # 'OPENPGPKEY': [
-            #     'mG8EXtVIsRMFK4EEACIDAwQSZPNqE4tSxLFJYhX+uabSgMrhOqUizJhkLx82',  # key incomplete
-            # ],
+            'OPENPGPKEY': [
+                 'mG8EXtVIsRMFK4EEACIDAwQSZPNqE4tSxLFJYhX+uabSgMrhOqUizJhkLx82',  # key incomplete
```
peterthomassen, line 368: what does the comment in the end mean?

api/desecapi/models.py
```diff
@@ -426,7 +426,6 @@ class Meta:
     'DNAME',  # "do not combine with DNSSEC", https://doc.powerdns.com/authoritative/settings.html#dname-processing
     'IPSECKEY',  # broken in pdns, https://github.com/PowerDNS/pdns/issues/9055 TODO enable support
     'KEY',  # Application use restricted by RFC 3445, DNSSEC use replaced by DNSKEY and handled automatically
-    'OPENPGPKEY',  # large records not supported https://github.com/desec-io/desec-slave/issues/34 TODO enable support
```
peterthomassen: also needs to be added to `SUPPORTED_RR_SET_TYPES` in `api/desecapi/tests/base.py`?

api/desecapi/tests/base.py
peterthomassen: See also #436

api/desecapi/dns.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+
+    @classmethod
+    def from_text(cls, rdclass, rdtype, tok, origin=None, relativize=True):
+        strings = []
+        while 1:
+            token = tok.get().unescape()
+            if token.is_eol_or_eof():
+                break
+            if not token.is_quoted_string():
```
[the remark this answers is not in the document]
nils-wisiol: Nope, we just don't allow "bare" strings, i.e. unquoted content, to avoid confusion.

api/desecapi/models.py
```diff
@@ -415,6 +419,32 @@ class Meta:
...
+    'DNAME',  # "do not combine with DNSSEC", https://doc.powerdns.com/authoritative/settings.html#dname-processing
+    'IPSECKEY',  # broken in pdns, https://github.com/PowerDNS/pdns/issues/9055 TODO enable support
+    'KEY',  # Application use restricted by RFC 3445, DNSSEC use replaced by DNSKEY and handled automatically
+    'OPENPGPKEY',  # large records not supported https://github.com/desec-io/desec-slave/issues/34 TODO enable support
+    'WKS',  # General usage not recommended, "SHOULD NOT" be used in SMTP (RFC 1123)
+}
+# restricted types are managed in use by the API, and cannot directly be modified by the API client
+RR_SET_TYPES_AUTOMATIC = {
+    # automatically managed by the backend:
+    'CDNSKEY', 'CDS', 'DNSKEY', 'KEY', 'NSEC', 'NSEC3', 'OPT', 'RRSIG',
```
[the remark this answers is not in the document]
nils-wisiol: This changes the error message from "unsupported" to "automatically managed" record type. Admittedly, this never happens, but my idea was that if someone wanted to add a "KEY" to secure their records with some signatures, then we would tell them that we do that, instead of saying it's unsupported.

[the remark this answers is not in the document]
peterthomassen: ah, cool. I think it will be clear if the comment is `# corresponding functionality is automatically managed ...`

api/desecapi/models.py
```diff
@@ -474,6 +501,90 @@ def save(self, *args, **kwargs):
...
+            try:
+                r_canonical_format = RR.canonical_presentation_format(r, rdatatype.to_text(rdtype))
+                if r_canonical_format in records_canonical_format:
+                    errors.append(_error_msg(r, f'Duplicate record content: this is identical to '
+                                                f'\'{r_canonical_format}\''))
+                else:
+                    records_canonical_format.add(r_canonical_format)
+            except binascii.Error:
+                # e.g., odd-length string
+                errors.append(_error_msg(r, 'Cannot parse hexadecimal or base64 record contents'))
```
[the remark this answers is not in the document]
nils-wisiol: It's accurate: dnspython only uses binascii's `hexlify` and `unhexlify`, which directly raise binascii.Error. For base64 content (OPENPGPKEY record type), binascii.Error is raised by the base64 module if [the string is incorrectly padded](https://docs.python.org/3/library/base64.html#base64.b64decode); if non-ASCII chars are present, ValueError is raised.

test/e2e2/conftest.py
nils-wisiol: converted to fixtures

api/desecapi/tests/test_rrsets.py
nils-wisiol: it's not a valid key, but a truncated one (couldn't find one that fits the lmdb 500 byte limit)

nils-wisiol:
ready for second round, everything after the force push is new and addressing review comments

test/e2e2/conftest.py
```diff
@@ -13,6 +13,36 @@
...
+
+@pytest.fixture()
+def random_local_public_suffix_domainname() -> Callable[[], str]:
+    return lambda: (
+        "".join(random.choice(string.ascii_lowercase) for _ in range(16))
+        + ".dedyn."
+        + os.environ['DESECSTACK_DOMAIN']
+    )
+
+
```
peterthomassen: dead code

nils-wisiol: not for long I suppose :)

api/desecapi/models.py
```diff
@@ -474,6 +501,90 @@ def save(self, *args, **kwargs):
...
+                else:
+                    errors.append(_error_msg(r, f'Record content malformed: {",".join(e.args)}'))
+            except dns.name.NeedAbsoluteNameOrOrigin:
+                errors.append(_error_msg(r, 'Hostname must be fully qualified (i.e., end in a dot: "example.com.")'))
+            except ValueError:
+                # e.g., string ("asdf") cannot be parsed into int on base 10
+                errors.append(_error_msg(r, 'Cannot parse record contents'))
+            except AssertionError:
+                # e.g., HINFO token too long, cf. https://github.com/rthalley/dnspython/issues/493
+                errors.append(_error_msg(r, 'Invalid record content'))
```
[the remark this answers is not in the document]
nils-wisiol: done

api/desecapi/serializers.py
nils-wisiol: deleted without replacement; if dnspython starts supporting SMIME so do we -- tests are already prepared

nils-wisiol:
ready for next round of review

docs/dns/rrsets.rst
```diff
@@ -93,8 +93,8 @@ Field details:
     Records must be given in presentation format (a.k.a. "BIND" or zone file
     format). Record values that are not given in canonical format, such as
     ``127.000.0.1`` will be converted by the API into canonical format, e.g.
```
peterthomassen: adjust (leading zeros disallowed)

api/desecapi/dns.py
```diff
@@ -30,7 +30,7 @@ def from_text(cls, rdclass, rdtype, tok, origin=None, relativize=True):
             raise dns.exception.UnexpectedEnd
         return cls(rdclass, rdtype, strings)
 
-    def to_wire(self, file, compress=None, origin=None):
+    def _to_wire(self, file, compress=None, origin=None, canonicalize=False):
```
peterthomassen, line 33: dead code (and `canonicalize` argument is also dead)

peterthomassen (changes requested):
two small changes

api/desecapi/dns.py
nils-wisiol: very much alive :) change due to dnspython 2.0

peterthomassen (approved):

Issue #337, opened 2020-04-13
nslord,api: refactor how SOA is managed, and its field values


peterthomassen:
The SOA refresh field is currently 10800 seconds, which causes each slave to ask for the serial of each zone every 3 hours. This makes 120 queries per zone which pass through our VPN unnecessarily every day.

We don't need these queries as our replication mechanism only uses the serial field. However, pdns will always honor the refresh field value for zones which are in slave mode. Thus, let's increase it to a little over 2 weeks, as zones get AXFR'ed biweekly anyways due to resigning.

The obvious way to go at this would be https://doc.powerdns.com/authoritative/settings.html#soa-refresh-default, but the setting is deprecated and will disappear soon. The alternative approach is probably to specify the SOA record contents explicitly in the API request during zone creation. Need some insight on which serial to set etc.

peterthomassen:
Proposal for SOA field numbers:

- SOA RRset TTL: 300 (used as TTL for negative replies including NSEC3 records)
- SOA refresh: 2 weeks (our replication doesn't rely on this, and a high value keeps our frontends from asking all the time)
- SOA retry = refresh
- SOA expire: 4 weeks (all signatures will have expired anyways)
- SOA minimum: 3600 (for CDS, CDNSKEY, DNSKEY, NSEC3PARAM)

peterthomassen:
All of pdns' settings that allow defining a default value for an SOA field will be removed in pdns 4.4.0. This includes the currently used `default-soa-mail` and `default-soa-name`.

We need to refactor how SOA records are created. The straightforward solution is probably to just specify the RRset in the pdns domain creation request, https://github.com/desec-io/desec-stack/blob/master/api/desecapi/pdns_change_tracker.py#L89.

peterthomassen:
For the record, pdns 4.3 supports automatic publication of CDS and CDNSKEY records for key rollovers (provided the registry supports the RFC 7344 method).

I was considering to turn on automatic publication by default, but then realized that it's only needed when a key rotation is under way. [RFC 7344 Appendix B](https://tools.ietf.org/html/rfc7344.html#appendix-B) actually suggests deleting the records once the rollover is completed, so I guess it is not considered best practice to have them in place without an imminent key rollover.

One may think that permanent presence of those records could help registries determine how widespread CDS/CDNSKEY support is on auth servers, potentially expediting their own support. But that's purely hypothetical. If no other opinions are voiced, I am going to leave it off by default.

PR #388, opened 2020-05-17
20200517 SOA refactor

peterthomassen:
Removes deprecated pdns settings and changes SOA contents for new domains to more reasonable values, see #337.

SOA records of existing domains will not be altered. After this is merged, I suggest running the following query on dblord:
```
UPDATE records SET content=REPLACE(content, " 10800 3600 604800 60", " 1209600 1209600 2419200 3600") WHERE type = "SOA";
```

I verified locally that this works, and I verified that all production zones have SOA records endwing with ` 10800 3600 604800 60`.


api/desecapi/pdns_change_tracker.py
```diff
@@ -93,7 +93,16 @@ def pdns_do(self):
...
                     'dnssec': True,
                     'nsec3param': '1 0 127 %s' % salt,
-                    'nameservers': settings.DEFAULT_NS
+                    'nameservers': settings.DEFAULT_NS,
+                    'rrsets': [{
+                        'name': self.domain_name_normalized,
+                        'type': 'SOA',
+                        'ttl': 300,
+                        'records': [{
+                            'content': 'set.an.example. get.desec.io. 1 1209600 1209600 2419200 3600',
```
nils-wisiol, line 102: Let's document the rationale for each choice here or elsewhere (but in the code) https://github.com/desec-io/desec-stack/issues/337#issuecomment-623488329

peterthomassen: done

PR #392, opened 2020-05-18
Update delegation instructions when creating "managed DNS" domains

nils-wisiol:
Instructions on what to do with NS and DS records when creating a new "managed" domain are easy to be misunderstood, as the delegation process is somewhat unintuitive. This is an attempt at improving the wording, but I'm honestly not very convinced of this. Any better propositions?

In any case, the first two commits should be merged eventually.


nils-wisiol:
image

Just an example

webapp/src/views/Console/DomainDetailsDialog.vue
```diff
@@ -24,7 +24,8 @@
       </v-alert>
       <v-card-text>
         <p>
-          To properly secure your domain with DNSSEC, please forward the following information to your domain registrar:
+          Please forward the following information to the administrator/registry of
+          <code>.{{name.split('.').slice(1).join('.')}}</code>:
```
peterthomassen, line 28: I think it would be better to remove the left-most dot.

webapp/src/views/Console/DomainDetailsDialog.vue
```diff
@@ -24,7 +24,8 @@
       </v-alert>
       <v-card-text>
         <p>
-          To properly secure your domain with DNSSEC, please forward the following information to your domain registrar:
+          Please forward the following information to the administrator/registry of
```
peterthomassen, line 27: For most TLDs, domain owners cannot interact directly with the registry and instead have to talk to their registrar or provider. I would say:

"Please forward the following information to the organization where you registered the domain <code>...</code> (usually your provider or technical administrator):"

webapp/src/views/CustomSetup.vue
```diff
@@ -26,7 +26,9 @@
             </p>
             <h2 class="title">Secure Your Domain</h2>
             <p>
-              To properly secure your domain with DNSSEC, please forward the following information to your domain registrar:</p>
+              Please forward the following information to the administrator/registry of
+              <code>.{{$route.params.domain.split('.').slice(1).join('.')}}</code>:
```
peterthomassen, line 30: same comments as above apply

webapp/src/views/Console/DomainDetailsDialog.vue
nils-wisiol: With the text as suggested below, just {{name}} is fine.

webapp/src/views/Console/DomainDetailsDialog.vue
nils-wisiol: done

webapp/src/views/CustomSetup.vue
nils-wisiol: done

nils-wisiol:
@peterthomassen merge at your discretion

PR #398, opened 2020-05-20
feat(webapp): use built-in env support, remove bloat

### LINKED ISSUE #396: Fix dotenv setup for webapp
Remove envsubst construction and inject env variables directly using docker compose and `VUE_APP_` env variable name prefix (see #384)

nils-wisiol:
As suggested by @andreasnuesslein, this uses vuejs built-in env support and removes our custom env solution.

I checked that this works in both development and production, a version running in production mode is available at https://desec.t202004081956.dedyn.io/

Fixes #396


nils-wisiol:
`webapp/src/env.js` can savely be deleted from deployments

nils-wisiol:
Not sure what's the problem with the tests; looks like they are cancelled without reason. Will try re-running them after some time ...

webapp/src/views/SignUp.vue
```diff
@@ -157,9 +157,10 @@
 
 <script>
   import axios from 'axios';
-  import {LOCAL_PUBLIC_SUFFIXES} from '../env';
   import {domain_pattern, email_pattern} from '../validation';
 
+  let LOCAL_PUBLIC_SUFFIXES = process.env.VUE_APP_LOCAL_PUBLIC_SUFFIXES.split(' ');
```
peterthomassen, line 162: `const`

Also, we don't define this as a file-scoped variable in other contexts (but fine with me, I guess you have a good reason)

nils-wisiol: yeah, `this` is unavailable in the pattern checking mechanism further down the file

Issue #367, opened 2020-04-28
GitHub CI: Record Logs in Case of Failure


nils-wisiol:
Otherwise debugging is next to impossible. This can be a road block to contributions.

nils-wisiol:
This needs to be addressed asap, it is now blocking #401 and #398 and, I'm guessing, will block all PRs due to some broken dependency or something. Working on it in #402

nils-wisiol:
Closed by #402

PR #401, opened 2020-05-25
Expand update URL for FritzBox DynDNS

hcc23:
Small change in order to provide a copy-pastable URL for FritzBox users.


hcc23:
Hm... I don't quite get how changing the docu should lead to an [_internal server error_](https://github.com/desec-io/desec-stack/pull/401/checks?check_run_id=706381388#step:4:95)... Anything I need or can do to fix this?

nils-wisiol:
Hard to tell, you fell victim to #367 

I'm working on this right now in #402 and will hopefully change the GitHub Action such that there is useful debug output

nils-wisiol:
I fixed this, please rebase your changes onto current master. Thanks!

hcc23:
done, sorry for the delay.

peterthomassen:
Thanks, @hcc23 !

The problem with embedding the secret token ("password") in the URL is that this way, credentials end up in log files (URLs are logged for a few days). That's why our [docs recommend Basic authentication](https://desec.readthedocs.io/en/latest/dyndns/update-api.html#authentication).

Do you know whether it is possible to have a decent copy-paste URL for Fritzbox users, while still using Basic auth?

hcc23:
Hi @peterthomassen,

as far as I know there is no way to use the "custom" scheme with http basic authentication. I checked the [dyndns docu](https://service.avm.de/help/en/FRITZ-Box-7580/018/hilfe_dyndns) for my Fritz Box again and didn't find anything.
There might, however, be a different mechanic at use when a FritzBox user uses one of the pre-defined dyndns services.*

In my mind this would leave us with a couple of options which (not mutually exclusive):
* approach AVM about including desec.io as a dyndns provider in Fritz!OS (which would most likely also come with an influx of new customers) in the hopes that this method supports basic auth
* add a note in the docu highlighting the fact that URLs including the password will be logged 
* move the API token scoping up in your backlog so that the password in the URL is fairly limited in its powers ;)

*: on my box I can currently select from these dyndns providers: AnyDNS, DNS4BIZ.DE Business, DNS4BIZ.DE Premium, dyndns.org, Dyndnsfree.de, No-IP.com, selfhost.de, STRATO AG

nils-wisiol:
Maybe @domenukk can establish contact with AVM regarding this

nils-wisiol:
We can also consider configuring nginx such that the tokens are not logged. (Not sure if that's possible at all.)

hcc23:
I added a note about the credentials logging on the configure page and "tried" to also add a link to the basic auth section...
(never played with rst files before, so the linking is pureley open-loop and untested... sorry :/ )

nils-wisiol:
I'm not sure if I understand the benefit of these changes. Currently, my Fritzbox is using the old URL and is working fine. The new URL, on the other hand, still requires me to to also enter password and domain name in the separate fields in the Fritzbox UI. So what do we gain from these changes?

hcc23:
Hi @nils-wisiol 

Now _I_ am not sure if I understand what is happening here :/

So far I thought the the Fritzbox - when using the custom dyndns update method - does a simple curl call to the URL you specify and it is able to include certain parameters into that call by means of replacing placeholders (the stuff in the angled brackets, i.e. <ipaddr>, <ip6addr>, ...).

To me this would mean that, in order to do a joint ipv4 and ipv6 update, the URL would need to use the dreaded [query string method](https://desec.readthedocs.io/en/latest/dyndns/update-api.html#query-string-method-discouraged) in order to pass credentials:

    curl https://update.dedyn.io/?username=<username>&password=<passwd>&myipv4=<ipaddr>&myipv6=<ip6addr>

where FritzOS will substitute `<username>`, etc. with whatever the user provided in the respective GUI fields.

If I understand you comment above correctly you do _not_ provide username and password to the Fritz system - yet it still seems to work? How can that be? In my mind, using the URL that is currently in the docs is equivalent to 

    curl https://update.dedyn.io/?myipv4=<ipaddr>&myipv6=<ip6addr>

Sorry for maybe not seeing the obvious, but that cannot work... :/
Could you please give me some feedback on where I am wrong in my assumptions? Because if the original URL is indeed working than yes you are right: this update of the docs is unnecessary - yet there might be the need to other changes in order for newbies like myself to actually understand what is going on. :)

nils-wisiol:
I'm using the following configuration

image

I.e., my update URL only includes the IP6, but not username, token, password, etc. (Note that IP4 is missing as my connection uses carrier-grade NAT.)

But other fritzbox versions may behave differently?

peterthomassen:
@hcc23 The fact that the username and password placeholders are not part of the URL does not mean that Fritzbox does not use them for the "basic authentication" type. In this case, the credentials would be transmitted in a header and not in the URL.

Does this clarify the situation?

hcc23:
Ah, OK. This explains a lot :)

I didn't realize that the Fritzbox does indeed use basic auth and hence simply never populated the fields in the UI that I didn't include in the URL...

So it boils down to something like this:

    curl -u <username>:<passwd> https://update.dedyn.io/?myipv4=<ipaddr>&myipv6=<ip6addr>

hcc23:
Well, with this understanding: let's *close* this PR as it indeed doesn't make sense :)
(I might, however, create another one where this use case is described in a little more detail in the docs.)

nils-wisiol:
Alright, improvement on the documentation is always appreciated! Thanks!

PR #402, opened 2020-05-25
feat(): add log output to GitHub CI action

nils-wisiol:
Adresses #367


nils-wisiol:
This also fixes the broken tests by adapting to the API change of prometheus-client

PR #408, opened 2020-06-03
feat(api): adds AXFR retrieve endpoint, without DNSSEC data

nils-wisiol:
Enables the API user to retrieve her zone data as AXFR in one query. Data is taken straight from nslord, but could also be generated by dnspython using the API data in a later version. This is a first shot at #373. 

What's still missing:

- [x] make sure permissions are set correctly
- [x] change the docs


api/desecapi/pdns.py
```diff
@@ -98,6 +98,15 @@ def get_zone(domain):
     return r.json()
 
 
+def get_axfr(domain):
+    """
+    Retrieves a dict representation of the zone from pdns
+    """
+    r = _pdns_get(NSLORD, '/zones/' + pdns_id(domain.name) + '/export')
+
+    return r.json()['zone']
```
peterthomassen, line 107: The response body is the zone file, there's no `zone` field and no JSON

api/desecapi/models.py
```diff
@@ -334,6 +334,10 @@ def _partitioned_name(self):
         subname, _, parent_name = self.name.partition('.')
         return subname, parent_name or None
 
+    @property
+    def axfr(self):
```
peterthomassen, line 338: AXFR is the DNS Zone Transfer Protocol, RFC 5936. I don't think we should use that term here anywhere, because you can't "get" it, nor is it semantically clear would it would mean to "request https://.../domains/.../axfr" (one may expect a master-slave sync to happen?)

This is just getting the zone file in Bind format. This is different from AXFR format, which repeats the SOA record at the end to indicate "end of zone", for example. Maybe the clearest endpoint name would be `export`, and internally we can say `get_zonefile()` or something?

api/desecapi/models.py
```diff
@@ -334,6 +334,10 @@ def _partitioned_name(self):
...
+        return pdns.get_axfr(self)
```
peterthomassen, line 339: Should we allow filtering some records, e.g. export non-DNSSEC records only? (People usually don't want to import RRSIGs elsewhere as they are going to expire soon.)

peterthomassen:
> This is a first shot at #373.

Isn't #373 about import, while this is about export?

api/desecapi/models.py
peterthomassen: `export` also allows to later return things in another format, based on `Accept`, for example. That also makes me think if one could not just do `GET .../rrsets/` with `Accept: text/dns` (MIME type taken from Wikipedia)?

Also, can we stream this from the name server? The zone file may be very large.

api/desecapi/pdns.py
nils-wisiol: I read that in the docs, but apparently it's a JSON anyway. Maybe we need to adjust the `Accept` header...

nils-wisiol:
> > 
> 
> Isn't #373 about import, while this is about export?

The idea was to have an endpoint that can GET and POST, possibly even PATCH a zone file

api/desecapi/models.py
nils-wisiol: nslord only returns non-DNSSEC records this way

api/desecapi/pdns.py
peterthomassen: I tried it out in my nsmaster container and did not see any JSON (unless I completely misread the output). Anyway, let's double-check

api/desecapi/models.py
peterthomassen: aha! Cool. I did my experiments with nsmaster.

api/desecapi/models.py
peterthomassen: If `GET` and `POST` should be possible, we could override `rrsets/`, or call it `zone` (I think that's better than `axfr`).

peterthomassen:

api/desecapi/pdns.py
nils-wisiol: Note sure what's going on but this concern is covered by the e2e2 tests.

api/desecapi/models.py
nils-wisiol: Updated the name to be called "zonefile"; other suggestions warmly welcome. The ("outer") encoding is still JSON, maybe that offers better usability than messing with the Accept header?

nils-wisiol:
> > This is a first shot at #373.
> 
> Isn't #373 about import, while this is about export?

Merged #601 allows import of zonefiles; this PR allows export. An open question is if we accept updates to zones via zonefiles? (If so, how?) A related question is if we accept presigned zones, i.e. containing DNSSEC RRs.

Docs still missing (will be added after we decided the API).

api/desecapi/pdns.py
peterthomassen: Indeed, we're sending an `Accept: application/json` header, and when that is the case the response is as you observed. When sending '*/*' or no such header, the response is just the plain zonefile, which saves the conversion in both directions.

I've added two commits that (1) change the pdns request interface to allow specifying the Accept: header, (2) makes use of that and directly returns the pdns output as text/dns resource to the client, without any serialization. What do you think about that?

We may consider prepending a line or two, as follows:
```
; Zonefile $domain.name
; Exported from desec.io at $time
```

api/desecapi/models.py
peterthomassen: I think it's better to just return the plain zonefile as text/dns by default, that is when Accept: is missing or text/*. (I pushed some commits for inspiration). Perhaps we can return a JSON document when application/json is given, but I consider the plain export more important. (That's what all the media stuff in HTTP is about.)

OTOH, I don't feel strongly, so feel free to ignore. :)

peterthomassen (commented):
I think this is close to ready! Let's settle on what the output format should be, everything else should be a no-brainer. :)

peterthomassen:
Also, the view needs rate limit configuration. I think `throttle_scope = "dns_api_write_rrsets"` is sufficient (currently `2/s`, but per domain).

api/desecapi/pdns.py
nils-wisiol: done & done

api/desecapi/models.py
nils-wisiol: It's text/dns now without regard to the sent Accept header.

peterthomassen:
For reference, this (otherwise completely unrelated) comment shows how to mock a "download" with JavaScript: https://github.com/scopewu/qrcode.vue/issues/50#issuecomment-1055585303

nils-wisiol:
Before and after merging this, repo config needs to be updated wrt. which checks are required. I suggest making everything required except for the black formatting.

nils-wisiol:
I'm not planning to add GUI in this PR.

docs/rate-limits.rst
```diff
@@ -33,17 +33,17 @@ the API.  When several rates are given, all are enforced at the same time.
 +-----------------------------------------+----------+-------------------------------------------------------------------------------------------+
 | ``dyndns``                              | 1/min    | dynDNS updates (per domain).                                                              |
 +-----------------------------------------+----------+-------------------------------------------------------------------------------------------+
-| ``dns_api_read``                        | 10/s     | DNS read operations (e.g. fetching an RRset)                                              |
+| ``dns_api_cheap``                       | 10/s     | most DNS read operations (e.g. fetching an RRset)                                         |
```
peterthomassen, line 36: "most" - which not?

api/desecapi/views/records.py
```diff
@@ -140,3 +147,18 @@ def get_serializer(self, *args, **kwargs):
     def perform_create(self, serializer):
         with PDNSChangeTracker():
             super().perform_create(serializer)
+
+
+class ZonefileView(DomainViewMixin, generics.RetrieveAPIView):
```
peterthomassen, line 152: I realize only now that this could be an `@action`-decorated method on the `DomainViewSet`, in which case things would fit more neatly into the DRF structure, and also the domain object would be readily available (via `self.get_object()`), so you wouldn't need to do the `DomainViewMixin` magic.

I have a slight preference for going this way, because (1) it saves code duplication (e.g. `get_queryset()`) and (2) is easier to maintain (e.g. we won't forget adding 2FA permissions here when merging 2FA functionality, as the `DomainViewSet`'s permissions would apply).

docs/dns/domains.rst
```diff
@@ -257,6 +257,25 @@ and ``subname`` would just be ``_acme-challenge``.
...
 
 
+Exporting a Domain as Zonefile
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To export domain data in zonefile format, send a ``GET`` request to the
+``zonefile`` endpoint of this domain, i.e. to ``/domains/{name}/zonefile``::
+
+    curl -X GET https://desec.io/api/v1/domains/{name}/zonefile \
+        --header "Authorization: Token {token}"
```
peterthomassen, line 267: `{secret}`

docs/dns/domains.rst
```diff
@@ -257,6 +257,25 @@ and ``subname`` would just be ``_acme-challenge``.
...
+
+Note that this will return all data for this domain in plain zonefile
+format without JSON formatting, e.g.::
+
+    ; Zonefile for example.com exported from desec.io at 2022-08-26 16:03:18.258961+00:00
+    example.com.	1234	IN	NS	ns1.example.com.
+    example.com.	1234	IN	NS	ns2.example.com.
+    example.com.	300	IN	SOA	get.desec.io. get.desec.io. 2022082602 86400 3600 2419200 3600
+
+DNSSEC-specific record types are not included.
```
peterthomassen, line 277: This should preferably go before the code listing, because in the visual rendering, this sentence is easy to miss. (Don't feel strongly.)

nslord/Dockerfile
```diff
@@ -1,11 +1,12 @@
 FROM ubuntu:bionic
 
 RUN apt-get update && apt-get install -y \
-		dnsutils \
-		net-tools \
-		dirmngr gnupg \
-		faketime \
-	--no-install-recommends && apt-get clean && rm -rf /var/lib/apt/lists/*
+    dnsutils \
+    iptables \
```
peterthomassen, line 5: I don't think that's the right approach. We should first install the common stuff, and further down the non-common stuff, instead of installing everything everywhere.

Also, multi-stage builds may help (but I'm not sure).

nslord/Dockerfile
```diff
@@ -19,9 +20,11 @@ RUN echo 'deb [arch=amd64] http://repo.powerdns.com/ubuntu bionic-auth-46 main'
 RUN set -ex \
 	&& apt-key adv --keyserver keyserver.ubuntu.com --recv 0x1B0C6205FD380FBB \
 	&& apt-get update \
-	&& apt-get install -y pdns-server pdns-backend-mysql \
+	&& apt-get install -y pdns-server pdns-backend-mysql pdns-backend-pgsql postgresql-client-10 \
```
peterthomassen, line 23: I don't think that's the right approach. We should first install the common stuff, and further down the non-common stuff, instead of installing everything everywhere.

Also, multi-stage builds may help (but I'm not sure).

nslord/Dockerfile
```diff
@@ -19,9 +20,11 @@ RUN echo 'deb [arch=amd64] http://repo.powerdns.com/ubuntu bionic-auth-46 main'
...
 	# credentials management via envsubst
 	&& apt-get -y install gettext-base \
+	# VPN route
+	&& apt-get -y install iproute2 \
```
peterthomassen, line 27: I don't think that's the right approach. We should first install the common stuff, and further down the non-common stuff, instead of installing everything everywhere.

Also, multi-stage builds may help (but I'm not sure).

nsmaster/Dockerfile
```diff
@@ -19,7 +20,7 @@ RUN echo 'deb [arch=amd64] http://repo.powerdns.com/ubuntu bionic-auth-46 main'
 RUN set -ex \
 	&& apt-key adv --keyserver keyserver.ubuntu.com --recv 0x1B0C6205FD380FBB \
 	&& apt-get update \
-	&& apt-get install -y pdns-server pdns-backend-pgsql postgresql-client-10 \
+	&& apt-get install -y pdns-server pdns-backend-mysql pdns-backend-pgsql postgresql-client-10 \
```
peterthomassen, line 23: I don't think that's the right approach. We should first install the common stuff, and further down the non-common stuff, instead of installing everything everywhere.

Also, multi-stage builds may help (but I'm not sure).

peterthomassen (changes requested):

api/desecapi/views/records.py
peterthomassen: ... just realized that we already fell into that trap: `ZonefileView` is still missing at `throttle_scope` :)

docs/rate-limits.rst
nils-wisiol: Zonefile export :)

api/desecapi/views/records.py
nils-wisiol: I've added a fixup to set the throttle scope correctly. For using the `@action` decorator, I think it is more efficient if you do that in the 2FA PR.

docs/dns/domains.rst
nils-wisiol: fixed

docs/dns/domains.rst
nils-wisiol: fixed

nslord/Dockerfile
nils-wisiol: This deserves some more discussion. I've excluded this commit from this PR and moved it to #633.

nslord/Dockerfile
nils-wisiol: This deserves some more discussion. I've excluded this commit from this PR and moved it to #633.

nsmaster/Dockerfile
nils-wisiol: This deserves some more discussion. I've excluded this commit from this PR and moved it to #633.

nslord/Dockerfile
nils-wisiol: This deserves some more discussion. I've excluded this commit from this PR and moved it to https://github.com/desec-io/desec-stack/pull/633.

nils-wisiol:
This looks ready to go and is already rebased on top of #632

docs/rate-limits.rst
peterthomassen: I made this more explicit, see fixup

api/desecapi/views/records.py
peterthomassen: Permissions were inconsistent with those for `GET rrsets/`. While adjusting that, I also moved the functionality to an `@action` decorator-style view.

peterthomassen:
Rebased, and adjusted to things (see conversations above). Can you please review the two fixups?

docs/rate-limits.rst
```diff
@@ -33,11 +33,11 @@ the API.  When several rates are given, all are enforced at the same time.
 +-----------------------------------------+----------+-------------------------------------------------------------------------------------------+
 | ``dyndns``                              | 1/min    | dynDNS updates (per domain).                                                              |
 +-----------------------------------------+----------+-------------------------------------------------------------------------------------------+
-| ``dns_api_cheap``                       | 10/s     | most DNS read operations (e.g. fetching an RRset)                                         |
+| ``dns_api_cheap``                       | 10/s     | DNS read operations (e.g. fetching an RRset)                                              |
```
nils-wisiol: I understand the zonefile export as a 'read' operation, hence the change.

peterthomassen: yes, but it was in the wrong commit. The commit that adds the functionality now changes this line to say "except zonefile export" (see other fixup)

api/desecapi/views/domains.py
```diff
@@ -86,6 +101,12 @@ def perform_destroy(self, instance: Domain):
             with PDNSChangeTracker():
                 parent_domain.update_delegation(instance)
 
+    @action(detail=True, renderer_classes=[PlainTextRenderer])
+    def zonefile(self, request, name=None):
+        instance = self.get_object()
```
nils-wisiol, line 106: self.domain?

peterthomassen: The canonical thing in DRF is actually `self.get_object()`. I added the `domain` property only because the `TokenHasDomainRRsetsPermission` permission expects the view to have it.

nils-wisiol:
This can be squashed; before merging repo config needs to be adjusted to the new required tests.

peterthomassen (approved):

peterthomassen:
> This can be squashed; before merging repo config needs to be adjusted to the new required tests.

Done. - I'm not sure what to adjust. Can you make the changes?

nils-wisiol:
> 

done!

PR #409, opened 2020-06-04
Homepage breaking news about service disruption

nils-wisiol:
Adds information about the service disruption today to the home page, and uses the opportunity to introduce an webapp-wide user alerting system that could also be used for error conditions and other important messages.

TODO:

- [x] insert link to detailed service disruption thread

### DIFF
--- a/webapp/src/App.vue
+++ b/webapp/src/App.vue
@@ -89,6 +89,33 @@
     </v-app-bar>
 
     <v-content>
+      <v-banner v-for="alert in $store.state.alerts" :key="alert.id">
+        <v-icon
+          slot="icon"
+          color="warning"
+          size="36"
+        >
+          {{ alert.icon }}
+        </v-icon>
+        {{ alert.teaser }}
+        <template v-slot:actions>
+          <v-btn
+            color="primary"
+            text
+            :href="alert.href"
+            v-if="alert.href"
+          >
+            {{ alert.button || 'More' }}
+          </v-btn>
+          <v-btn
+            color="primary"
+            text
+            @click="$store.commit('unalert', alert.id)"
+          >
+            Hide
+          </v-btn>
+        </template>
+      </v-banner>
       <v-progress-linear
               :active="$store.getters.working"
               :indeterminate="$store.getters.working"
--- a/webapp/src/store/index.js
+++ b/webapp/src/store/index.js
@@ -8,6 +8,7 @@ export default new Vuex.Store({
     authenticated: false,
     token: {},
     work_count: 0,
+    alerts: [],
   },
   mutations: {
     login(state, token) {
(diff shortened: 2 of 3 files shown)


webapp/src/views/Home.vue
```diff
@@ -359,6 +364,18 @@ export default {
         text: "We provide easy integration with Let's Encrypt and their certbot tool.",
       },
     ],
+    breaking_news: [
+      {
+        id: 'news-20200604001',
+        start: new Date(Date.UTC(2020, 6 - 1, 4)),  // first day of showing
```
peterthomassen, line 370: You can just use something like `new Date('2020-06-04')` or `new Date('2020-06-04T02:00:00+02:00')`

webapp/src/views/Home.vue
```diff
@@ -359,6 +364,18 @@ export default {
...
+        end: new Date(Date.UTC(2020, 6 - 1, 8)),  // first day of not showing
+        icon: 'mdi-heart-broken',
+        teaser: 'deSEC.io DNS operations experienced a partial service disruption from 4 June 2020 12am UTC to ' +
```
peterthomassen, line 373: Other proposal:
> DNS operations experienced a partial service disruption on June 4, 2020, starting at 12am UTC. The issue was resolved at 7am UTC.

webapp/src/views/Home.vue
```diff
@@ -359,6 +364,18 @@ export default {
...
+                '4 June 2020 7am UTC. The issue has been resolved.',
+        button: 'Learn More',
```
peterthomassen, line 375: If the default is changed, this can go away

webapp/src/App.vue
```diff
@@ -89,6 +89,33 @@
...
+        </v-icon>
+        {{ alert.teaser }}
+        <template v-slot:actions>
+          <v-btn
+            color="primary"
+            text
+            :href="alert.href"
+            v-if="alert.href"
+          >
+            {{ alert.button || 'More' }}
```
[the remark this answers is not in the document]
nils-wisiol: I opted for the more neutral 'more' as standard text to make the alert implementation content agnostic.

webapp/src/store/index.js
```diff
@@ -21,9 +22,30 @@ export default new Vuex.Store({
     working(state, working = true) {
       state.work_count += working ? 1 : -1;
     },
+    alert(state, alert) {
+      for (const known_alert of state.alerts) {
+        if (alert.id === known_alert.id) {
+          return;
+        }
+      }
+      state.alerts.push(alert);
```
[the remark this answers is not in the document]
nils-wisiol: [As far as I understand the docs](https://vuejs.org/v2/guide/list.html#Mutation-Methods), vue cannot track object changes, hence the use of an array is required for automatic update of the DOM.

webapp/src/views/Home.vue
```diff
@@ -359,6 +364,18 @@ export default {
         text: "We provide easy integration with Let's Encrypt and their certbot tool.",
       },
     ],
+    breaking_news: [
+      {
+        id: 'news-20200604001',
```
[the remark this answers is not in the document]
nils-wisiol: It should be consistent between instantiations of `Home.vue`, if we ever wanted to introduce a list of read messages. (Currently, it's displayed every time you go on the home page.)

webapp/src/views/Home.vue
nils-wisiol: The usage of that is [strongly discouraged](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) by Mozilla, and I'm not sure how it goes together with UTC dates

webapp/src/views/Home.vue
nils-wisiol: done

webapp/src/views/Home.vue
nils-wisiol: see comment above

webapp/src/views/Home.vue
peterthomassen: I immediately believe there are browser incompatibilities, if you don't specify a timezone. However, if you do (UTC is denoted by `Z` after time part of the string), then I can't actually imagine any ambiguities.

It's not discouraged because it's going to disappear, but because the behavior for certain inputs is not well defined. I think it would be fine if we use it with "fully qualified date-times" only. But if you feel uncomfortable with that, let's stick to what you have.

PR #420, opened 2020-07-03
feat(webapp): RRset GUI, closes #358

### LINKED ISSUE #358: Add web-interface for zone management
The API is nice but if you just want to check/edit some static records using the API is of no comfort and a no-go for John Does.

I suggest to offer a HTML5/ECMAscript web-interface based on the API to view/add/update/delete RRs. This would also allow to add assistants for complex records like SRV, NAPTR, TLSA, SMIMEA, OpenPGPkey, etc.

peterthomassen:
Create/modify/delete RRsets.


nils-wisiol:
@Jennybaga and I played around with the GUI and found the following issues, bugs, suggestions; important and unimportant, in no particular order:

(In the following, I'm using 'record' and 'record sets' interchangeably. Shame on me!)

- [ ] The rows in the domain table do not look clickable, maybe some users cannot found the GUI.
- [ ] Pagination is broken, right? Then we should hide the controls
- [ ] The subname cannot be edited -- that's a limitation of the API, right?
- [ ] For TXT records in the first row of the table, the content field has two content headers: one from the content field, one from the column header. My suggestion would be to not use any text field header if there is only one text field (as is currently for most types).
- [ ] Add content length limit to TXT fields, or in general: add more validation rules on the client side
- [ ] If you edit the content after changing the type of any row, saving yields an error message saying "not found".
- [ ] Feature idea: display TTLs with units instead of just integers. (Also, is it a float field? I can input 'e', but not 'a'.)
- [ ] TXT content fields are too narrow, especially in the 'create new' dialoge
- [ ] For clarity, I suggest to change the label of the 'add value' button below the content fields to 'add another value'
- [ ] 'add value' button should be disabled if the last content field is empty
- [ ] If you add a too long TXT record in the 'create new' dialog, the error message will contain the faulty value, resulting in content overflow in the error message. The error message field should have it's overflow setting such that it breaks the line.
- [ ] For MX, "bla." is (on the client-side) not accepted as a valid hostname
- [ ] For saving rows, the user does not get 'pro-active' feedback if saving worked or not. (Just the absence of an error message is indicating that it should have worked.)
- [ ] Related: there is no way to tell if a row has unsaved changes or not. (How about an undo button somewhere?)
- [ ] In the 'create new' dialog, there is no error message when I try to create a DNSKEY record (API reponded with 'you cannot tinker')
- [ ] In the table, the content fields are quite narrow, especially for TXT, but also for MX and possibly AAAA. Let's remove the margin around the table (like GitHub does when you switch to the 'Projects' tab.)
- [ ] In the table, saving does not work after changing the record type
- [ ] Table search doesn't include content fields

peterthomassen:
Thanks @Jennybaga and @nils-wisiol for trying to break the GUI ;-) I took care of all items I checked above (reordered them a bit). For the remaining ones, here are some comments:


- [x] The subname cannot be edited -- that's a limitation of the API, right?

Yes. I made the field uneditable in the GUI.

- [ ] Pagination is broken, right? Then we should hide the controls

The pagination here is clientside pagination and is there to keep the table short and fast. People who don't like it can choose a different page size at the bottom ("All"). 

[Server-side pagination only kicks in case there are more than 500 RRsets](https://github.com/desec-io/desec-stack/blob/master/api/api/settings.py#L106) (or domains, or tokens, in the respective tables). At the moment, a red error balloon appears at the bottom, displaying the API message which starts with "Pagination required", and the table will not load any items. I would say that right now, that's good enough; people who have that many records or domains will probably use the API anyways.

We should improve that at some point. However, it's not clear how exactly to handle search etc, so I'd skip it for now.
 
- [ ] Add content length limit to TXT fields, or in general: add more validation rules on the client side

Doesn't that depend on how long the other record values in the same RRset are? The complicated logic is under way in #385, and I'm not sure if it is a good idea to replicate it in the GUI. (Constraints may change on the server side when #334 is fixed, and I'd like to avoid having to change several places.) People will almost never encounter this problem, and if they do, I think they can afford waiting for the roundtrip to get the API error message. I'd propose to make this a "won't fix"/

- [ ] Feature idea: display TTLs with units instead of just integers. (Also, is it a float field? I can input 'e', but not 'a'.)

I think that confuses users who are trying to copy-paste some number that someone told them, and then suddenly are confronted with having to choose a unit. So, if anything, the default unit should be "seconds", and other units should be optional. I'm not sure if it's worth developing a neat widget for that.

The field currently is rendered as `<input type="number">` which apparently allows for scientific notation. It also allows for negative values. TODO: Add validation for numbers only, and minimum value = `domain.minimum_ttl`. The second is a bit of work, so I'd like to move it to a separate issue if you're ok with that?

- [ ] 'add value' button should be disabled if the last content field is empty

I think there's a use case for several empty fields, namely when you create a few empty fields (with the mouse) and then fill them in quick succession using the tab key. While that's not common, it's a small benefit, and I wanted to share the thought. -- if you think it's more harmful or confusing that anything, let me know and I'll fix it.

- [ ] If you add a too long TXT record in the 'create new' dialog, the error message will contain the faulty value, resulting in content overflow in the error message. The error message field should have it's overflow setting such that it breaks the line.

TODO

- [ ] Related: there is no way to tell if a row has unsaved changes or not. (How about an undo button somewhere?)

Undoing requires state, so I'd rather not do it. We can sort of add a flag, but I don't think it's necessary for the first round.

- [ ] For saving rows, the user does not get 'pro-active' feedback if saving worked or not. (Just the absence of an error message is indicating that it should have worked.)

TODO

nils-wisiol:
>> Related: there is no way to tell if a row has unsaved changes or not. (How about an undo button somewhere?)

> Undoing requires state, so I'd rather not do it. We can sort of add a flag, but I don't think it's necessary for the first round.

While I don't think undoing is very important, I think it's necessary to see if a row has changes or not. How would the user know when/if her changes were saved?

Undo may be easy to implement by reloaded some state from the server.

nils-wisiol:
It really looks awesome now! :star_struck: 

Second round:

- [ ] for MX, when I input 'example.com' as hostname, now the neat 'must end in a dot' error message does not show up anymore; instead it just says 'invalid hostname'
- [ ] in MX hostname field, I cannot undo pasting (by pressing ctrl-z or right click -> undo)
- [ ] in MX hostname field, I cannot highlight content by setting the cursor somewhere in the middle and then pressing shift+end
- [ ] in MX record, pasting
    ```
    10 example.com.
    20 example.com.
    ```
    results in a single record with hostname 'example.com.20 example.com.'
- [ ] I can save a record set even though client validation did not go through. Not sure if that's a bug or a feature :smile: 
- [ ] In the create new dialog, TXT and SPF fields do not have a label. (I guess it's better to have the double-label in the table than no label here.)

> I think it's necessary to see if a row has changes or not.

This is especially true after deleting single records from a record set (user doesn't know if changes were already saved) and after saving threw an error (let's say, TXT record without quotes).

webapp/src/components/DomainDetailsDialog.vue
```diff
@@ -0,0 +1,130 @@
...
+        Your domain <b>{{ name }}</b> has been successfully created!
+      </v-alert>
+      <v-card-text>
+        <p>Please forward the following information to your domain registrar:</p>
+        <div class="caption font-weight-medium">
+          NS records
+        </div>
+        <pre class="mb-3 pa-3">
+ns1.desec.io
+ns2.desec.io
```
nils-wisiol, line 32: Take those from the environment setting (also, second one is outdated)

webapp/src/components/DomainDetailsDialog.vue
```diff
@@ -0,0 +1,130 @@
...
+</pre>
+        <v-layout
+          flex
+          align-end
+        >
+          <div class="caption font-weight-medium">
+            DS records
+          </div>
+          <v-spacer />
+          <div v-if="!copied">
```
nils-wisiol, line 42: I think copy-paste was broken in other instances of this code. (Also, can we avoid this code duplication?)

webapp/src/views/Console/DomainList/NewDomainDialog.vue
```diff
@@ -0,0 +1,97 @@
+<template>
```
nils-wisiol, line 1: This file seems to be unrelated to the record set management?

webapp/src/views/Console/DomainList/index.vue
```diff
@@ -0,0 +1,258 @@
+<template>
```
nils-wisiol, line 1: extra file?

webapp/src/views/Domain/CrudDomain.vue
```diff
@@ -0,0 +1,101 @@
...
+          value: 'touched',
+          readonly: true,
+          datatype: 'TimeAgo',
+          searchable: false,
+        },
+      },
+      actions: [
+      ],
+      paths: {
+        list: 'domains/::{domain}/rrsets/', // TODO dangerous?
```
nils-wisiol, line 89: The domain value inserted here is part of the URL. Is it possible to craft an URL such that the user here triggers and authenticates unwanted API actions?

webapp/src/views/Domain/RRset.vue
```diff
@@ -0,0 +1,144 @@
...
+<script>
+  import {required, integer, minValue} from 'vuelidate/lib/validators';
+  import Record from '@/components/Field/Record.vue';
+  import RecordA from '@/components/Field/RecordA.vue';
+  import RecordCNAME from '@/components/Field/RecordCNAME.vue';
+  import RecordMX from '@/components/Field/RecordMX.vue';
+  import RecordSRV from '@/components/Field/RecordSRV.vue';
+
+
+  const MinTTL = 10;
```
nils-wisiol, line 76: inaccurate, and in my test, ineffective

webapp/src/views/Domain/index.vue
```diff
@@ -0,0 +1,291 @@
+<template>
```
nils-wisiol, line 1: extra file?

webapp/src/components/DomainDetailsDialog.vue
peterthomassen: dead file, removed

webapp/src/components/DomainDetailsDialog.vue
peterthomassen: dead file, removed

webapp/src/views/Console/DomainList/NewDomainDialog.vue
peterthomassen: dead flie, removed

webapp/src/views/Console/DomainList/index.vue
peterthomassen: dead flie, removed

webapp/src/views/Domain/RRset.vue
peterthomassen: dead flie, removed

webapp/src/views/Domain/index.vue
peterthomassen: dead flie, removed

webapp/src/views/Domain/CrudDomain.vue
peterthomassen: As far as I can see, that is not possible: it only would be possible to construct other paths that end in `rrsets/`, but there are in fact no other such endpoints. If you try to insert something else so that the `rrsets/` in the end is in excess, our API gives 404.

I admit though that this is probably not the ideal solution. I also remember we had the same discussion when you initially wrote this, but we decided it's ok.

peterthomassen:
I fixed a few things, except the following:

- The selection / undo issues are not currently feasible to solve without undue effort due to the field structure.
- I'd like to fix the formatting of error messages in the next iteration.
- I am going to add a way to tell whether a row has changed, and visual feedback on saving (including when hitting Enter, and including error conditions).

Once the last bullet is done, I'd like to roll this out as a first release. Ok for you?

peterthomassen:
There is no visual feedback on saving. You can hit enter in any field in the RRset editing table, and will see the row flashing as confirmation that the save request has been sent. If it fails, a red snackbar with the error message is displayed at the bottom.

I did not yet implement the "changed" state indicator yet, mainly because it would be nice to have it on a per-row level, but it is difficult to get that state information into the `item.<name>` slot (where also the actions live, on the right) -- this is due to the fact that at this point in the code, the row index is not available. A solution would either involve event signaling, or a shadow copy of all rows that are constantly compared. Given the complexity, I didn't want to start it in a rush. -- Also, I think most people will not expect a "changed" indicator to be there and just hit "enter" when done.

From my perspective, this is ready for the first rollout (although, obviously, far from perfect). Release fast and often. :)

PR #426, opened 2020-08-10
20200810 miscellaneous fixes

peterthomassen:
- schedule savenger daily
- improve GUI validation

I figured both is not urgent and changes are small, so I created just one PR to avoid rebasing and other logistics.


nils-wisiol (approved):
nice

PR #433, opened 2020-08-25
feat(docs): document rate limits

peterthomassen:


nils-wisiol (approved):
lgtm; maybe add a code-comment that reminds future us to update the docs when we update limits.

PR #437, opened 2020-08-27
feat(api): introduce immortal domains (exempt from scavenger)

peterthomassen:
This PR introduces a distinct flag to tell whether a domain should be affected by the scavenger or not.

The scavenger itself only processes domains that end with a local public suffix (as it has always done). Strictly, this check is now unnecessary, as we could now explicitly label all other domains as immortal (new non-dyndns domain are set immortal automatically). But I'm not sure if it's a good idea to remove the suffix filter from the scavenger. Thoughts?


nils-wisiol:
Let's have only one mechanism for one thing: I'm voting to remove the additional check

api/desecapi/models.py
```diff
@@ -231,10 +232,17 @@ class RenewalState(models.IntegerChoices):
...
     minimum_ttl = models.PositiveIntegerField(default=get_minimum_ttl_default)
-    renewal_state = models.IntegerField(choices=RenewalState.choices, default=RenewalState.FRESH)
+    renewal_state = models.IntegerField(choices=RenewalState.choices, default=RenewalState.IMMORTAL)
     renewal_changed = models.DateTimeField(auto_now_add=True)
     _keys = None
 
+    def __init__(self, *args, **kwargs):
+        name = kwargs.get('name')
+        if name and kwargs.get('renewal_state') is None \
+                and any(name.endswith(f'.{lps}') for lps in settings.LOCAL_PUBLIC_SUFFIXES):
```
nils-wisiol, line 242: move `any(name.endswith(f'.{lps}') for lps in settings.LOCAL_PUBLIC_SUFFIXES)` into a property of `Domain`

api/desecapi/tests/test_user_management.py
```diff
@@ -900,21 +900,30 @@ def assertRenewDomainVerificationSuccessResponse(self, response):
...
-    def test_renew_domain_non_local_public_child(self):
-        user = User.objects.get(email=self.email)
-        domain = self.create_domain(owner=user)
+    def _test_renew_domain_immortal(self, domain):
         for days in [182, 184]:
             domain.published = timezone.now() - timedelta(days=days)
             domain.renewal_changed = domain.published
             domain.save()
 
-            self.assertEqual(Domain.objects.get(pk=domain.pk).renewal_state, Domain.RenewalState.FRESH)
```
nils-wisiol, line 911: test for FRESH not needed anymore?

nils-wisiol (approved):
nice

api/desecapi/models.py
peterthomassen: done

api/desecapi/tests/test_user_management.py
peterthomassen: refactored tests to have a DYN class and a non-DYN class. In each class, there's a test that FRESH domains only reach the next scavenging step when older than 182 days (compare `test_renew_domain_fresh_182_days()` vs `test_renew_domain_fresh_183_days()`). I think that covers it.

peterthomassen:
> Let's have only one mechanism for one thing: I'm voting to remove the additional check

Remove, and reworked the tests a bit to address your inline comment. It would be great if you could look at it once more before we proceed!

api/desecapi/tests/base.py
```diff
@@ -779,13 +779,22 @@ def requests_desec_domain_creation(cls, name=None):
...
+            delegate_at = cls._find_auto_delegation_zone(domain.name)
+            requests += [
+                cls.request_pdns_zone_update(name=delegate_at),
+                cls.request_pdns_zone_axfr(name=delegate_at),
+            ]
+
+        return requests
+
     @classmethod
     def requests_desec_domain_creation_auto_delegation(cls, name=None):
```
nils-wisiol, line 799: same for creation? 😬

api/desecapi/tests/test_user_management.py
```diff
@@ -900,20 +900,27 @@ def assertRenewDomainVerificationSuccessResponse(self, response):
...
+            self.assertEqual(Domain.objects.get(pk=domain.pk).renewal_state, Domain.RenewalState.IMMORTAL)
             self.assertEqual(Domain.objects.get(pk=domain.pk).renewal_changed, domain.renewal_changed)
             self.assertEqual(Domain.objects.get(pk=domain.pk).published, domain.published)
-            self.assertEqual(len(mail.outbox), 0)
+
+    def test_renew_domain_non_local_public_child(self):
+        domain = self.my_domains[0]
+        self._test_renew_domain_immortal(domain)
+
+    def test_renew_domain_immortal_local_public_suffix_child(self):
```
nils-wisiol, line 919: naming confusion

api/desecapi/tests/test_user_management.py
```diff
@@ -961,8 +970,9 @@ def test_renew_domain_fresh_183_days(self):
         self.assertEqual(Domain.objects.get(pk=domain.pk).renewal_state, Domain.RenewalState.FRESH)
         self.assertLess(timezone.now() - Domain.objects.get(pk=domain.pk).renewal_changed, timedelta(seconds=1))
 
+        # Check that other domains aren't affected
         for domain in self.my_domains[1:]:
```
nils-wisiol, line 974: assert len(self.my_domains[1:])? just in case ...

api/desecapi/tests/base.py
peterthomassen: no, because during creation time, we don't have an object (only the name)

api/desecapi/tests/test_user_management.py
peterthomassen: done

api/desecapi/tests/test_user_management.py
peterthomassen: done

PR #448, opened 2020-09-08
20200813 email case

peterthomassen:


peterthomassen:
Before deploying this, we need to check and clean up any accounts that would be considered duplicates with the new collation.

peterthomassen:
I just realized that converting the email field to `django.contrib.postgres.fields.CIText` may be an easier solution, with less drawbacks:

- no manual messing with raw SQL required
- no thinking about collations
- no dropping of the index required (the index is necessary for `LIKE` comparisons, which are broken when the index is dropped)
- ...?

On the other hand:

- `CIText` requires running a `CREATE EXTENSION` command, which the API database user can only do starting with Postgres 13 (will be released this fall).
- Having collations at hand may come in handy in case we want to pursue Universal Acceptance at some point.

@nils-wisiol Let me know what you think, and I'll implement this with `CIText` if you think it make sense.

nils-wisiol (approved):
lgtm

nils-wisiol:
Let's test CIText with the tests from this PR and then decide a way forward; let's take the discussion for existing accounts off-line.

peterthomassen:
superseded by #455

PR #450, opened 2020-09-08
feat(api): remove expiration from domain renewal code, closes #422

### LINKED ISSUE #422: api: Extend validity of domain renewal confirmation links
This only affects dynDNS domains.

Currently, the validity is 24 hours. It should be at least a week, or until the scheduled domain deletion date.

The validity period is not encoded in the link, but determined at validation time. At that point, the payload (including domain name) is not yet known, so using the scheduled domain deletion date is not an option.

Alternatively, we may just set it to 4 weeks: We delete domains 4 weeks after the first notification, so that would cover all cases.

No expiration would also work (links are invalidated anyway when the domain owner or other significant things change). ~However, that opens the opportunity for people to renew their domain infinitely without reading their emails. That's not what we want - if people want automation, they should renew their domain by changing some DNS information.~ `Domain.renewal_changed` is included in the action state, so using the link invalidates it as well.

Implementation proposal: At validation time in `AuthenticatedActionSerializer`, use `self.context['view']` to retrieve the view, and make the validity period an attribute of the view classes (with a reasonable default in the base class).

peterthomassen:
Links are still invalidated when significant domain attributes change,
such as the domain owner, or the renewal timestamp.  Using the link
thus renders it invalid.


api/desecapi/serializers.py
```diff
@@ -681,7 +681,8 @@ def to_internal_value(self, data):
         except KeyError:
             raise serializers.ValidationError({'code': ['This field is required.']})
         except ValueError:
-            validity = self.context['view'].validity_period or 'unlimited'
+            if (validity := self.context['view'].validity_period) is None:
+                validity = 'unlimited'
```
peterthomassen, line 685: The previous version (`validity = self.context['view'].validity_period or 'unlimited'`) would give `'unlimited'` if e.g. `validity_period == datetime.timedelta(hours=0)`, which was not intended. I don't like this solution, but it's the least excessive one I could come up with. What do you think, @nils-wisiol ?

peterthomassen: How about this?

```python
validity = 'unlimited' if (_ := self.context['view'].validity_period) is None else _
```

Looks very PHP-ey ...

peterthomassen: maybe the following is best:

```python
validity = self.context['view'].validity_period
if validity is None:
    validity = 'unlimited'
```

peterthomassen:
I just realized that the emails we send also need to be adjusted, as they contain hints on the link validity period which now depends on the verification view. Earlier, we had hardcoded the setting into the template context in `User.send_email()`.

I now changed it so that the view generating the link itself provides the hint via the context dictionary. This involves parsing the confirmation link to figure out the verification view, of which the validity period is an attribute. We repeat this small piece of code essentially very similarly in the 4 cases we have. -- I'm not really satisfied with this, but I'm not sure whether any extra engineering is appropriate. Looking forward to learning what you think.

api/desecapi/serializers.py
nils-wisiol: validity = 'unlimited' if (_ := self.context['view'].validity_period) is None else _

omg I didn't know this was possible 😳

Let's stick with the current version. Explicit it better. (But also this comes to mind: `getattr(self.context['view'], 'validity_period', default=None`)

api/desecapi/serializers.py
```diff
@@ -668,15 +667,25 @@ def to_representation(self, instance: models.AuthenticatedUserAction):
...
         except KeyError:
             raise serializers.ValidationError({'code': ['This field is required.']})
         except ValueError:
-            validity = settings.VALIDITY_PERIOD_VERIFICATION_SIGNATURE
+            validity = self.context['view'].validity_period
+            if validity is None:
+                validity = 'unlimited'
             raise serializers.ValidationError({
-                'code': [f'This code is invalid, most likely because it expired (validity: {validity}).']
+                'code': [f'This code is invalid, possibly because it expired (validity: {validity}).']
```
nils-wisiol, line 688: Now this can result in an error message reading `This code is invalid, possibly because it expired (validity: unlimited)`, let's avoid that

api/api/settings_quick_test.py
```diff
@@ -1,14 +1,15 @@
...
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': 'desecapi.sqlite',
-        'TEST': {
-            'CHARSET': 'utf8mb4',
-        },
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': 'desec',
+        'USER': 'desec',
+        'HOST': '127.0.0.1' if os.environ.get('DJANGO_TEST', '') == '1' else 'dbapi',
```
nils-wisiol, line 12: Is this variable name defined and used by django? Otherwise is should be prefixed with DESECSTACK

nils-wisiol (commented):
Regarding the computation of which view class is responsible, I couldn't find a better solution than putting the whole thing through the router/resolver. But still, I think this code should be centralized somewhere in order to reduce duplication.

Regarding the Django tests using postgres, I couldn't get it to work on my system. I pushed updated instructions for the README, but am stuck with 

> Got an error creating the test database: permission denied to create database

api/api/settings_quick_test.py
peterthomassen: done

api/desecapi/serializers.py
peterthomassen: done

peterthomassen:
I addressed all comments and also moved the construction of the confirmation link into a separate, more general function.

nils-wisiol (approved):
lgtm

PR #454, opened 2020-09-29
20200929 git zones

nils-wisiol:
This PR introduces a repository of zone files for every domain at the api container.

Open todos:

- [x] periodically update zone files after key rotation
- [x] check initialization with pre-existing domains
- [x] check AXFR timeout (large zone & empty signature cache?)
- [x] use relaxed check for equality in e2e2 record content tests
- [x] make sure that "eventual" checks in e2e2 are executed correctly

If this is ever used as replication method, this needs to be done:

- [ ] remove outdated replication code
- [ ] improve 2-phase commit procedure


api/desecapi/replication.py
```diff
@@ -0,0 +1,125 @@
...
+        #  but due to celery privilege drop, $HOME is still /root
+        #  and this gives a 'permission denied' warning in the log
+
+    def get_head(self):
+        return self.get_commit('HEAD')
+
+    def get_commit(self, rev):
+        try:
+            commit_hash = self._git('show', rev, '--format=%s', '-s')[:-1]
+            commit_message = self._git('show', rev, '--format=%H', '-s')[:-1]
```
peterthomassen, line 80: other way around

api/desecapi/replication.py
```diff
@@ -0,0 +1,125 @@
...
+        if rcode != 0:
+            raise GitRepositoryException('git returned nonzero error code', cmd=cmd, rcode=rcode, stdout=stdout, stderr=stderr)
+
+        print('\n'.join(['<<< ' + s for s in stdout.split('\n')]))
+        return stdout
+
+    def commit_all(self, msg=None):
+        self._git('add', '.')
+        status = self._git('status')
+        if 'nothing to commit' not in status:
```
peterthomassen, line 61: May always be false if a zone name contains this string. Can the same be achieved by inspecting the return code?

api/desecapi/replication.py
```diff
@@ -0,0 +1,125 @@
...
+            stdout = p.stdout.read()
+            try:
+                stderr, stdout = stderr.decode(), stdout.decode()
+            except UnicodeDecodeError:
+                GitRepositoryException('git stdout or stderr was not valid unicode!', cmd=cmd, stderr=stderr, stdout=stdout)
+
+        if rcode != 0:
+            raise GitRepositoryException('git returned nonzero error code', cmd=cmd, rcode=rcode, stdout=stdout, stderr=stderr)
+
+        print('\n'.join(['<<< ' + s for s in stdout.split('\n')]))
```
peterthomassen, line 55: I associate `<<<` with input (not with output). Consider switching? (Anyway, it seems to me that there is no best choice.)

api/desecapi/replication.py
```diff
@@ -0,0 +1,125 @@
...
+        try:
+            commit_hash = self._git('show', rev, '--format=%s', '-s')[:-1]
+            commit_message = self._git('show', rev, '--format=%H', '-s')[:-1]
+            return commit_hash, commit_message
+        except GitRepositoryException as e:
+            print('@@@@', e.stderr)
+            return None, None
+
+    def remove_history(self, before: timedelta):
+        rev = self._git('log', f'--before={(datetime.now() - before).isoformat()}', '-1', '--format=%H')
```
peterthomassen, line 87: What does `-1` do?

api/desecapi/replication.py
```diff
@@ -0,0 +1,125 @@
...
+            self.commit_all(msg='Inception or Recovery')
+            self.full_refresh()
+            self.commit_all(msg='Full Refresh')
+
+    def full_refresh(self):
+        print('WARNING: Full refresh of replication zone repository!')
+        for name in models.Domain.objects.values_list('name', flat=True):
+            self.axfr(name)
+
+    def axfr(self, name, ns='172.16.1.11'):
```
peterthomassen, line 109: I assumed this would trigger an AXFR to some frontend server. Maybe call this `export` or `dump`? (It's also not important that it causes an AXFR query; we might do something else in the future, e.g. query the pdns API.)

api/desecapi/replication.py
```diff
@@ -0,0 +1,125 @@
...
+        z = dns.zone.from_xfr(dns.query.xfr(ns, name, timeout=2))
```
peterthomassen, line 110: Under what load and zone size conditions is the 2s timeout exhausted?

api/desecapi/replication.py
```diff
@@ -0,0 +1,125 @@
...
+        if '/' in name or '\x00' in name:
```
peterthomassen, line 111: check before fetching zone

api/desecapi/replication.py
nils-wisiol: Oh I see. I chose < and > in analogy to the communication with the API: there, ">>>" means sending, "<<<" receiving.

api/desecapi/replication.py
nils-wisiol: shows just the first log entry

api/desecapi/replication.py
nils-wisiol: still needs to be evaluated. Probably depends heavily on the signature cache 😐

api/desecapi/replication.py
nils-wisiol: still needs to be evaluated. Probably depends heavily on the signature cache neutral_face

api/desecapi/replication.py
nils-wisiol: I changed it to consider the output of `git diff --numstat`, which is empty iff there are no changes

api/desecapi/replication.py
nils-wisiol: image

nils-wisiol:
# Key rollover:

Over time, pdns generates our signatures like this:

```
RRSIG(1) Inception                                                    RRSIG(1) Expiry
|                                                                                   |
v                                                                                   v
|================================ RRSIG(1) validity ================================|
                            |================================ RRSIG(2) validity ================================|
                            ^                                                                                   ^
                            |                                                                                   |
                            RRSIG(2) Inception                                                    RRSIG(2) Expiry

                            |----- RRSIG(1) served -----|----- RRSIG(2) served -----|

|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
Thu Fri Sat Sun Mon Tue Wed Thu Fri Sat Sun Mon Tue Wed Thu Fri Sat Sun Mon Tue Wed Thu Fri Sat Sun Mon Tue Wed Thu
                                                        ^
                                                        |
                                                        RRSIG roll-over(1 to 2)
```

The earliest point in time we can start replicating new signatures to the frontends is Thursday morning 00:00hrs, as that is when pdns starts serving the new signatures.

The latest point in time we need to push a zone to the frontend is (Thursday 00:00hrs a week later) - (the longest TTL of an RR set in that zone).

Let's say our maximum TTL is three days, then we had four days for refreshing all zones on the frontends:

```
RRSIG(1) Inception                                                    RRSIG(1) Expiry
|                                                                                   |
v                                                                                   v
|================================ RRSIG(1) validity ================================|
                            |================================ RRSIG(2) validity ================================|
                            ^                                                                                   ^
                            |                                                                                   |
                            RRSIG(2) Inception                                                    RRSIG(2) Expiry

                            |----- RRSIG(1) served -----|----- RRSIG(2) served -----|

|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
Thu Fri Sat Sun Mon Tue Wed Thu Fri Sat Sun Mon Tue Wed Thu Fri Sat Sun Mon Tue Wed Thu Fri Sat Sun Mon Tue Wed Thu
                                                        ^               ^
                                                        |               |
                                                        RRSIG roll-over(1 to 2)
                                                                        |
                                                                        All zones need to have RRSIG(2) on the frontend
```

This is somewhat unsatisfying as there would be four weekdays with higher load and three weekdays with lower load.

Independently, we wish to roll over the signatures "in the background", while giving priority to any RR set changes that should be deployed immediately. ~~I hope~~ I'm assuming that there is an appropriate configuration for celery.

api/desecapi/tests/base.py
```diff
@@ -158,6 +160,7 @@ def __init__(self, test_case, expected_requests, single_expectation_single_reque
         self.single_expectation_single_request = single_expectation_single_request
         self.expect_order = expect_order
         self.old_httpretty_entries = None
+        self.exit_hook = exit_hook or (lambda: None)
```
peterthomassen, line 163: I think `self.exit_hook = exit_hook` and `if callable(self.exit_hook): self.exit_hook()` down below is cleaner.

api/desecapi/tests/test_replication.py
```diff
@@ -32,3 +40,98 @@ def test_serials(self):
...
+        if sha:
+            self.assertEqual(actual_sha, sha, f'Expected HEAD to have hash "{sha}" but had "{actual_sha}".')
+        if message:
+            self.assertIn(
+                message, actual_message,
+                f'Expected "{message}" to appear in the last commit message, but only found "{actual_message}".',
+            )
+
+    def assertHasCommit(self, repo: Repository, commit_id):
+        self.assertNotEqual(
```
peterthomassen, line 67: `assertIsNotNone`

api/desecapi/tests/test_replication.py
```diff
@@ -32,3 +40,98 @@ def test_serials(self):
...
+            repo.get_commit(commit_id)[0], None,
+            f'Expected repository to have commit {commit_id}, but it had not.'
+        )
+
+    def assertHasCommits(self, repo: Repository, commit_id_list):
+        for commit in commit_id_list:
+            self.assertHasCommit(repo, commit)
+
+    def assertHasNotCommit(self, repo: Repository, commit_id):
```
peterthomassen, line 76: `assertIsNone`

docker-compose.yml
```diff
@@ -242,14 +245,17 @@ services:
         tag: "desec/rabbitmq"
     restart: unless-stopped
 
-  celery:
+  celery-email:
     build: api
     image: desec/dedyn-api:latest
     init: true
-    command: celery -A api worker -l info --uid nobody --gid nogroup
+    command: celery worker -A api -E -Q email_slow_lane,email_fast_lane,email_immediate_lane -c 8 -n email -l info --uid nobody --gid nogroup
```
peterthomassen, line 252: `-Q`: wildcard possible? (`email_*`)
`-c`: Why concurrency?
`-n`: Sets the hostname. Why not call it `celery_email` or something?

docker-compose.yml
```diff
@@ -242,14 +245,17 @@ services:
...
     depends_on:
     - dbapi
     - nslord
+    - rabbitmq
+    volumes:
+    - zones:/zones:rw
```
peterthomassen, line 258: Does email celery have to write to zones?

test/e2e2/conftest.py
```diff
@@ -238,17 +251,101 @@ def query(self, qname: str, qtype: str):
...
+            tsprint(f'RPL <<< RR Set {qname}/{qtype} not found')
+            return {}
+        except dns.zone.NoSOA:
+            tsprint(f'RPL <<< Zone {zone} not found')
+            return None
+
+
+@pytest.fixture()
+def replication() -> Replication:
+    return Replication()
```
peterthomassen, line 299: Why does this need to be a fixture? It just serves as a library here, doesn't it?

test/e2e2/conftest.py
```diff
@@ -238,17 +251,101 @@ def query(self, qname: str, qtype: str):
...
+                return
+            except AssertionError as e:
+                if (datetime.now() - started).total_seconds() > timeout:
+                    tsprint(f'{assertion.__code__} eventually failed with {e}, no more retries')
+                    raise e
+
+                time.sleep(wait)
+                wait = min(2 * wait, max_pause)
+
+    return _assert_eventually
```
peterthomassen, line 345: I think this can be encapsulated in something like `return lambda: return_eventually(assertion, ...)`, to avoid the code duplication. `return_eventually` could take an exception class and error message argument to accommodate the difference, or you just catch the `Exception` from `return_eventually` and turn it into an `AssertionError`.

peterthomassen (changes requested):
Very cool stuff!

I have only one major comment, which is: `Repository.remove_history()` is not used anywhere! So this code keeps deleted stuff. Maybe it can be scheduled as a part of `rotate_signatures`? One may also argue that it should be separate.

Also, I think we should implement zone deletion before we deploy this, so that we can observe the behavior most productively.

Finally, in case things go wrong in production, it would be great to have an environment variable which can be used to turn off a) the changes in pdns_change_tracker, b) make the replication tasks no-op.

docker-compose.yml
peterthomassen: In Celery 5, [global options can no longer be positioned after the sub-command](https://docs.celeryproject.org/en/latest/whatsnew-5.0.html#step-1-adjust-your-command-line-invocation), so `-A api` must remain in front of `worker`.

.github/workflows/main.yml
```diff
@@ -63,19 +63,19 @@ jobs:
       if: always()
       run: |
         docker-compose -f docker-compose.yml -f docker-compose.test-e2e.yml ps
-        docker-compose -f docker-compose.yml -f docker-compose.test-e2e.yml logs
+        grep 'desec/' /var/log/syslog
```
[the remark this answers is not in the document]
nils-wisiol: because `docker-compose.test-e2e.yml` does not change log location for all services, and `docker-compose.yml` defines it to be syslog. (On the contrary, `docker-compose.dev.yml` defines log location to be `"json-file"`.) I'll change the remainder of the services having `"json-file"` in `docker-compose.test-e2e.yml` also to log to syslog, so everything is in one place.

api/api/settings.py
```diff
@@ -167,14 +169,32 @@
 CATALOG_ZONE = 'catalog.internal'
 
 # Celery
+# TODO upgrade settings https://docs.celeryproject.org/en/stable/userguide/configuration.html#new-lowercase-settings
 CELERY_BROKER_URL = 'amqp://rabbitmq'
 CELERY_EMAIL_MESSAGE_EXTRA_ATTRIBUTES = []  # required because djcelery_email.utils accesses it
-CELERY_TASK_TIME_LIMIT = 30
-TASK_CONFIG = {  # The first entry is the default queue
+CELERY_TASK_TIME_LIMIT = 300  # as zones become larger, AXFR takes longer, this needs to be increased
+TASK_CONFIG = {  # The first entry is the default queue  # TODO rename in CELERY_TASK_CONFIG
```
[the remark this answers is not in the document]
nils-wisiol: right, corrected

api/api/settings.py
```diff
@@ -167,14 +169,32 @@
...
+        'schedule': crontab(minute=0, hour=0, day_of_week='thursday'),
+        'options': {'priority': 5},  # priority must be higher than rotation jobs for individual domains
+    },
+}
+CELERY_TASK_QUEUES = [
+    Queue(
+        'replication',
+        Exchange('replication'),
+        routing_key='replication',
+        queue_arguments={'x-max-priority': 10},  # make the replication queue a priority-queue
```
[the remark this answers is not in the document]
nils-wisiol: yes - queues are per default non-priority queues

docker-compose.yml
nils-wisiol: No wildcard possible. Concurrency is enabled to account for the i/o bound task of sending emails to use resources efficiently. Note sure where the "hostname" shows up, but it's not the hostname in the unix sense. I haven't seen it outside of Celery context at all, hence I opted for keeping this line as short as possible.

docker-compose.yml
nils-wisiol: Currently not, but it runs on the code base that requires write-access to the zones. I do not see an immediate security problem here, so I opted to keep the write access enabled across all containers running the api image.

test/e2e2/conftest.py
nils-wisiol: fixed

api/api/settings.py
```diff
@@ -167,14 +169,32 @@
 CATALOG_ZONE = 'catalog.internal'
 
 # Celery
+# TODO upgrade settings https://docs.celeryproject.org/en/stable/userguide/configuration.html#new-lowercase-settings
```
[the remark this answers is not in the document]
nils-wisiol: is this still up-to-date as you already upgraded Celery?

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+    def _git(self, *args):
+        cmd = ['/usr/bin/git'] + list(args)
+        print('>>> ' + str(cmd))
+
+        with subprocess.Popen(
+                cmd,
+                bufsize=0,
+                cwd=self.path,
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE,
```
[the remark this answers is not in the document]
nils-wisiol: I think that was done by some auto formatter. If indented by one less tab, it shares the same indent with the compound statement of the `with` block, to distinguish these two the additional indent.

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+        ) as p:
+            rcode = p.wait()
+            stderr = p.stderr.read()
+            stdout = p.stdout.read()
+            try:
+                stderr, stdout = stderr.decode(), stdout.decode()
+            except UnicodeDecodeError:
+                GitRepositoryException('git stdout or stderr was not valid unicode!', cmd=cmd, stderr=stderr, stdout=stdout)
```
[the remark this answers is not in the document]
nils-wisiol: done

api/api/settings.py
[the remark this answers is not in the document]
peterthomassen: no, out of date - I ran this when I did the upgrade, and no changes emanated.

api/Dockerfile
```diff
@@ -20,6 +23,10 @@ ADD ["cronhook/crontab", "cronhook/start-cron.sh", "/root/cronhook/"]
 RUN crontab /root/cronhook/crontab
 RUN chmod +x /root/cronhook/start-cron.sh
 
+# TODO fix permissions on these:
+RUN mkdir /zones /var/run/celerybeat-schedule
+RUN chmod 777 /zones /var/run/celerybeat-schedule
```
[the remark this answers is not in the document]
nils-wisiol: celery runs as `nobody`. What should the permissions be on this?

api/api/settings.py
```diff
@@ -167,14 +169,32 @@
...
 CELERY_BROKER_URL = 'amqp://rabbitmq'
 CELERY_EMAIL_MESSAGE_EXTRA_ATTRIBUTES = []  # required because djcelery_email.utils accesses it
-CELERY_TASK_TIME_LIMIT = 30
-TASK_CONFIG = {  # The first entry is the default queue
+CELERY_TASK_TIME_LIMIT = 300  # as zones become larger, AXFR takes longer, this needs to be increased
```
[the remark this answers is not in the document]
nils-wisiol: When a Celery task fails, I implemented an email notification that will roughly look like this:

image

api/Dockerfile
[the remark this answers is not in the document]
nils-wisiol: adjusted permissions as discussed offline

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+
+        if rcode != 0:
+            raise GitRepositoryException(f'{cmd} returned nonzero error code', cmd=cmd, rcode=rcode, stdout=stdout, stderr=stderr)
+
+        print('\n'.join(['<<< ' + s for s in stdout.split('\n')]))
+        return stdout
+
+    def commit_all(self, msg=None):
+        self._git('add', '.')
+        if self._git('diff', '--numstat', '--staged'):
```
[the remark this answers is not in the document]
nils-wisiol: Changed as proposed.

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+
+        with subprocess.Popen(
+                cmd,
+                bufsize=0,
+                cwd=self.path,
+                stderr=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+        ) as p:
+            rcode = p.wait()
+            stderr = p.stderr.read()
```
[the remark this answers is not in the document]
nils-wisiol: I changed it to raise an exception when `rcode` is zero but `stderr` is non-empty

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+            stdout = p.stdout.read()
+            try:
+                stderr, stdout = stderr.decode(), stdout.decode()
+            except UnicodeDecodeError:
+                GitRepositoryException('git stdout or stderr was not valid unicode!', cmd=cmd, stderr=stderr, stdout=stdout)
+
+        if rcode != 0:
+            raise GitRepositoryException(f'{cmd} returned nonzero error code', cmd=cmd, rcode=rcode, stdout=stdout, stderr=stderr)
+
+        print('\n'.join(['<<< ' + s for s in stdout.split('\n')]))
```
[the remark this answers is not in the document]
nils-wisiol: fixed

api/desecapi/replication.py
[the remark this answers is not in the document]
nils-wisiol: Further changes were necessary: now there are `_git_do`, which insists on zero exit code and empty stderr, returning stdout, and `_git_check`, which returns the rcode regardless of stdout and stderr.

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+    def commit_all(self, msg=None):
+        self._git('add', '.')
+        if self._git('diff', '--numstat', '--staged'):
+            self._git('commit', '-m', msg or 'update')
+
+    def init(self, name='deSEC', email='git@desec.internal'):
+        self._git('init')
+        self._git('config', 'user.name', name)
+        self._git('config', 'user.email', email)
+        # TODO on commit, git tries to access $HOME/.config/git
```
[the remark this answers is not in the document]
nils-wisiol: done

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+        self._git('reflog', 'expire', '--expire=now', '--all')
+        self._git('gc', '--prune=now')
+
+
+class ZoneRepository(Repository):
+    AXFR_SOURCE = '172.16.1.11'
+
+    def __init__(self, path):
+        super().__init__(path)
+        if not os.path.exists(os.path.join(self.path, '.git')):
```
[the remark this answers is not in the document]
nils-wisiol: pathlib is great, but here it seems like an overkill to instantiate so many objects

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+        t = datetime.now()
+        z = dns.zone.from_xfr(xfr)
+        print(f'Parsing took {(datetime.now() - t).total_seconds():.3f}s')
+
+        # TODO sort AXFR? (but take care with SOA)
+        # TODO do not parse AXFR to save time
+
+        # write zone file
+        filename = os.path.join(self.path, name + '.zone')
+        with open(filename + '~', 'w') as f:
```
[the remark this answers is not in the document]
nils-wisiol: This method runs one-at-a-time; if it breaks while writing the ~-file it can (and will) be safely ignored. Subsequent code will overwrite the ~-file. I guess a file lock would achieve the same thing while being more complicated

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+        os.rename(filename + '~', filename)
+
+
+ZONE_REPOSITORY_PATH = '/zones'
+
+
+@shared_task(queue='replication')
+def update(name: str):
+    print(f'updating {name}')
+    zones = ZoneRepository(ZONE_REPOSITORY_PATH)
```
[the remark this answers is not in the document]
nils-wisiol: Actually, `rotate_signatures` queues everything and will do just fine for now

api/desecapi/replication.py
```diff
@@ -0,0 +1,148 @@
...
+    # TODO update DB to indicate successful replication
+
+
+@shared_task(queue='replication', priority=9)
+def rotate_signatures():
+    names = models.Domain.objects.all().values_list('name', flat=True)
+    # TODO restrict rotation to domains that weren't published since last pdns rotation occured
+    print(f'Starting signature rotation for {len(names)} zones.')
+    for name in names:
+        update.s(name).apply_async(priority=1)
```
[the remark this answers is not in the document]
nils-wisiol: done

api/desecapi/tests/base.py
nils-wisiol: done

api/desecapi/tests/test_replication.py
nils-wisiol: fixed

api/desecapi/tests/test_replication.py
nils-wisiol: fixed

test/e2e2/conftest.py
nils-wisiol: done

nils-wisiol:
I addressed all comments, ready for the next round. Let me know if you prefer it squashed!

nils-wisiol:
Actually, I found two things that you said that I didn't address yet:

- [ ] implement zone deletion
- [ ] implement a way to deactivate this

Also, we should look into

- [ ] connect Celery and/or replication jobs to the Prometheus metrics
- [ ] check newly introduced TODOs in the code

peterthomassen:
> I addressed all comments, ready for the next round. Let me know if you prefer it squashed!

Cool! Yes, squashed would be preferred.

nils-wisiol:
squashed and rebased

nils-wisiol:
Most functionality introduced in this PR can be deactivated by stopping `celery-replication` (instead of using, say, an environment variable).

I connected some metrics about the replication to Prometheus, some are stored directly in `dbapi`. Not sure how to tell which ones go where; the main motivation for the current implementation where

- enable lookups on replication state of all domains on `dbapi`:
    - which domains have been successfully replicated?
    - are there domains that haven't been *replicated* after being *published*? (Note that there is a corner-case where this query can miss domains.)
    - Which domains take how long to replicate?
- get a general sense of information in Prometheus:
    - how long do AXFRs take on average?
    - which ones are the longest?
    - how many do we do?

I feel like there are many more interesting metrics, but maybe this PR is already getting too big. My suggestion would be to (if at all) only introduce further replication related metrics in here.

Regarding TODOs in the code, the following are left:

- [ ] upgrade celery settings names: this, strictly speaking, isn't the concern of this PR
- [ ] increase replication performance: if possible at all with pdns, this would make this PR even more involved. Let's first see how performance is in the wild -- on my laptop it's quite fast.
- [ ] AXFR sorting: has potential to save time and bandwidth, but client support needs to be clarified first. So this is blocked by a git zones client implementation. My suggestion here would be to leave it for now.
- [ ] SSHFP record: would be very nice to have, but also complicate things. I guess the `auth` tool does the job for now.

Possible TODOs not mentioned in the code:

- [ ] merge the migration files: a lot of work for little gain
- [ ] more metrics (but see above)

test/e2e2/spec/test_api_rr.py
```diff
@@ -166,15 +165,16 @@ def test_soundness():
...
+    assert NSLordClient.query(f"a.{api_user_domain.domains[0]}", rr_type) == {value}
+    assert_eventually(lambda: query_replication(api_user_domain.domains[0], "a", rr_type) == {value})
 
 
 @pytest.mark.parametrize("rr_type,value", generate_params(VALID_RECORDS_NON_CANONICAL))
-def test_create_valid_non_canonical(api_user_domain: DeSECAPIV1Client, ns_lord: NSClient, rr_type: str, value: str):
+def test_create_valid_non_canonical(api_user_domain: DeSECAPIV1Client, rr_type: str, value: str):
     assert api_user_domain.rr_set_create(api_user_domain.domains[0], rr_type, [value], subname="a").status_code == 201
-    assert len(ns_lord.query(f"a.{api_user_domain.domains[0]}", rr_type)) == 1
+    assert len(NSLordClient.query(f"a.{api_user_domain.domains[0]}", rr_type)) == 1
```
peterthomassen, line 178: Shall we not also call `query_replication()` and check that the record is at least there?

gitzones/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:latest
+
+RUN apk add --no-cache openssh git python3
+
+RUN adduser -D -s /usr/bin/git-shell git \
+  && mkdir /home/git/.ssh \
+  && ln -s /etc/ssh/keys/git_authorized_keys /home/git/.ssh/authorized_keys \
+  && passwd -u git
```
peterthomassen, line 8: The unlocks a locked password. Why is that needed here?

gitzones/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+COPY git-shell-commands /home/git/
+COPY sshd_config /etc/ssh/
+COPY entrypoint.sh /usr/local/bin/
+COPY auth /usr/local/bin/
```
peterthomassen, line 13: `COPY auth entrypoint.sh /usr/local/bin/`

gitzones/README.md
```diff
@@ -0,0 +1,51 @@
+# git Zones Repository SSH Server
+
+Provides *read only* git access to the zones git repository which is stored in volume `zones`.
+
+
+## Server Authentication
+
+The server identity is based on an ED25519 key pair generated on first startup and stored in the `gitzones_keys` volume. 
+To make sure clients are connecting to the correct zone server, use the `knownhost` command of the gitzones container:
```
peterthomassen, line 9: "`knownhost` command"?

gitzones/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+    $ docker-compose exec gitzones auth id                        
+    desec.example.dedyn.io ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIKxVBPSvHDFGzorms9x76+nAo7Zs+0PhaKnMblcdVPos root@c269a29f451d
+
+The output can be appended to any client's `~/.ssh/known_hosts` file.
+
+TODO: Could be replaced with SSHFP record.
```
peterthomassen, line 16: Should not to avoid catch-22.

gitzones/README.md
```diff
@@ -0,0 +1,51 @@
...
+    docker-compose exec gitzones auth rm ns23.desec.io
+
+To list all labels of currently authorized keys,
+
+    docker-compose exec gitzones auth ls
+
+A `-v` flag can be added to also display the keys.
+To see the bare contents of the authorized_keys file,
+
+    docker-compose exec gitzones auth cp
```
peterthomassen, line 43: `cat` instead of `cp`? `cp` suggest "copy", and would usually have a target

gitzones/sshd_config
```diff
@@ -0,0 +1,24 @@
...
+
+# Features
+AllowAgentForwarding no
+AllowTcpForwarding no
+X11Forwarding no
+PermitTTY no
+PrintMotd no
+
+# Logging
+LogLevel VERBOSE
```
peterthomassen, line 16: How much logging does that produce per connection?

gitzones/sshd_config
```diff
@@ -0,0 +1,24 @@
...
+
+# Keys
+HostKey /etc/ssh/keys/ssh_host_ed25519_key
+
+# Mozilla's "Modern" Cipher Settings (https://infosec.mozilla.org/guidelines/openssh)
+KexAlgorithms curve25519-sha256@libssh.org,ecdh-sha2-nistp521,ecdh-sha2-nistp384,ecdh-sha2-nistp256,diffie-hellman-group-exchange-sha256
+Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes192-ctr,aes128-ctr
+MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com
```
peterthomassen, line 24: cool!

gitzones/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine:latest
+
+RUN apk add --no-cache openssh git python3
+
+RUN adduser -D -s /usr/bin/git-shell git \
```
peterthomassen, line 5: sweet

gitzones/auth
```diff
@@ -0,0 +1,167 @@
...
+                f'Could not find authorized key with label "{label}".'
+            )
+        self._save()
+
+    def ls(self) -> Dict[str, str]:
+        """ Returns dictionary of authorized keys, identified by their labels. """
+        return self.db.copy()
+
+    def _save(self) -> None:
+        with open(self.location, "w") as f:
```
peterthomassen, line 67: As this is used in production and logins may be happening while the file is written, better write to a temporary file and then move it into place. This also prevents issues from error conditions occurring during write.

gitzones/auth
```diff
@@ -0,0 +1,167 @@
...
+    with open(host_public_key_file, "r") as f:
+        print(f'desec.{os.environ["DESECSTACK_DOMAIN"]} {f.readline().strip()}')
+
+
+def cmd_add(args: List[str]) -> None:
+    """ Entrypoint for CLI command that adds an authorized SSH key. """
+    parser = argparse.ArgumentParser()
+    parser.add_argument("proto", help="Protocol used with given key.")
+    parser.add_argument("key", help="Public key of authorized key pair.")
+    parser.add_argument("label", help="Label under which the key is stored.", nargs="+")
```
peterthomassen, line 112: Can't `print_usage` be derived from this? The global `usage` variable seems redundant (in terms of content). Am I missing something?

gitzones/auth
```diff
@@ -0,0 +1,167 @@
...
+    parsed = parser.parse_args(args)
+    AuthorizedKeys().add(parsed.proto, parsed.key, " ".join(parsed.label))
+
+
+def cmd_rm(args: List[str]) -> None:
+    """ Entrypoint for CLI command that removes an authorized SSH key. """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "label", help="All keys with this label will be removed.", nargs="+"
```
peterthomassen, line 121: Isn't the label unique?

gitzones/auth
```diff
@@ -0,0 +1,167 @@
...
+    )
+    parsed = parser.parse_args(args)
+    AuthorizedKeys().rm(" ".join(parsed.label))
```
peterthomassen, line 124: Why the `join`?

gitzones/auth
```diff
@@ -0,0 +1,167 @@
...
+
+
+def cmd_ls(args: List[str]) -> None:
+    """ Entrypoint for CLI command that shows all authorized SSH keys. """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        action="store_true",
+        help="Key with this label will be removed.",
```
peterthomassen, line 134: wrong text

api/desecapi/metrics.py
```diff
@@ -1,54 +1,43 @@
 from prometheus_client import Counter, Histogram, Summary
 
-metrics = {}
 
+class Model:
+    captcha_content_created = Counter('desecapi_captcha_content_created', 'number of times captcha content created',['kind'])
+    autodelegation_created = Counter('desecapi_autodelegation_created', 'number of autodelegations added')
+    autodelegation_deleted = Counter('autodelegation_deleted', 'number of autodelegations deleted')
```
peterthomassen, line 7: Counter name is missing `desecapi_` prefix.

The previous structure had the advantage of not duplicating the name (i.e. using the same thing on the lhs and rhs). I had made that choice a) to ensure consistency, and b) that you can infer the metric name _as it is exposed in Prometheus_ from the code line where you increment the counter. That connection is not enforced in the new structure.

Before I double-check all entries for completeness and spelling: Have you done the refactoring / reformatting in an automated fashion, or should I be in the "proofread a human" mindset?

peterthomassen (changes requested):
Very nice work, and in particular a lot of work! :pray: 

I have a bunch of inline comments. Regarding the other todos:
- celery settings names are up to date,
- we should try replication performance in practice before putting an effort into optimizing it further,
- AXFR sorting can happen at a later stage, I'd say,
- SSHFP creates a catch-22, and I'm not in favor of it,
- migration files should be squashed (isn't it just copying a few lines from the second to the first, then deleting the second? -- also, I made the suggestion to change one of the field names, in which case they'd have to be touched anyways)
- more metrics: sure, why not! :-) but I think we're ok.

Thanks again! Looking forwarding to experimenting with this for real.

api/desecapi/models.py
```diff
@@ -219,6 +219,8 @@ class RenewalState(models.IntegerChoices):
                             validators=validate_domain_name)
     owner = models.ForeignKey(User, on_delete=models.PROTECT, related_name='domains')
     published = models.DateTimeField(null=True, blank=True)
+    replicated = models.DateTimeField(null=True, blank=True)
+    replication_time = models.DurationField(null=True, blank=True)
```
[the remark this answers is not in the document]
nils-wisiol: Renamed. Should be nullable, otherwise an analysis of average replication time will be biased.

api/api/settings.py
```diff
@@ -167,14 +169,37 @@
...
 TASK_CONFIG = {  # The first entry is the default queue
     'email_slow_lane': {'rate_limit': '3/m'},
     'email_fast_lane': {'rate_limit': '1/s'},
     'email_immediate_lane': {'rate_limit': None},
 }
+CELERY_BEAT_SCHEDULE = {
+    'rotate_signatures': {
+        'task': 'desecapi.replication.rotate_signatures',
+        'schedule': crontab(minute=0, hour=0, day_of_week='thursday'),
```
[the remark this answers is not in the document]
nils-wisiol: Good catch - I set it to UTC

api/desecapi/pdns_change_tracker.py
```diff
@@ -268,6 +270,8 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
         self.transaction.__exit__(None, None, None)
 
+        for name in replication_required:
+            replication.update.delay(name)
         for name in axfr_required:
             _pdns_put(NSMASTER, '/zones/%s/axfr-retrieve' % pdns_id(name))
         Domain.objects.filter(name__in=axfr_required).update(published=timezone.now())
```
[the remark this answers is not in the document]
nils-wisiol: I agree in principle, but `NSMASTER` is planned to retire if this works, so let's not waste too much effort on covering the corner cases of `NSMASTER` and git zones replication mechanism interactions.

api/desecapi/replication.py
```diff
@@ -0,0 +1,192 @@
...
+
+        return stdout
+
+    def _git_check(self, *args):
+        _, rcode, _, _ = self._git(*args)
+        return rcode
+
+    def commit_all(self, msg=None):
+        self._git_do('add', '.')
+        if self._git_do('diff', '--numstat', '--staged'):
```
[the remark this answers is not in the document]
nils-wisiol: that one slipped through :)

api/desecapi/replication.py
```diff
@@ -0,0 +1,192 @@
...
+        self._git_do('init')
+
+    def get_head(self):
+        return self.get_commit('HEAD')
+
+    def get_commit(self, rev):
+        try:
+            commit_msg = self._git_do('show', rev, '--format=%s', '-s')[:-1]
+            commit_hash = self._git_do('show', rev, '--format=%H', '-s')[:-1]
+            return commit_msg, commit_hash
```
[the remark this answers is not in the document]
nils-wisiol: Done

api/desecapi/replication.py
```diff
@@ -0,0 +1,192 @@
...
+        except GitRepositoryException as e:
+            return None, None
+
+    def remove_history(self, before: datetime):
+        rev = self._git_do('log', f'--before={before.isoformat()}Z', '-1', '--format=%H')
+        with open(os.path.join(self.path, '.git', 'shallow'), 'w') as f:
+            f.writelines([rev])
```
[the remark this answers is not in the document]
nils-wisiol: fixed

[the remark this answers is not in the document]
nils-wisiol: wait that was to include a line break at the end

api/desecapi/replication.py
```diff
@@ -0,0 +1,192 @@
...
+    print(f'updating {name}')
+    t = timezone.now()
+    zones = ZoneRepository(ZONE_REPOSITORY_PATH)
+    zones.refresh(name)
+    zones.commit_all(f'Update for {name}')
+    models.Domain.objects.filter(name=name).update(replicated=timezone.now(), replication_time=timezone.now() - t)
+
+
+@shared_task(queue='replication', priority=9)
+def rotate_signatures():
```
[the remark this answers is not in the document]
nils-wisiol: `update_all()`, analogous to the `update(zone)` method above

api/desecapi/replication.py
```diff
@@ -0,0 +1,192 @@
...
+    names = models.Domain.objects.all().values_list('name', flat=True)
+    print(f'Starting signature rotation for {len(names)} zones.')
```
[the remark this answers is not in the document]
nils-wisiol: done

api/desecapi/tests/base.py
```diff
@@ -467,18 +474,24 @@ def request_pdns_update_catalog(cls):
             'priority': 1,  # avoid collision with DELETE zones/(?P<id>[^/]+)$ (httpretty does not match the method)
         }
 
-    def assertPdnsRequests(self, *expected_requests, expect_order=True):
+    def __init__(self, methodName: str = ...) -> None:
+        super().__init__(methodName)
+        self._mock_replication = None
+
```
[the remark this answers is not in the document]
nils-wisiol: that would change that to be [class attribute instead of an instance attribute](https://dzone.com/articles/python-class-attributes-vs-instance-attributes). I want it to be an instance attribute

api/desecapi/tests/test_replication.py
```diff
@@ -32,3 +40,97 @@ def test_serials(self):
...
+            repo.get_commit(commit_id)[0], f'Expected repository to have commit {commit_id}, but it had not.'
+        )
+
+    def assertHasCommits(self, repo: Repository, commit_id_list):
+        for commit in commit_id_list:
+            self.assertHasCommit(repo, commit)
+
+    def assertHasNotCommit(self, repo: Repository, commit_id):
+        self.assertIsNone(
+            repo.get_commit(commit_id)[1], f'Expected repository to not have commit {commit_id}, but it had.'
```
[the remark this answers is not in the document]
nils-wisiol: I dont think so, as the asserts all raise directly. So a "not" implementation would require assert that this exception was raised. Also, Django provides a number of "complementary" assertions, such as `self.assertEqual`, `self.assertNotEqual`, suggesteding that there is no general negation implementation. :-(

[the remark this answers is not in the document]
nils-wisiol: I did align the implementations though.

docker-compose.yml
```diff
@@ -242,14 +245,17 @@ services:
         tag: "desec/rabbitmq"
     restart: unless-stopped
 
-  celery:
+  celery-email:
     build: api
     image: desec/dedyn-api:latest
     init: true
-    command: celery -A api worker -l info --uid nobody --gid nogroup
+    command: celery -A api worker -E -Q email_slow_lane,email_fast_lane,email_immediate_lane -c 8 -n email -l info --uid nobody --gid nogroup
```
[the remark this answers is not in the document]
nils-wisiol: removed `-E`. As for concurrency, we have i/o-bound tasks, which we can handle many in parallel, even if we have few cores. Fixing the number of threads to equal the number of cores makes sense when you got cpu-bound tasks that all eat up everything one core can provide

test/e2e2/conftest.py
```diff
@@ -82,21 +87,26 @@ def _filter_response_output(output: dict) -> dict:
         return output
 
     def _do_request(self, *args, **kwargs):
-        verify_list = [self.verify] + self.verify_alt
+        # do not verify SSL if we're in faketime
+        verify_list = [self.verify] + self.verify_alt if faketime_get() == '+0d' else [False]
```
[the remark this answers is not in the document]
nils-wisiol: yes, the `if` binds very weakly. Parenthesis trigger my code style checker, and having unnecessary warnings around makes me overlook warnings that are important. So I changed it to a regular `if` with compound statement.

test/e2e2/spec/test_api_rr.py
nils-wisiol: I rebased much of those tests onto master, does this comment still apply? Where is the new location in the rebased code?

gitzones/Dockerfile
nils-wisiol: Otherwise SSH refuses login

gitzones/Dockerfile
nils-wisiol: done

gitzones/README.md
nils-wisiol: I meant `auth`. fixed

gitzones/README.md
nils-wisiol: removed

gitzones/README.md
nils-wisiol: done

gitzones/sshd_config
nils-wisiol: A lot, but it's also the least verbose level that logs the key fingerprint used for log in. Hence it's recommended by Mozilla. I guess we don't need that? It's not like we do not expect our authorized keys to *not* login.

nils-wisiol: Set to `INFO`, which is also the default

gitzones/auth
nils-wisiol: done

gitzones/auth
nils-wisiol: this code doesn't run unless the command has been recognized; and only one command at a time will be accepted. So there is no way to list all details for all commands at the same time. I'm sure there is some very elaborate elegant solution, though :)

gitzones/auth
nils-wisiol: yup, fixed

gitzones/auth
nils-wisiol: so that labels can contain spaces

gitzones/auth
nils-wisiol: fixed

api/desecapi/metrics.py
nils-wisiol: oh I see. Well let's exclude this commit then for now.

api/api/settings.py
```diff
@@ -167,14 +169,37 @@
 CATALOG_ZONE = 'catalog.internal'
 
 # Celery
+# TODO upgrade settings https://docs.celeryproject.org/en/stable/userguide/configuration.html#new-lowercase-settings
```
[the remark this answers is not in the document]
nils-wisiol: But the page behind the link says that they won't accept uppercase settings anymore. Our's are all uppercase?

api/api/celery.py
```diff
@@ -12,3 +16,24 @@
 @app.task(bind=True)
 def debug_task(self):
     print('Request: {0!r}'.format(self.request))
+
+
+@task_failure.connect()
+def task_failure(task_id, exception, args, kwargs, traceback, einfo, **other_kwargs):
+    context = {
```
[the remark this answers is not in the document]
nils-wisiol: Done, will now look like this

image

Secrets are redacted

test/e2e2/spec/test_api_rr.py
nils-wisiol: Also, unrelated, is here an 'eventually' missing?

nils-wisiol:
Your turn @peterthomassen 

Could you resolve all the conversations that you think are resolved? There are still quite many open right now :)

nils-wisiol:
@peterthomassen could you also take a look at the failing e2e2 tests? It's a problem with the `DNSKEY` test (and friends) in `test_create_valid_canonical` likely introduced when I rebased it.

api/api/celery.py
```diff
@@ -6,34 +7,53 @@
...
-        'sender': str(other_kwargs.get('sender', '')),
-    }
-    subject = get_template(f'emails/exception-celery/subject.txt').render(context).strip()
-    print(subject, task_id, exception, args, kwargs, traceback, einfo, other_kwargs)
-    mail_admins(
-        subject=subject,
-        message=get_template('emails/exception-celery/content.txt').render(context),
-        connection=get_connection('django.core.mail.backends.smtp.EmailBackend'),
+    try:
+        sender = other_kwargs.get('sender', '').name
```
peterthomassen, line 33: `, ''` is unnecessary

api/desecapi/migrations/0012_domain_replicated.py
```diff
@@ -1,18 +0,0 @@
-# Generated by Django 3.1.4 on 2021-01-02 15:24
```
peterthomassen: belongs to 2f0dd672e5174ffafaa028fbf4a0013914907a70, but I guess we're going to squash that anyways

api/desecapi/templates/emails/exception-celery/content.txt
```diff
@@ -1,15 +0,0 @@
-{% autoescape off %}A Celery task raised a {{ exception_type }} exception.
```
peterthomassen: belongs to 729c4c493dbd38d9d3fa5b9cddc92761dc71416a, but I guess we're going to squash that anyways

api/api/settings.py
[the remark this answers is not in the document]
peterthomassen: That's if you configure Celery directly. If you configure it through Django, [things remain uppercase](https://docs.celeryproject.org/en/stable/history/whatsnew-4.0.html#latentcall-django-admonition):

> If you’re loading Celery configuration from the Django settings module then you’ll want to keep using the uppercase names.
> 
> You also want to use a `CELERY_` prefix so that no Celery settings collide with Django settings used by other apps.

Although it's a bit hard to find in the docs, it's pretty clear. Maybe we should replace the TODO comment with a comment linking to that section specifically.

api/api/settings.py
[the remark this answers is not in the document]
peterthomassen: Cool, although [that appears to be the default](https://docs.celeryproject.org/en/stable/userguide/configuration.html#std-setting-timezone). But better explicit and all ...

api/desecapi/pdns_change_tracker.py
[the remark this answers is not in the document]
peterthomassen: Sure. I'll leave this conversation open for later reference (we don't know yet if and for how long we'll be in dual-stack operation).

api/desecapi/replication.py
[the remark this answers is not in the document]
peterthomassen: The line break is not needed. :-)

api/desecapi/tests/base.py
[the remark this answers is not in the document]
peterthomassen: The following is an "academic assessment" because this attribute stuff is interesting. I have not stakes about how it's done.

The article you linked has a section called "Class Attributes Mutate to Be Instance Attributes" which says that class attributes become instance attributes when assigned inside a method. (Technically, the attribute is added to the instance attribute dict which shadows the class attribute dict.)

So, when you assign to the attribute in `setUp`, you have an instance attribute -- even when initializing it as a class attribute. Thus, your goal is achieved. :)

Now, I don't know if there's a convention if "placeholder variables" are better declared in `__init__` or on the class level. You certainly could do without declaring it at all, as the assignment in `setUp` serves just as well as the one in `__init__`. Having said that, another option would be adding an [instance variable annotation without default](https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations) (`_mock_replication: None`).

docker-compose.yml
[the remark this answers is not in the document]
peterthomassen: I am not following: We have 8 cores in production, and you're fixing it to 8 child processes. If we did not do that, we'd have 16 child processes in case we upgrade to a machine with 16 cores, while with the current implementation, we'll stick to 8 child processes. Given that that we're i/o-bound, why should we fix it at such low number? (I would understand if you had fixed it at a higher number than we have cores -- but fixing it at exactly that number currently does nothing, but may limit us in the future.)

test/e2e2/spec/test_api_rr.py
peterthomassen: Regarding your first question: At the end of `test_create_valid_canonical`, [we check that replication worked](https://github.com/desec-io/desec-stack/blob/253837b/test/e2e2/spec/test_api_rr.py#L362), but we don't do that at the end of `test_create_valid_non_canonical`. It looks like you found the code location already.

Regarding the "eventually": I don't think so -- we're querying nslord, which we provision with the RRset in a _synchronous_ pdns API request two lines further up. As soon as that returns, the RRset should be queryable, and not `_eventually` should be needed.

gitzones/Dockerfile
peterthomassen: That's weird. On my machines, I always lock my root accounts using `passwd -l root`, and can still do key-based login ...

gitzones/auth
peterthomassen: Ah, I had missed the `nargs="+"`. I guess that's one way. On the other hand, wouldn't we expect the user to enclose labels in quotes `"` if they contain whitespace (just like for any other shell argument)?

I'm pointing this out because it's a deviation from how shell arguments with whitespace are usually handled. Maybe I'm getting something wrong here ...

peterthomassen (commented):
I marked most conversations as resolved. It looks like we're taking the last turns on this PR's road!

I think it would be great to drop 3844314b31df1bbd3ffbf0548b80f961edc87304, squash the other improvements, and once more go over the conversations that have remained open. None of them are substantial though, so all good. Meanwhile, I'll take a look at the failing e2e2 tests.

peterthomassen:
Did I observe correctly that `get_commit` and `get_head` in `desecapi.replication` are only used from the tests? In that case, shouldn't we move these functions somewhere into `tests/`?

api/api/settings.py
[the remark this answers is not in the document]
nils-wisiol: done

api/api/settings.py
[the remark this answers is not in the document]
nils-wisiol: I read somewhere that Django timezone may be used, but I can't find that anymore. Anyways.

api/desecapi/tests/base.py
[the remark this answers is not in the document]
nils-wisiol: Um, the code style I followed thus far is to declare instance attributes during `__init__`, so I prefer to keep it this way. Gotta read up on instance vs class attributes though ...

docker-compose.yml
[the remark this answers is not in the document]
nils-wisiol: You're right, 8 is essentially a function of "how much traffic do I except" and not of "how many cores do we have". If we had only one core, I'd still use 8 (or 16 or whatever). So declaring the attribute was to make this code meaningful for deployments with few cores (or the result of my ignorance as to how many cores we have). :blush:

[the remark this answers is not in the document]
nils-wisiol: Closing this as if we ever send more emails than 8 processes can handle, we need to revisit email sending anyways.

gitzones/auth
nils-wisiol: nah you're right, I just thought it's a cheap way to do it. I can remove that?

test/e2e2/spec/test_api_rr.py
nils-wisiol: fixed by @peterthomassen in 137c2b904cc72092e8bd17ab3e2be7e1f8cfa788

gitzones/auth
peterthomassen: I think you can -- what do you mean?

api/api/celery.py
```diff
@@ -6,34 +7,53 @@
...
+            'exception': exception,
+            'task_args': args,
+            'task_kwargs': pprint.pformat(kwargs),
+            'sender': sender,
+        },
+        exc_info=einfo,
     )
+
+
+django.setup()
```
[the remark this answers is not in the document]
nils-wisiol: nope, `import django.utils.log` implies `import django`

api/api/celery.py
nils-wisiol: fixed

gitzones/auth
nils-wisiol: I removed it, now it behaves like a shell application should

gitzones/Dockerfile
nils-wisiol: I confirmed that password-based login is impossible for user git even after passwd -u

gitzones/auth
```diff
@@ -64,10 +64,12 @@ class AuthorizedKeys(list):
         return self.db.copy()
 
     def _save(self) -> None:
-        with open(self.location + '~', "w") as f:
+        real_location = os.path.realpath(self.location)
+        with open(real_location + '~', "w") as f:
             for label, proto_key in self.db.items():
                 f.write(proto_key + " " + label + "\n")
-        os.rename(self.location + '~', self.location)
+        os.rename(real_location + '~', os.path.realpath(real_location))
```
peterthomassen, line 71: `os.path.realpath(real_location)` can be simply `real_location`, I think

nils-wisiol:
I rebased this, adjusted webapp development settings to my new docker images (maybe you need do build with a fresh pull), and then:

- reduced the number of rr sets in the replication performance test
- fixed a bug where the authorized keys of the gitzones server were not stored in a docker volume
- changed gitzones to run sshd as git (instead of root)

Last remaining TODO that should be resolved by this PR is to include some more metrics. @peterthomassen could you do that?

peterthomassen:
> Last remaining TODO that should be resolved by this PR is to include some more metrics. @peterthomassen could you do that?

Sure, shall do tomorrow!

gitzones/auth
nils-wisiol: fixed

nils-wisiol:
I prepared the code for squashing, leaving my changes since last review as fixups

peterthomassen (approved):

PR #455, opened 2020-09-29
20200929 email citext

peterthomassen:
supersedes #448 

Before deploying this, we need to check and clean up any accounts that would be considered duplicates.


nils-wisiol (approved):
noice

PR #456, opened 2020-10-02
20200916 token policy domains

peterthomassen:


peterthomassen:

peterthomassen:
I added a commit with version bumps.

We currently cannot upgrade to Django 4.0 as django-prometheus has to catch up. According to upstream, that will take a few days. -- That aside, I checked that no changes to our code base are expected to be necessary when upgrading to Django 4.

peterthomassen:
Docs are missing; happy to add them after (first round of) technical review.

Also, the following occurred to me: When the owner attribute of a domain is changed, the old user may have a token with a policy that references the domain. What should be done here? (We can enforce consistency with an exclusion constraint on either side (at the policy or at the domain), I think, but it would span three levels of foreign-key relations ...)

api/desecapi/views.py
```diff
@@ -26,7 +27,10 @@
 from desecapi.exceptions import ConcurrencyException
 from desecapi.pdns import get_serials
 from desecapi.pdns_change_tracker import PDNSChangeTracker
-from desecapi.permissions import ManageTokensPermission, IsDomainOwner, IsOwner, IsVPNClient, WithinDomainLimitOnPOST
+from desecapi.permissions import (
+    IsDomainOwner, IsOwner, IsVPNClient, HasViewTokenPermission, ManageTokensPermission,
+    TokenHasDomainDynDNSPermission, TokenHasDomainRRsetsPermission, TokenNoDomainPolicy, WithinDomainLimit,
+)
```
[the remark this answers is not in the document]
nils-wisiol: Also, can the naming be made more consistent? Most of the older permissions have some name beginning with a verb ("Is..", "Has...")

api/desecapi/views.py
```diff
@@ -150,6 +156,43 @@ def perform_destroy(self, instance: models.Domain):
...
+            pass
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_queryset(self):
+        return models.TokenDomainPolicy.objects.filter(token_id=self.kwargs['token_id'], token__user=self.request.user)
+
+    def perform_destroy(self, instance):
+        try:
+            super().perform_destroy(instance)
+        except InternalError:
```
nils-wisiol, line 191: exception clause too broad?

nils-wisiol (commented):
First round of comments. Gotta continue reviewing after clarifying the meaning of the permission classes. Havn't looked at the tests at all yet.

api/desecapi/permissions.py
```diff
@@ -21,6 +21,55 @@ def has_object_permission(self, request, view, obj):
...
+
+
+class TokenHasDomainRRsetsPermission(TokenHasDomainBasePermission):
+    """
+    Custom permission to check whether a token authorizes accessing RRsets for the view domain.
+    """
+    perm_field = 'perm_rrsets'
+
+
+class HasViewTokenPermission(permissions.BasePermission):
```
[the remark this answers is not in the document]
peterthomassen: rename to `IsPresentToken` or similar

api/desecapi/serializers.py
```diff
@@ -75,6 +76,30 @@ def get_fields(self):
...
+        fields = ('domain', 'perm_dyndns', 'perm_rrsets',)
+
+    def to_internal_value(self, data):
+        return {**super().to_internal_value(data),
+                'token': self.context['request'].user.token_set.get(id=self.context['view'].kwargs['token_id'])}
+
+    def save(self, **kwargs):
+        try:
+            return super().save(**kwargs)
+        except InternalError:
```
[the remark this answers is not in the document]
peterthomassen: add explicit checks

api/desecapi/views.py
peterthomassen: add explicit checks

api/desecapi/models.py
```diff
@@ -457,6 +456,72 @@ def generate_key(self):
     def make_hash(plain):
         return make_password(plain, salt='static', hasher='pbkdf2_sha256_iter1')
 
+    #@cache  <-- Do we need this? Interferes with tests that change permissions quickly
```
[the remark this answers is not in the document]
peterthomassen: removed

api/desecapi/models.py
```diff
@@ -457,6 +456,72 @@ def generate_key(self):
...
+    def get_policy(self, *, domain=None):
```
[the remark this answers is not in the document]
peterthomassen: As discussed in person, we're not changing the "no-policy behavior" now, as it would be a breaking change.

api/desecapi/models.py
```diff
@@ -457,6 +456,72 @@ def generate_key(self):
...
+        # Perhaps this will change with https://code.djangoproject.com/ticket/21961
+        self.tokendomainpolicy_set.filter(domain__isnull=False).delete()
+        self.tokendomainpolicy_set.filter(domain__isnull=True).delete()
+        return super().delete()
+
+
+# We ensure that if there is *any* domain policy for a token, there is always one with domain=None.
+# Thus, the case where there exists a domain-specific policy, but no default, does not exist.
+# If during permission checking, there is neither a domain-specific policy nor a default policy for
+# the given token, this means that the token has no policy at all, and no restrictions are applied.
```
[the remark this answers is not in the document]
peterthomassen: As discussed in person, we're not changing the "no-policy behavior" now, as it would be a breaking change.

api/desecapi/permissions.py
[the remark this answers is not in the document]
peterthomassen: added class doc, kept name. (Other names I could come up with were even more inconsistent with existing naming.)

api/desecapi/serializers.py
```diff
@@ -75,6 +76,30 @@ def get_fields(self):
         return fields
 
 
+class DomainSlugRelatedField(serializers.SlugRelatedField):
```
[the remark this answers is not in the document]
peterthomassen: yes, can be used for any related Domain field that is identified by name

api/desecapi/serializers.py
```diff
@@ -75,6 +76,30 @@ def get_fields(self):
...
+
+    def get_queryset(self):
+        return self.context['request'].user.domains
+
```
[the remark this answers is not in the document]
peterthomassen: done

api/desecapi/serializers.py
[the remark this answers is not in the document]
peterthomassen: done

api/desecapi/urls/version_1.py
```diff
@@ -18,6 +21,7 @@
 
     # Token management
     path('tokens/', include(tokens_router.urls)),
+    path('tokens/<token_id>/domain_policies/', include(tokendomainpolicies_router.urls)),
```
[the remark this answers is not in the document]
peterthomassen: I don't know. Perhaps role policies, at some point?

api/desecapi/views.py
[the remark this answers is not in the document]
peterthomassen: `import` fixed. I could not come up with a more consistent naming convention with sufficient clarity ...

api/desecapi/views.py
```diff
@@ -67,31 +67,6 @@ def destroy(self, request, *args, **kwargs):
         return Response(status=status.HTTP_204_NO_CONTENT)
 
 
-class DomainViewMixin:
```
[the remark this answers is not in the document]
peterthomassen: as discussed in person, we'll look at this again once it comes up

api/desecapi/views.py
peterthomassen: done

api/desecapi/views.py
```diff
@@ -233,7 +231,7 @@ def get_queryset(self):
 
                 rrsets = rrsets.filter(**{'%s__exact' % filter_field: value})
 
-        return rrsets
+        return rrsets.all()  # without .all(), cache is sometimes inconsistent with actual state in bulk tests. (Why?)
```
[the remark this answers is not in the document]
peterthomassen: if I remember correctly, it's unrelated

peterthomassen:
e2e2 tests are currently broken. However, this is not due to this PR: On my machine, they are now also broken for the current main branch. Both for the main branch (locally) and for this PR (on GitHub), the full logs reveal that during the start of `celery-email` and `celery-replication`, there are errors like the following:

```
rabbitmq_1                | 2021-12-19 22:14:45.691 [info] <0.949.0> accepting AMQP connection <0.949.0> (172.16.6.4:34796 -> 172.16.6.2:5672)
rabbitmq_1                | 2021-12-19 22:14:45.696 [info] <0.949.0> connection <0.949.0> (172.16.6.4:34796 -> 172.16.6.2:5672): user 'guest' authenticated and granted access to vhost '/'
celery-replication_1      | [2021-12-19 22:14:46,697: INFO/MainProcess] mingle: all alone
rabbitmq_1                | 2021-12-19 22:14:46.699 [error] <0.880.0> Error on AMQP connection <0.880.0> (172.16.6.5:58252 -> 172.16.6.2:5672, vhost: '/', user: 'guest', state: running), channel 1:
rabbitmq_1                |  operation channel.open caused a connection exception channel_error: "second 'channel.open' seen"
rabbitmq_1                | 2021-12-19 22:14:46.700 [info] <0.880.0> closing AMQP connection <0.880.0> (172.16.6.5:58252 -> 172.16.6.2:5672, vhost: '/', user: 'guest')
rabbitmq_1                | 2021-12-19 22:14:46.702 [info] <0.963.0> Closing all channels from connection '172.16.6.5:58252 -> 172.16.6.2:5672' because it has been closed
celery-replication_1      | [2021-12-19 22:14:46,701: WARNING/MainProcess] consumer: Connection to broker lost. Trying to re-establish the connection...
celery-replication_1      | Traceback (most recent call last):
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/connection.py", line 510, in channel
celery-replication_1      |     return self.channels[channel_id]
celery-replication_1      | KeyError: None
celery-replication_1      | 
celery-replication_1      | During handling of the above exception, another exception occurred:
celery-replication_1      | 
celery-replication_1      | Traceback (most recent call last):
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/celery/worker/consumer/consumer.py", line 311, in start
celery-replication_1      |     blueprint.start(self)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/celery/bootsteps.py", line 116, in start
celery-replication_1      |     step.start(parent)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/celery/worker/consumer/gossip.py", line 107, in start
celery-replication_1      |     super().start(c)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/celery/bootsteps.py", line 396, in start
celery-replication_1      |     channel = c.connection.channel()
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/kombu/connection.py", line 282, in channel
celery-replication_1      |     chan = self.transport.create_channel(self.connection)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/kombu/transport/pyamqp.py", line 166, in create_channel
celery-replication_1      |     return connection.channel()
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/connection.py", line 513, in channel
celery-replication_1      |     channel.open()
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/channel.py", line 436, in open
celery-replication_1      |     return self.send_method(
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/abstract_channel.py", line 66, in send_method
celery-replication_1      |     return self.wait(wait, returns_tuple=returns_tuple)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/abstract_channel.py", line 86, in wait
celery-replication_1      |     self.connection.drain_events(timeout=timeout)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/connection.py", line 521, in drain_events
celery-replication_1      |     while not self.blocking_read(timeout):
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/connection.py", line 527, in blocking_read
celery-replication_1      |     return self.on_inbound_frame(frame)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/method_framing.py", line 53, in on_frame
celery-replication_1      |     callback(channel, method_sig, buf, None)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/connection.py", line 533, in on_inbound_method
celery-replication_1      |     return self.channels[channel_id].dispatch_method(
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/abstract_channel.py", line 143, in dispatch_method
celery-replication_1      |     listener(*args)
celery-replication_1      |   File "/usr/local/lib/python3.9/site-packages/amqp/connection.py", line 663, in _on_close
celery-replication_1      |     raise error_for_code(reply_code, reply_text,
celery-replication_1      | amqp.exceptions.ChannelNotOpen: Channel.open: (504) CHANNEL_ERROR - second 'channel.open' seen
```

Note that Celery's exception just repeats the earlier error message from RabbitMQ, line 5.

I don't know yet what the reason for the problem is (but the version bumps aren't, as it also happens on main). However, I wanted to share it, because it's conceivable that the problem also would occur in production (main branch) if a restart was necessary. We should thus debug this quickly.

nils-wisiol:
I believe this is symptom of a larger issue: because we do not fix the minor versions, we rely on our dependencies to not break compatibility when upgrading minor versions. Build and tests are not reproducible even for the same commits if minor dependency versions are upgraded.

I'll verify if this is the root cause of the current incident by digging out the last dependency versions for which main built and see if it still does.

If the hypothesis holds up, my suggestion to fix this would be to introduce fully defined dependency versions rather than leave minor versions to be upgraded automatically and without our control.

nils-wisiol:
I've confirmed that this error did not occur on my machine with an 11-months-old rabbitmq version and also not with the newest version of rabbitmq docker image. The cause seems to be the Python amqp package. My investigation shows:

- amqp 5.0.6 working (latest main GitHub action)
- amqp 5.0.8 broken (GitHub action mentioned above)
- amqp 5.0.9 working (latest GitHub action on this PR)

This confirms hypothesis above, we should consider pinning minor versions.

api/desecapi/urls/version_1.py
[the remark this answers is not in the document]
nils-wisiol: `tokens/<token_id>/policies/domain`?

api/desecapi/views.py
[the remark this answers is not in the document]
nils-wisiol: I don't fully understand yet what's going on here. Should this be in a separate commit? Is this to be considered a fix or a cleanup?

api/desecapi/urls/version_1.py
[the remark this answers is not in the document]
peterthomassen: Why is that better?

I am not opposed, and I like the idea semantically because it is more extensible. But, it also kind of implies that something non-404 will happen if a request is sent to `policies/`. What would that be?

Reminds me of DNS empty non-terminals ...

api/desecapi/models.py
```diff
@@ -522,6 +512,27 @@ class Meta:
             models.UniqueConstraint(fields=['token'], condition=Q(domain__isnull=True), name='unique_entry_null_domain')
         ]
 
+    def clean(self):
+        default_policy = self.token.get_policy(domain=None)
+        if self.pk:  # update
+            # Can't change policy's default status ("domain NULLness") to maintain policy precedence
+            if (self.domain is None) != (self.pk == default_policy.pk):
+                raise ValidationError({'domain': 'Policy precedence: A default policy must exist before others.'})
```
nils-wisiol, line 532: error message needs to be adjusted

api/desecapi/permissions.py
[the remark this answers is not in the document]
nils-wisiol: Inspire by class doc: `AuthTokenCorrespondsToViewToken`

This may be legible when looking at the `permission_classes` list in views.py

api/desecapi/permissions.py
```diff
@@ -82,6 +91,9 @@ def has_permission(self, request, view):
 
 
 class ManageTokensPermission(permissions.BasePermission):
```
nils-wisiol, line 93: `HasManageTokensPermission`?

api/desecapi/views.py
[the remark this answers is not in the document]
peterthomassen: This line is part of commit 098fc825ffcddd9dce643cea9d7a213afd90f259 and as such belongs to the refactoring to reduce code duplication. The refactoring makes both `RRsetDetail` and `RRsetList` inherit from `RRsetView`, and the child views' access to the querysets thus makes use of `RRsetView.get_queryset()` (via `super()`). Before the change, both classes constructed fully independent querysets.

When using `RRsetView.get_queryset()` and not changing the above line, 8 of the 33 `test_bulk_*` tests fail with unexpected test outcome (wrong pdns requests etc.). This suggests that some caching is off. The problem is resolved by adding `.all()` to force evaluation of the queryset.

I don't fully understand which conditions lead to the problem, but it's apparent that the solution is to force evaluation at the time when the rows are needed.

As this is a consequence of the refactoring, it is in the correct commit.

api/desecapi/models.py
peterthomassen: I think it is accurate; please suggest a specific message

api/desecapi/permissions.py
peterthomassen: done

api/desecapi/urls/version_1.py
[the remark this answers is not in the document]
nils-wisiol: as discussed offline, added "overview" endpoint similar to api root

api/desecapi/models.py
peterthomassen: replaced by more specific messages

api/desecapi/models.py
```diff
@@ -454,6 +460,91 @@ def generate_key(self):
...
+@pgtrigger.register(
+    # Ensure that token_user is consistent with token
+    pgtrigger.Trigger(
+        name='token_user',
+        operation=pgtrigger.Update | pgtrigger.Insert,
+        when=pgtrigger.Before,
+        func='NEW.token_user_id = (SELECT user_id FROM desecapi_token WHERE id = NEW.token_id); RETURN NEW;',
+    ),
+
+    # Ensure that f there is *any* domain policy for a given token, there is always one with domain=None.
```
nils-wisiol, line 486: typo

peterthomassen: fixed

api/desecapi/tests/test_token_domain_policy.py
```diff
@@ -0,0 +1,395 @@
...
+
+        ## verify object creation
+        response = self.client.get_policy(self.token, using=self.token_manage, domain=None)
+        self.assertStatus(response, status.HTTP_200_OK)
+        self.assertEqual(response.data, self.default_data | data)
+
+        ## can't create policy for other user's domain
+        data = dict(domain=self.other_domain.name, perm_dyndns=True, perm_rrsets=True)
+        response = self.client.create_policy(self.token, using=self.token_manage, data=data)
+        self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
```
nils-wisiol, line 141: is this leaking information?

api/desecapi/tests/test_token_domain_policy.py
```diff
@@ -0,0 +1,395 @@
...
+        self.assertEqual(response.data, [])
+
+        # Create
+        ## default policy
+        data = dict(domain=None, perm_rrsets=True)
+        response = self.client.create_policy(self.token, using=self.token_manage, data=data)
+        self.assertStatus(response, status.HTTP_201_CREATED)
+
+        ## can't create another default policy
+        with transaction.atomic():
```
nils-wisiol, line 129: why this is sometimes needed, sometimes not? Shouldn't atomicity be guaranteed by the API?

nils-wisiol: from the code comment below, according to https://stackoverflow.com/a/23326971/6867099, this is only needed for status code >= 400, right?

api/desecapi/tests/test_token_domain_policy.py
```diff
@@ -0,0 +1,395 @@
...
+                self.token.save()
+
+    def test_domain_deletion(self):
+        domains = [None] + self.my_domains[:2]
+        for domain in domains:
+            models.TokenDomainPolicy(token=self.token, domain=domain).save()
+
+        domain = domains.pop()
+        domain.delete()
+        self.assertEqual(list(map(lambda x: x.domain, self.token.tokendomainpolicy_set.all())), domains)
```
nils-wisiol, line 366: no domains property at `tokendomainpolicy_set`?

nils-wisiol (approved):
some questions, but this lgtm

api/desecapi/tests/test_token_domain_policy.py
peterthomassen: No, the error is that the domain does not exist in the user's domain set, and the serializer turns it into 400 (instead of 404) because it's only about a domain reference from a payload field (and not a URL resource).

The same error is returned when the domain does not exist at all, i.e. the error does not expose whether the domain is owned by another user. Further, you can always check whether a domain is owned by another user by checking whether deSEC nameservers have SOA records.

---

If you have a token that is restricted to `example.com`, you can try to work on another domain's RRsets (of the same user), and you will get 403 instead of 400 or 404. That exposes to the token user that the other domain is in the same account. However, the token user could also just `GET domains/` and look at the list of domains, so there's no additional exposure. (Unsafe methods on `domains/` require a token without a domain policy, i.e. an unrestricted token.) One can argue that a domain-scoped token should not be able to access the domain list. However, if we block that, then the 403 side-channel occurs.

Do you think it's fine that a restricted token can read the list of domains in the account?

api/desecapi/tests/test_token_domain_policy.py
peterthomassen: that is correct, and I believe it's what these tests are doing

api/desecapi/tests/test_token_domain_policy.py
peterthomassen: no: `tokendomainpolicy_set` is a set of policies, and each policy only has *one* domain. The set collection does not have a "collective attribute" to access attributes of its members (`{p.domain for p in tokendomainpolicy_set}` is not generically exposed).

api/desecapi/tests/test_token_domain_policy.py
nils-wisiol: I think it's fine, but I also don't see what could be the benefit. So maybe remove that permission?

nils-wisiol:
As far as I can tell, the last remaining question is [whether a scoped token should have the permission to list all domains belonging to user account](https://github.com/desec-io/desec-stack/pull/456#discussion_r779489592).

peterthomassen:
As discussed in person, we'll keep it as is for now, and may later implement something like `perm_view`; the `domains/` queryset can then be filtered such that it only has domains for which the token has view permission.

docs/auth/tokens.rst
```diff
@@ -268,6 +268,126 @@ If you do not have the token UUID, but you do have the token value itself, you
 can use the :ref:`log-out` endpoint to delete it.
 
 
+Token Scoping: Domain Policies
+``````````````````````````````
+Tokens by default can be used to authorize any DNS operation on any domain
+within the user's account.
```
nils-wisiol, line 274: and also for admin work, mention that here to avoid misunderstandings?

docs/auth/tokens.rst
```diff
@@ -268,6 +268,126 @@ If you do not have the token UUID, but you do have the token value itself, you
...
+Token Policies provide a means to restrict the scope of influence for a given
+API token.
+
+Using policies, the token's power is limited in two ways:
```
nils-wisiol, line 278: can be limited

docs/auth/tokens.rst
```diff
@@ -268,6 +268,126 @@ If you do not have the token UUID, but you do have the token value itself, you
...
+
+1. the types of DNS operations that can be performed, such as :ref:`dynDNS
+   updates <update-api>` or :ref:`general RRset management <manage-rrsets>`.
+
+2. the set of domains on which these actions can be performed.
+
```
nils-wisiol, line 284: say somewhere that tokens without policies are unrestricted

nils-wisiol: or introduce the term "privileged tokens" here instead of further below

nils-wisiol (changes requested):
endpoint reference is missing

docs/auth/tokens.rst
peterthomassen: done

docs/auth/tokens.rst
peterthomassen: done

docs/auth/tokens.rst
peterthomassen: done

nils-wisiol (approved):

PR #457, opened 2020-10-07
feat(docs): extend LR instructions

barrydegraaff:
adding some text that will help new users to understand what to do to set up LE

### DIFF
--- a/docs/dyndns/lets-encrypt.rst
+++ b/docs/dyndns/lets-encrypt.rst
@@ -27,12 +27,18 @@ steps.
      wget https://raw.githubusercontent.com/desec-io/certbot-hook/master/hook.sh
      wget https://raw.githubusercontent.com/desec-io/certbot-hook/master/.dedynauth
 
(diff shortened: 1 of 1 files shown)


nils-wisiol:
Nice, thanks! 

Could you please change the commit title to "feat(docs): extend LR instructions"? Then it will match our regular commit message scheme.

@peterthomassen I realized that on the confirmation page for custom domains ([example](https://desec.io/custom-setup/test20201007.example)), there are no instructions on how to proceed. We should either add a token, or a link to the passwort reset form, or something.

barrydegraaff:
@nils-wisiol changed title of PR, please let me know if you need anything else.

nils-wisiol:
Could also change the title of the commit? Unfortunately, the title of the PR is only shown on GitHub and is not stored inside the actual git repository.

nils-wisiol:
When merging this we should update the README at https://github.com/desec-utils/certbot-hook

nils-wisiol:
@barrydegraaff Also, in the title I of course meant "LE" instead of "LR"

barrydegraaff:
moving to https://github.com/desec-io/desec-stack/pull/459

PR #461, opened 2020-10-07
fix(webapp): allow _ in CNAME targets

peterthomassen:
Received a corresponding report via email. Tested locally.


nils-wisiol (approved):
oh whops

PR #468, opened 2020-11-06
20201106 increase domain limit

peterthomassen:
I noticed an uptick in registration today, and I think it's due to the c't interview that was pre-released online today. Some users have added multiple domains.

I expected more will sign up over the weekend, and I don't see a reason for restricting them to 5 domains (we picked that number conservatively for our launch). The PR increases the default domain limit to 15.

It also bumps a dependency that has an upstream vulnerability (but we do not use the affected functionality).


nils-wisiol (approved):
lgtm

Issue #347, opened 2020-04-18
api: Scoped tokens


peterthomassen:
e.g. by domain, or use case (_acme-challenge)

appliedprivacy:
coming from: https://talk.desec.io/t/restricted-access-tokens-for-letsencrypt-dns-challenge/37

When designing scoped tokens please allow for a token that is authorized to create multiple 
DNS records in one or multiple zones so that this use case can be achieved:

Token is authorized to edit/create:

- _acme-challenge.desec1.example.com TXT
- _acme-challenge.desec2.example.com TXT
- _acme-challenge.desec1.example2.com TXT

but not
- _acme-challenge.desec1.example.com AAAA
- _acme-challenge.desec3.example.com
- _acme-challenge.desec4.example2.com

Multiple tokens should be possible with the same set of permissions.
This would be handy when a multiple servers requests a certificate for multiple domains using the dns-01 challenge.

Knight1:
From my other Ticket I would like to see a Limit for IP addresses in Ranges like (1.1.1.1/32), (1.1.1.1/30).

nils-wisiol:
I'd love to see tokens limited by IP range, but any implementation should be compatible to further extension of scopes, e.g. by domain, namespace (as in "everything below foobar.example.com"), record type, yada yada.

peterthomassen:
consider to allow token expiration after a certain time

peterthomassen:
drafting in #456

peterthomassen:
#474 (subnet scoping) is done and has been deployed.

peterthomassen:
#478 (expiration by age and non-use) is done and has been deployed.

Valodim:
This is pretty useful already, thanks :+1: 

As originally requested, please consider implementing scoped tokens by domain and record type. That would be a killer feature and a considerable security improvement over the current "all or nothing" permission scheme.

generalmanager:
Great to see that some parts of this feature are already implemented.
I'm especially looking forward to be able to hand out subdomain-specific tokens to be saved on several routers, vms and appliances which update dyndns records of my custom domains.
Right now I have to use several accounts as I don't like credentials with rw access to my whole account saved on hardware of customers, family & friends.

MrSuicideParrot:
I saw the accepted pull request to implement the scoped tokens. Do you have any idea when it will be put into production?

peterthomassen:
[...]

PR #469, opened 2020-11-06
feat(api): add Token.perm_manage_tokens and surrounding functionality

peterthomassen:
The migration sets `perm_manage_tokens = True` for all existing tokens,
and then sets the default to `False`.

Related: #347

---

Todos:
- docs update
- GUI update

It may also be worthwhile to discuss why I called the field `perm_manage_tokens` (I figured it would play along best with the other things we're going to add, but that we should talk about).


peterthomassen:
Docs update and GUI adjustments are done. Ready for review!

docs/auth/tokens.rst
```diff
@@ -3,77 +3,195 @@
...
+``perm_manage_tokens``
+    :Access mode: read, write
+    :Type: boolean
+
+    Permission to manage tokens (this one and also all others).  A token which
+    does not have this flag set cannot access the ``auth/tokens/`` endpoints.
+
+``token``
+    :Access mode: read-once
+    :Type: string (28 characters)
```
nils-wisiol, line 82: Let's be not too specific with details here - this value could change in the future

webapp/src/components/Field/SwitchBox.vue
```diff
@@ -0,0 +1,52 @@
...
+    :required="required"
+    :rules="[v => !required || !!v || 'Required.']"
+    @change="input($event)"
+    @keyup="keyup($event)"
+  />
+</template>
+
+<script>
+export default {
+  name: 'Checkbox',
```
nils-wisiol, line 16: typo

nils-wisiol (approved):
very nice. just two nits :)

nils-wisiol (approved):

docs/auth/tokens.rst
peterthomassen: done

webapp/src/components/Field/SwitchBox.vue
peterthomassen: done

PR #473, opened 2020-11-10
feat(webapp): highlight dirty rows

peterthomassen:


peterthomassen:
Added a fix for broken vertical alignment. (This was broken by 6781c09f621b53ab87bab7067776bb4e3735eef4 which is already in production, unfortunately.)

nils-wisiol (approved):
lgtm

PR #492, opened 2020-12-29
feat(api,webapp): adds audio CAPTCHA for a11y

### LINKED ISSUE #470: Add captcha audio output
Visually impaired users currently cannot solve our captcha and thus cannot sign up. Alternative captcha output, in particular audio output, can solve this issue.

We use `ImageCaptcha` from the [captcha package](https://pypi.org/project/captcha/), which also provides a `AudioCaptcha` captcha class. We should add some way to retrieve the audio representation through the GUI.

nils-wisiol:
Resolves #470


api/desecapi/models.py
```diff
@@ -897,20 +897,36 @@ def _act(self):
         self.domain.save(update_fields=['renewal_state', 'renewal_changed'])
 
 
-def captcha_default_content():
-    alphabet = (string.ascii_uppercase + string.digits).translate({ord(c): None for c in 'IO0'})
-    content = ''.join([secrets.choice(alphabet) for _ in range(5)])
+def captcha_default_content(kind: str) -> str:
```
peterthomassen, line 900: Now that this is not used as a default for the field, but depends on `.kind`, it should be method of the `Captcha` class

api/desecapi/models.py
```diff
@@ -897,20 +897,36 @@ def _act(self):
...
+        alphabet = (string.ascii_uppercase + string.digits).translate({ord(c): None for c in 'IO0'})
+        length = 5
+    elif kind == Captcha.Kind.AUDIO:
+        alphabet = string.digits
+        length = 8
+    else:
+        raise ValueError(f'Unknown Captcha kind: {kind}')
+
+    content = ''.join([secrets.choice(alphabet) for _ in range(length)])
     metrics.get('desecapi_captcha_content_created').inc()
```
peterthomassen, line 911: Consider parameterizing by kind. You can do that by adding `['kind']` as the last argument [here](https://github.com/desec-io/desec-stack/blob/master/api/desecapi/metrics.py#L21) (see other examples there), and then changing this line to

```
metrics.get('desecapi_captcha_content_created').labels(kind).inc()
```

We also could do that before the `if` construction if we want to catch unknown kinds as well in monitoring. (Not sure if that's a good idea.)

api/desecapi/models.py
nils-wisiol: fixed

api/desecapi/models.py
```diff
@@ -897,20 +897,36 @@ def _act(self):
...
+        AUDIO = 'audio'
+
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     created = models.DateTimeField(auto_now_add=True)
-    content = models.CharField(
-        max_length=24,
-        default=captcha_default_content,
-    )
+    content = models.CharField(max_length=24, default="")
+    kind = models.CharField(choices=Kind.choices, default=Kind.IMAGE, max_length=24)
```
peterthomassen, line 924: Why 24?

api/desecapi/serializers.py
```diff
@@ -22,14 +23,23 @@
 
 class CaptchaSerializer(serializers.ModelSerializer):
     challenge = serializers.SerializerMethodField()
+    kind = serializers.ChoiceField(
+        choices=['image', 'audio'],  # TODO this list could be automatically generated
```
peterthomassen, line 27: `Captcha.Kind` inherits from `enum.Enum`. You should therefore be able to do:

```
choices=list(map(lambda t: t.value, Captcha.Kind))  # or list comprehension
```

However, [the `choices` argument also accepts "a list of `(key, display_name)` tuples"](https://www.django-rest-framework.org/api-guide/fields/#choicefield). It may be sufficient to do `choices=Captcha.Kind` or at least `choices=list(Captcha.Kind)`.

api/desecapi/models.py
nils-wisiol: moving this code requires fiddling with the initial migration. I say let's keep it where it is.

api/desecapi/serializers.py
```diff
@@ -22,14 +23,23 @@
...
-        fields = ('id', 'challenge') if not settings.DEBUG else ('id', 'challenge', 'content')
+        fields = ('id', 'challenge', 'kind') if not settings.DEBUG else ('id', 'challenge', 'kind', 'content')
 
     def get_challenge(self, obj: models.Captcha):
         # TODO Does this need to be stored in the object instance, in case this method gets called twice?
-        challenge = ImageCaptcha().generate(obj.content).getvalue()
+        if obj.kind == models.Captcha.Kind.IMAGE:
+            challenge = ImageCaptcha().generate(obj.content).getvalue()
+        elif obj.kind == models.Captcha.Kind.AUDIO:
+            challenge = AudioCaptcha().generate(obj.content)
```
peterthomassen, line 40: not sure why audio doesn't need `getvalue()`, but I trust you did it right

webapp/src/views/ResetPassword.vue
```diff
@@ -79,19 +79,30 @@
...
+                                          indeterminate
+                                          v-if="captchaWorking"
                                     ></v-progress-circular>
                                     <img
-                                            v-if="captcha && !captchaWorking"
-                                            :src="'data:image/png;base64,'+captcha.challenge"
-                                            alt="Passwords can also be reset by sending an email to our support."
+                                          v-if="captcha && !captchaWorking && captcha_kind == 'image'"
+                                          :src="'data:image/png;base64,'+captcha.challenge"
+                                          alt="Sign up is also possible by sending an email to our support."
```
peterthomassen, line 90: `alt` text copy paste error

webapp/src/views/ResetPassword.vue
```diff
@@ -79,19 +79,30 @@
...
+                                    </audio>
+                                    <br/>
+                                    <v-btn-toggle>
+                                      <v-btn text outlined @click="getCaptcha(true)" :disabled="captchaWorking"><v-icon>mdi-refresh</v-icon></v-btn>
+                                    </v-btn-toggle>
+                                    &nbsp;
+                                    <v-btn-toggle v-model="captcha_kind">
+                                      <v-btn text outlined value="image" aria-label="Switch to Image CAPTCHA" :disabled="captchaWorking"><v-icon>mdi-eye</v-icon></v-btn>
+                                      <v-btn text outlined value="audio" aria-label="Switch to Audio CAPTCHA" :disabled="captchaWorking"><v-icon>mdi-ear-hearing</v-icon></v-btn>
+                                    </v-btn-toggle>
```
peterthomassen, line 105: <3

webapp/src/views/ResetPassword.vue
```diff
@@ -214,6 +226,13 @@
         this.working = false;
       },
     },
+    watch: {
+      captcha_kind: function (oldKind, newKind) {
+        if (oldKind !== newKind) {
+          this.getCaptcha();
+        }
+      },
+    },
```
peterthomassen, line 235: <3

webapp/src/views/SignUp.vue
```diff
@@ -100,21 +100,33 @@
                         class="uppercase"
                         ref="captchaField"
                         tabindex="4"
+                        hint="Can\'t see? Hear an audio CAPTCHA instead."
```
peterthomassen, line 103: You could make this depend on `kind`, like `:hint="kind == 'image' ? 'Can\'t see? Hear an audio CAPTCHA instead.' : 'Trouble hearing? Switch to an image CAPTCHA.'"`

webapp/src/views/ResetPassword.vue
```diff
@@ -79,19 +79,30 @@
                                             tabindex="3"
                                     />
                                   </v-col>
```
peterthomassen, line 81: hint for the audio alternative is missing (it's in the sign-up form)

www/conf/sites-available/90-desec.static.location
```diff
@@ -3,7 +3,7 @@
 #####
 location / {
     add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload" always;
-    add_header Content-Security-Policy "default-src 'self'; frame-src 'none'; connect-src 'self'; font-src 'self'; img-src 'self' data:; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; base-uri 'self'; frame-ancestors 'none'; block-all-mixed-content; form-action 'none';" always;
+    add_header Content-Security-Policy "default-src 'self'; frame-src 'none'; connect-src 'self'; font-src 'self'; img-src 'self' data:; media-src 'self' data:; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; base-uri 'self'; frame-ancestors 'none'; block-all-mixed-content; form-action 'none';" always;
```
peterthomassen, line 6: remove `'self'`

peterthomassen (changes requested):
Great PR, very cool! :-) I have a bunch of minor / stylistic comments. But looks very good!

api/desecapi/models.py
```diff
@@ -926,7 +911,22 @@ class Kind(models.TextChoices):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if not self.content:
-            self.content = captcha_default_content(self.kind)
+            self.content = self.captcha_default_content(self.kind)
+
+    @staticmethod
+    def captcha_default_content(kind: str) -> str:
```
peterthomassen: If had meant making it an object method, and using `self.kind` ;-) But now we'll keep it where it was anyways, so all good.

api/desecapi/models.py
peterthomassen: ah yes, that makes sense. Sorry for the ill-posed suggestion!

api/desecapi/serializers.py
nils-wisiol: yup

webapp/src/views/ResetPassword.vue
nils-wisiol: fixed

webapp/src/views/SignUp.vue
nils-wisiol: done

webapp/src/views/ResetPassword.vue
nils-wisiol: done

www/conf/sites-available/90-desec.static.location
nils-wisiol: done

api/desecapi/serializers.py
nils-wisiol: seems like it is sufficient to not declare this at all :-)

api/desecapi/models.py
nils-wisiol: it needs some limit. We could do 42 :) Seriously, I chose a multiple of 8 because I found that makes sense. But that's nothing more than an educated guess

peterthomassen: Aren't the only values "Image" and "Audio"? So I would have thought, the limit is 5.

api/desecapi/models.py
nils-wisiol: done. No need to catch unknowns, the serializer doesn't allow that anyways

nils-wisiol:
I addressed all comments :+1: when this looks good, I can squash things

peterthomassen:
lgtm, modulo the thing with the `max_length` of `Captcha.kind`, but that shouldn't hold us up.

After squashing, would you mind cherry-picking 875828631be3c46d764005d28d84599430a426c3 (small docs fix)?

www/conf/sites-available/90-desec.static.location
```diff
@@ -3,7 +3,7 @@
 #####
 location / {
     add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload" always;
-    add_header Content-Security-Policy "default-src 'self'; frame-src 'none'; connect-src 'self'; font-src 'self'; img-src 'self' data:; media-src 'self' data:; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; base-uri 'self'; frame-ancestors 'none'; block-all-mixed-content; form-action 'none';" always;
```
peterthomassen: oops. Just saw that that's the wrong `'self'`.

nils-wisiol: darn

peterthomassen (approved):

PR #496, opened 2021-01-04
fix(api): only use resolvable domain names in SOA record

nils-wisiol:
This fixes an issue with SOA record validation of nic.it, which does not allow the SOA to contain domain names that do not resolve.

image

https://dns-check.nic.it/


nils-wisiol:
This only changes the SOA for new domains!

peterthomassen (approved):
lgtm!

PR #501, opened 2021-01-05
Webapp Bug Fixes and Code Style

nils-wisiol:
Fixes two webapp UI bugs:

- maintain CrudTable row highlighting (error / changed) when changing sort order
- correct display of Record field contents when data changes

Also, this refactors the code in places to increase readability and to follow vue.js patterns.


peterthomassen (approved):
lvgtm!

Issue #489, opened 2020-12-16
api: allow to postpone captcha to email verification stage


peterthomassen:
To facilitate scripted sign-ups (e.g. in [Nextcloud VM](https://github.com/nextcloud/vm)), we should allow signing up without solving the captcha. The captcha then has to be solved upon verifying the email address. After verifying the email address with a valid captcha, and given that a dynDNS domain was provided during sign-up, an API token is displayed so that users can proceed with their configuration.

Caveat: Captchas currently have one shot and are invalidated if entered incorrectly. We would like to keep that, so that they can't be solved via brute force. If that happens, we should delete the pending registration, and a new sign-up request would be required. @nils-wisiol Do you agree?

@enoch85 Tagging you to be able to follow the issue. Also, would you be able to cope with issuing a new sign-up request in case the captcha is entered incorrectly?

enoch85:
> issuing a new sign-up request in case the captcha is entered incorrectly?

If there is a way to check that with your API, sure. Otherwise it might be hard.

PR #511, opened 2021-01-25
20210119 postpone captcha

peterthomassen:
Implementation is ready. Todos:

- [x] tests
- [x] docs update
- [x] GUI support

@enoch85


enoch85:
Aah finally! 🥇 🚀

Will read the code later and come back with questions.

api/desecapi/serializers.py
```diff
@@ -579,18 +583,11 @@ def get_fields(self):
         return fields
 
     def validate_name(self, value):
-        self.raise_if_domain_unavailable(value, self.context['request'].user)
+        user = self.context['request'].user if not isinstance(self.context['request'].user, AnonymousUser) else None
```
nils-wisiol, line 586: Why not pass `AnonymousUser` to `is_registerable`? I feel like that function should be able to handle that.

peterthomassen: This would imply that the whole `Domain` model can deal with `AnonymousUser` values for `owner`, but it doesn't. The problem is that when you filter a queryset for an owner value, the ORM produces an error when the filter value is `AnonymousUser`. We could now fix that specifically in `is_registrable()`, but I think that's very inconsistent with other functionality in the model then.

We can also fix `Domain.__init__` to set `owner = None` in this case. That would solve the problem. What do you think?

nils-wisiol: I'm fine with it either way. I expect that we won't have a lot of `AnonymousUser` domain owner cases, so I don't want to over-engineer it

peterthomassen: fixed

enoch85:
Just realized I'm not that knowledgeable in Python. I trust this is good stuff anyway. 

I'm open for testing when this is ready. :+1:

peterthomassen:
@nils-wisiol I finished tests, docs, and GUi, so this is ready for (final?) review.

@enoch85 I mainly tagged you to show that there is progress ;-) -- In #489, I said that we'd have to delete the pending registration and start over if the captcha is solved incorrectly. The implementation in this PR, however, does not have this drawback. Your flow would be:
- Sign the user up via `POST https://desec.io/api/v1/auth/ <<< {"email": "...", "password": null, "domain": "foobar.dedyn.io"}`
- If the input is invalid, you get `400 Bad Request` with details in the body. For example, if the domain is unavailable, you get: `{"domain":["This domain name conflicts with an existing zone, or is disallowed by policy."]}`
- Otherwise, the status is `202 Accepted`, and we send a verification email (if the user does not exist yet, otherwise we do nothing).
- The user clicks the link in the email and arrives in the "Activate Account Confirmation" dialog of our web interface. It has a captcha and a submit button.
- If the captcha is entered incorrectly, a new captcha is displayed, and the user can try again. The user can also switch between visual and audio captcha.
- Once the captcha was submitted correctly, the process continues as if the captcha had been given during registration (i.e. a confirmation is displayed etc.). The user will have to copy the update token from the confirmation page and enter it into the Nextcloud VM configuration.

Given that it's quite simple, I don't think we need to roll this out to a staging environment for testing. Once @nils-wisiol confirms the approach, feel free to make the necessary adjustments to the POST request in https://github.com/nextcloud/vm/pull/1725.

api/desecapi/serializers.py
```diff
@@ -645,7 +640,12 @@ class Meta:
         extra_kwargs = UserSerializer.Meta.extra_kwargs
 
     def validate_domain(self, value):
-        DomainSerializer.raise_if_domain_unavailable(value, self.context['request'].user)
+        serializer = DomainSerializer(data=dict(name=value), context=self.context)
+        try:
+            serializer.is_valid(raise_exception=True)
+        except serializers.ValidationError:
+            raise serializers.ValidationError(serializer.default_error_messages['name_unavailable'],
```
nils-wisiol, line 647: If `DomainSerializer` is ever extended to check other things as well, this will behave strangely. Also, why catch the exception and re-raise instead of just letting it bubble up?

api/desecapi/serializers.py
```diff
@@ -775,14 +778,24 @@ class Meta:
...
             'domain': {'default': None, 'allow_null': True}
         }
 
+    def to_internal_value(self, data):
+        internal_value = super().to_internal_value(data)
+        try:
+            internal_value.pop('captcha')  # remove captcha from internal value to avoid passing to Meta.model(**kwargs)
+        except KeyError:
+            if internal_value['user'].needs_captcha:
+                raise serializers.ValidationError({'captcha': fields.Field.default_error_messages['required']})
```
nils-wisiol, line 802: can this be raised during the validation phase?

webapp/src/components/ActivateAccountActionHandler.vue
```diff
@@ -1,14 +1,125 @@
...
+            <br/>
+            <v-btn-toggle>
+              <v-btn text outlined @click="getCaptcha(true)" :disabled="captchaWorking"><v-icon>mdi-refresh</v-icon></v-btn>
+            </v-btn-toggle>
+            &nbsp;
+            <v-btn-toggle v-model="captcha_kind">
+              <v-btn text outlined value="image" aria-label="Switch to Image CAPTCHA" :disabled="captchaWorking"><v-icon>mdi-eye</v-icon></v-btn>
+              <v-btn text outlined value="audio" aria-label="Switch to Audio CAPTCHA" :disabled="captchaWorking"><v-icon>mdi-ear-hearing</v-icon></v-btn>
+            </v-btn-toggle>
+          </v-col>
```
nils-wisiol, line 48: man we really need to encapsulate this. I think its now the third time in the code base

api/desecapi/serializers.py
```diff
@@ -710,16 +710,22 @@ class Meta:
...
+            payload = crypto.decrypt(code.encode(), context='desecapi.serializers.AuthenticatedActionSerializer',
+                                     ttl=ttl)
             return json.loads(payload.decode())
-        except (TypeError, UnicodeDecodeError, UnicodeEncodeError, json.JSONDecodeError, binascii.Error):
+        except ValueError:  # TODO remove this once all urlsafe_b64encode'd codes have expired (~30d after deployment)
+            if _retry:
+                return cls._unpack_code(urlsafe_b64decode(code.encode()).decode(), ttl=ttl, _retry=False)
+            else:
+                raise
+        except (TypeError, UnicodeDecodeError, UnicodeEncodeError, json.JSONDecodeError, binascii.Error) as ex:
```
nils-wisiol, line 728: remove "as ex"

nils-wisiol (commented):
nice, only minor stuff

webapp/src/components/ActivateAccountActionHandler.vue
nils-wisiol: that shouldnt be part of this PR. Just sayin'

api/desecapi/serializers.py
peterthomassen: I'm not sure what you mean, but if that really happens, we'll notice thanks to our tests.

The exception is re-raised because otherwise, the JSON error output will be nested like `{"domain": {"name": "..."}}` although the input field of that endpoint is just a string with name `domain`.

api/desecapi/serializers.py
peterthomassen: Fixed. I was somehow under the impression that `validate()` runs before `to_internal_value()` so that the user object would not yet be instantiated, but in fact it's the other way around.

api/desecapi/serializers.py
peterthomassen: done

api/desecapi/serializers.py
nils-wisiol: I meant: what if `DomainSerializer` is extended to have another mandatory field? The code that you propose here does the job, because currently the only validation of `DomainSerializer` is validity and available of the name. So I wanted to point out that want we want here is that the domain *name* is valid, but what we are checking is that `{name: <domainname>}` is a valid serialization for the `DomainSerializer`. These two things currently coincide, but this may change in the future.

peterthomassen: Got it. Previously, we took the other approach (`raise_if_name_unavailable()` or something), but that caused a potential inconsistency: `DomainSerializer` has an extra `trim_whitespace` setting which the `raise_if_name_unavailable()` code was not aware of. I felt that it would be best to use the same code path for any domain name validation.

nils-wisiol: My feeling is that the proper way forward with this would be to create a subclass of [CharField](https://www.django-rest-framework.org/api-guide/fields/#charfield) that checks if the given string is a valid and available domain name. This could then be used here and in `DomainSerializer`. I guess that's nice to have, not a requirement for this PR

nils-wisiol (approved):

PR #512, opened 2021-01-30
fix(api): pin amqp version that otherwise breaks Celery

nils-wisiol:
confirmed broken is amqp==5.0.5 of 28 Jan 2021.

How do we proceed with this?


nils-wisiol:
I re-run the checks on master, and they come out successful: https://github.com/desec-io/desec-stack/runs/1797033215 I also confirmed that they are using amqp 5.0.5

On my machine, however, master remains broken.

nils-wisiol:
This only happens for older versions of our base images. Pull new docker images to resolve this!

PR #515, opened 2021-02-14
Scavenger fix

peterthomassen:


api/desecapi/tests/test_user_management.py
```diff
@@ -975,6 +994,7 @@ def test_renew_domain_fresh_183_days(self):
         domain.renewal_changed = domain.published
         domain.renewal_state = Domain.RenewalState.FRESH
         domain.save()
+        domain.rrset_set.update(touched=domain.published)
```
nils-wisiol, line 997: What does this do?

nils-wisiol (approved):
lgtm

api/desecapi/tests/test_user_management.py
peterthomassen: The scavenger previously looked at `domain.published` only, but now uses something like `max(domain.published, max(rrset.touched for rrset in domain.rrsets_set))`.

The domain contains an `NS` RRset which is more or less "now". The test has been fixing `domain.published` at various points in the past to see if the scavenger acts accordingly. With this PR's fix, that alone in itself does nothing, and one needs to pin both `domain.published` and the RRsets' `.touched` attributes in the past.

Issue #412, opened 2020-06-10
[Feature Request] DynDNS API can update subdomains


hcc23:
**Feature Description:**

Using the [dynDNS API] it is possible to update `A` and `AAAA` records of sub-domains, e.g. `sub.main.dedyn.io`.

One possible way to achieve this would be via the already existing `hostname` parameter in the API:

```
curl -u main.dedyn.io:<token> https://update.dedyn.io/?hostname=sub.main.dedyn.io
```

**Context:**
* according to [this comment](https://github.com/desec-io/desec-stack/issues/411#issuecomment-640259255) the current behavior does_not_ allow to do so
* according to [this comment](https://github.com/desec-io/desec-stack/issues/411#issuecomment-640806623) the required changes _could be a relatively easy change_.
* this feature request stems from the issue #411.

[dynDNS API]: https://desec.readthedocs.io/en/latest/dyndns/update-api.html

nils-wisiol:
I'd love to see a way of changing subdomain records via the dyn interface!

Regarding the `hostname` query parameter, I'm not sure if this is a good way to introduce this: Currently, you can update the apex A/AAAA records of any domain (zone) in your account by setting the domain name as the username you use for authentication. So, if you have a domain example.dedyn.io and a.example.dedyn.io, then you can 

    curl -u example.dedyn.io:token https://update.dedyn.io/

and 

    curl -u a.example.dedyn.io:token https://update.dedyn.io/

and the correct domain will get updated.

However, if a.example.dedyn.io is not a separate domain (zone), but just ordinarily contained in example.dedyn.io, the second command would fail.

I think the username parameter should be extended to handle this case - instead of introducing an additional hostname query parameter, otherwise things could get really confusing.

rouben:
Another approach to this could be to adopt a "proper" API like Cloudflare.

peterthomassen:
Would you mind giving a reference?

nils-wisiol:
The reason the dynDNS API is so quirky is to have it compatible with many routers out there. If you have a proper client, available, you could just use our REST API instead of the dynDNS API, as described in our docs.

rouben:
@nils-wisiol - good point and thanks for clearing that up! Some routers have very limited capabilities in terms of leveraging APIs. I understand why the dynamic DNS API is so limited and kind of strange now...

@peterthomassen - the CloudFlare API is as follows: https://api.cloudflare.com/#dns-records-for-a-zone-update-dns-record
[...]

PR #516, opened 2021-02-16
20210216 arbitrary dyn updates

peterthomassen:
As people ask about it very often, this PR enables using the dynDNS update endpoints regardless of the domain's `minimum_ttl`. The only reason not to do that would load on the secondaries. Once that becomes a problem, we can revisit. (I don't really think it will be an issue.) I went for the "just works" approach, as opposed to adding GUI functionality to change the minimum TTL.

So far, there is no documentation. I'm not sure if there should be (it would encourage such use).

I am also planning to tackle #412 in this PR.

### DIFF
--- a/api/desecapi/tests/test_dyndns12update.py
+++ b/api/desecapi/tests/test_dyndns12update.py
@@ -187,12 +187,17 @@ class SingleDomainDynDNS12UpdateTest(DynDNS12UpdateTest):
 class MultipleDomainDynDNS12UpdateTest(DynDNS12UpdateTest):
     NUM_OWNED_DOMAINS = 4
 
-    def test_honor_minimum_ttl(self):
+    def test_ignore_minimum_ttl(self):
         self.my_domain.minimum_ttl = 61
         self.my_domain.save()
-        response = self.assertDynDNS12NoUpdate(self.my_domain.name)
-        self.assertStatus(response, status.HTTP_403_FORBIDDEN)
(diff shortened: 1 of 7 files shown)


api/desecapi/tests/test_dyndns12update.py
```diff
@@ -187,12 +187,16 @@ def test_identification_by_email(self):
...
-        response = self.assertDynDNS12NoUpdate(self.my_domain.name)
-        self.assertStatus(response, status.HTTP_403_FORBIDDEN)
-        self.assertEqual(response.data['detail'], 'Domain not eligible for dynamic updates, please contact support.')
+
+        # Test that dynDNS updates work both under a local public suffix (self.my_domain) and for a custom domains
+        for domain in [self.my_domain, self.create_domain(owner=self.owner)]:
+            self.assertGreater(domain.minimum_ttl, 60)
+            self.client.set_credentials_basic_auth(domain.name.lower(), self.token.plain)
+            response = self.assertDynDNS12Update(domain.name)
+            self.assertStatus(response, status.HTTP_200_OK)
```
nils-wisiol, line 199: test that resulting TTL is actually 60

nils-wisiol (approved):
lgtm

api/desecapi/tests/test_dyndns12update.py
peterthomassen: done

Issue #187, opened 2019-05-09
dyn: support multiple hostname update


peterthomassen:
See https://help.dyn.com/remote-access-api/perform-update/ and https://sourceforge.net/p/ddclient/wiki/usage/#configuring-ddclient

Specifying `hostname=a.dedyn.io,b.dedyn.io` is part of the dynDNS protocol, so it would be a nice-to-have. (A user contacted me and was confused why it didn't work.)

PR #518, opened 2021-02-18
20210217 dyn subdomains

peterthomassen:
That was easier than I thought. Need to add tests and docs, but that should be straightforward, too.

You might want to take a look at `utils.get_domain_for_qname()`.

### DIFF
--- a/api/desecapi/authentication.py
+++ b/api/desecapi/authentication.py
@@ -14,6 +14,16 @@ from desecapi.models import Token
 from desecapi.serializers import AuthenticatedBasicUserActionSerializer, EmailPasswordSerializer
 
 
+class DynAuthenticationMixin:
+    def authenticate_credentials(self, username, key):
+        user, token = TokenAuthentication().authenticate_credentials(key)
+        if not user.is_active:
+            raise exceptions.AuthenticationFailed
+        if username not in ['', user.email] and not user.domains.filter(name=username.lower()).exists():
+            raise exceptions.AuthenticationFailed
+        return user, token
+
+
 class TokenAuthentication(RestFrameworkTokenAuthentication):
     model = Token
 
@@ -55,7 +65,7 @@ class TokenAuthentication(RestFrameworkTokenAuthentication):
         return super().authenticate_credentials(key)
 
 
-class BasicTokenAuthentication(BaseAuthentication):
+class BasicTokenAuthentication(BaseAuthentication, DynAuthenticationMixin):
     """
     HTTP Basic authentication that uses username and token.
 
@@ -87,28 +97,17 @@ class BasicTokenAuthentication(BaseAuthentication):
             msg = 'Invalid basic auth token header. Basic authentication string should not contain spaces.'
             raise exceptions.AuthenticationFailed(msg)
 
-        return self.authenticate_credentials(auth[1])
-
-    def authenticate_credentials(self, basic):
-        invalid_token_message = 'Invalid basic auth token'
         try:
-            username, key = base64.b64decode(basic).decode(HTTP_HEADER_ENCODING).split(':')
-            user, token = TokenAuthentication().authenticate_credentials(key)
-            if username not in ['', user.email] and not user.domains.filter(name=username.lower()).exists():
-                raise Exception
+            username, key = base64.b64decode(auth[1]).decode(HTTP_HEADER_ENCODING).split(':')
+            return self.authenticate_credentials(username, key)
         except Exception:
-            raise exceptions.AuthenticationFailed(invalid_token_message)
-
-        if not user.is_active:
-            raise exceptions.AuthenticationFailed(invalid_token_message)
-
-        return user, token
+            raise exceptions.AuthenticationFailed("badauth")
 
     def authenticate_header(self, request):
         return 'Basic'
 
 
-class URLParamAuthentication(BaseAuthentication):
+class URLParamAuthentication(BaseAuthentication, DynAuthenticationMixin):
     """
     Authentication against username/password as provided in URL parameters.
     """
@@ -116,8 +115,8 @@ class URLParamAuthentication(BaseAuthentication):
 
     def authenticate(self, request):
         """
-        Returns a `User` if a correct username and password have been supplied
-        using URL parameters.  Otherwise returns `None`.
+        Returns `(User, Token)` if a correct username and token have been supplied
+        using URL parameters.  Otherwise raises `AuthenticationFailed`.
         """
 
         if 'username' not in request.query_params:
@@ -127,18 +126,10 @@ class URLParamAuthentication(BaseAuthentication):
             msg = 'No password URL parameter provided.'
             raise exceptions.AuthenticationFailed(msg)
 
-        return self.authenticate_credentials(request.query_params['username'], request.query_params['password'])
-
-    def authenticate_credentials(self, _, key):
         try:
-            user, token = TokenAuthentication().authenticate_credentials(key)
-        except self.model.DoesNotExist:
-            raise exceptions.AuthenticationFailed('badauth')
-
-        if not user.is_active:
-            raise exceptions.AuthenticationFailed('badauth')
-
-        return token.user, token
+            return self.authenticate_credentials(request.query_params['username'], request.query_params['password'])
+        except Exception:
+            raise exceptions.AuthenticationFailed("badauth")
 
 
 class EmailPasswordPayloadAuthentication(BaseAuthentication):
--- a/api/desecapi/authentication.py
+++ b/api/desecapi/authentication.py
@@ -10,18 +10,20 @@ from rest_framework.authentication import (
     TokenAuthentication as RestFrameworkTokenAuthentication,
     BasicAuthentication)
 
-from desecapi.models import Token
+from desecapi.models import Domain, Token
 from desecapi.serializers import AuthenticatedBasicUserActionSerializer, EmailPasswordSerializer
 
 
 class DynAuthenticationMixin:
     def authenticate_credentials(self, username, key):
         user, token = TokenAuthentication().authenticate_credentials(key)
-        if not user.is_active:
-            raise exceptions.AuthenticationFailed
-        if username not in ['', user.email] and not user.domains.filter(name=username.lower()).exists():
-            raise exceptions.AuthenticationFailed
-        return user, token
+        # Make sure username is not misleading
+        try:
+            if username in ['', user.email] or Domain.objects.filter_qname(username.lower(), owner=user).exists():
+                return user, token
+        except ValueError:
+            pass
+        raise exceptions.AuthenticationFailed
 
 
 class TokenAuthentication(RestFrameworkTokenAuthentication):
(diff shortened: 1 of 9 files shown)


api/desecapi/authentication.py
```diff
@@ -127,18 +126,10 @@ def authenticate(self, request):
...
-            user, token = TokenAuthentication().authenticate_credentials(key)
-        except self.model.DoesNotExist:
-            raise exceptions.AuthenticationFailed('badauth')
-
-        if not user.is_active:
-            raise exceptions.AuthenticationFailed('badauth')
-
-        return token.user, token
+            return self.authenticate_credentials(request.query_params['username'], request.query_params['password'])
+        except Exception:
```
nils-wisiol, line 132: this may be too broad and swallow exceptions that we would want to know about (think, e.g., a lost db connection)

api/desecapi/utils.py
```diff
@@ -0,0 +1,14 @@
+from django.db.models import CharField, F, Value
+from django.db.models.functions import Concat, Length
+
+from desecapi.models import Domain
+
+
+def get_domain_for_qname(qname, **kwargs):
+    domain = Domain(name=qname)
+    domain.clean_fields(exclude=['owner'])  # syntactically validate domain name
```
nils-wisiol, line 9: could break when we include more fields in `Domain` -- is there a way to define a positive-list instead of an exclude list?

api/desecapi/utils.py
```diff
@@ -0,0 +1,14 @@
+from django.db.models import CharField, F, Value
+from django.db.models.functions import Concat, Length
+
+from desecapi.models import Domain
+
+
+def get_domain_for_qname(qname, **kwargs):
```
nils-wisiol, line 7: 1. this function deserves extensive tests
2. this could live in the `Domain` class instead of creating this new module
3. add type annotations?

api/desecapi/views.py
```diff
@@ -349,7 +345,8 @@ def _find_ip_v6(self, request):
         return self.find_ip(request, ['myipv6', 'ipv6', 'myip', 'ip'], version=6)
 
     def get(self, request, *_):
-        domain = self._find_domain(request)
+        qname, domain = self._find_domain(request)
+        subname = qname.rpartition(domain.name)[0]
```
nils-wisiol, line 349: maybe move this logic into `_find_domain`, that way it seems to me that `_find_domain` has a nice self-contained interface?

nils-wisiol (commented):
nice!

api/desecapi/authentication.py
peterthomassen: I was thinking to catch both exceptions from `authenticate_credentials()` (that would be `AuthenticationFailed`), but also from the inner call to `get_domain_for_qname()`, which can raise a Django validation error (via `clean_fields`). I wasn't sure how to do it smoothly without the authentication code being aware of the latter.

Maybe we should catch the validation error in `get_domain_for_qname()` and turn it into a `ValueError`, then catch that here?

api/desecapi/utils.py
peterthomassen: Unfortunately not.

Btw, why do this check at all, why worry about it here? If it's invalid, the database query will fail. -- No, because we're comparing via `_endswith`. It can happen that an invalid `qname` has a valid suffix, and we thus extract the wrong name. For example, if you own `example.com` and set `qname = 'admin@internal.example.com'`, it would match!

api/desecapi/authentication.py
```diff
@@ -14,6 +14,16 @@
 from desecapi.serializers import AuthenticatedBasicUserActionSerializer, EmailPasswordSerializer
 
 
+class DynAuthenticationMixin:
+    def authenticate_credentials(self, username, key):
+        user, token = TokenAuthentication().authenticate_credentials(key)
+        if not user.is_active:
```
nils-wisiol, line 20: This is redundant (otherwise inactive users could authenticate against `TokenAuthentication`). `TokenAuthentication` inherits from `RestFrameworkTokenAuthentication`, where `authenticate_credentials` [unconditionally checks if the user of the given token is active](https://github.com/encode/django-rest-framework/blob/1ec0f86b585cd87e4b413aeaad1ecc947bacfef2/rest_framework/authentication.py#L205).

api/desecapi/authentication.py
nils-wisiol: > Maybe we should catch the validation error in get_domain_for_qname() and turn it into a ValueError, then catch that here?

For values that fail validation I suggest that `get_domain_for_qname()` returns either None or raises an "not found" exception, which should be converted by `authenticate_credentials()` into an authentication failure

api/desecapi/views.py
```diff
@@ -328,43 +281,85 @@ def find_ip(request, params, version=4):
...
+        # username parameter
+        try:
+            return self.request.query_params['username'].lower()
+        except KeyError:
+            pass
+
+        # only domain associated with this user account
+        try:
+            return self.request.user.domains.get().name
+        except (models.Domain.MultipleObjectsReturned, models.Domain.DoesNotExist):
```
nils-wisiol, line 331: maybe give separate error messages? Could lower support load

nils-wisiol (commented):
just nitpicking.

api/desecapi/utils.py
peterthomassen: done

api/desecapi/views.py
peterthomassen: refactored

api/desecapi/authentication.py
peterthomassen: done

api/desecapi/views.py
peterthomassen: yep, already fixed

api/desecapi/authentication.py
peterthomassen: I turned `get_domain_for_qname` into `filter_qname` which is a bit more generic. It now raises a ValueError if a syntactically invalid domain name is passed. We now catch that during authentication and in the dynDNS view.

api/desecapi/utils.py
nils-wisiol: There [is a way](https://github.com/desec-io/desec-stack/compare/20210127_id_manager#diff-53877dd549623d1da7ffa608733f27c53cb91a493ba8875226ddf8020cb45853R1108-R1112) to only use the domain name validation, not sure if that's more appropriate here.

nils-wisiol: Also, technically, `admin@internal.example.com` is a valid qname, so I'm not sure how we should deal with this edge case

docs/dyndns/update-api.rst
```diff
@@ -83,8 +83,19 @@ determine the hostname, we try the following steps until there is a match:
...
+Subdomains
+----------
+The dynDNS update API can also be used to update IP records for subdomains.
+To do so, make sure that in the above list of steps, the first value
+provided contains the full domain name (including the subdomain).
+
+Example: Your domain is ``yourdomain.dedyn.io``, and you're using HTTP Basic
+Authentication.  In this case, replace your authentication username with
+``sub.yourdomain.dedyn.io``.  Similarly, if you use the ``hostname`` query
+parameter, it needs to be set to the full domain (including subdomain).
```
nils-wisiol, line 98: "full domain" -> "complete domain name"

docs/dyndns/update-api.rst
```diff
@@ -83,8 +83,19 @@ determine the hostname, we try the following steps until there is a match:
 - After successful authentication (no matter how), the only hostname that is
   associated with your user account (if not ambiguous).
 
-If we cannot determine a hostname to update, the API will return a ``404 Not
-Found`` status code.
+If we cannot determine a hostname to update, the API will return a ``400 Bad
+Request`` or ``404 Not Found`` status code (depending on the cause).
```
nils-wisiol, line 87: 400 if the domain could not be uniquely identified and 404 if no matching domain could be found

api/desecapi/models.py
```diff
@@ -378,8 +378,12 @@ def update_delegation(self, child_domain: Domain):
             raise ValueError('Cannot update delegation of %s as it is not an immediate child domain of %s.' %
                              (child_domain.name, self.name))
 
+        # Always remove delegation so that we con properly recreate it
```
nils-wisiol, line 381: I'm wondering if this incurs significant overhead for dyn IP updates?

nils-wisiol (commented):
next round :muscle:

api/desecapi/models.py
```diff
@@ -204,6 +205,20 @@ def send_email(self, reason, context=None, recipient=None):
 ]
 
 
+class DomainManager(Manager):
+    def filter_qname(self, qname: str, **kwargs) -> Domain:
```
nils-wisiol, line 209: I think this doesn't return a `Domain`, but some queryset. Although I rather see it return `Optional[Domain]`

nils-wisiol (commented):
Looking at [the new implementation of `DynDNS12Update.get`](https://github.com/desec-io/desec-stack/pull/518/files#diff-ea17007e12d5348df0100d59f45cf5f26e8247e167e552de7be4fb8fe5db9adbR354-R384), I think it could be quite easily turned into a loop that supports updating many hostnames by splitting domain parameter at any commas and repeating the process. This could not be within one transaction, but could either stop after the first failure or attempt all updates. Also, these kind of updates seldomly fail with unexpected errors. I think, the missing atomicity should not stop us from offering this interface. If you don't have time to implement this, it's a different story. :)

api/desecapi/utils.py
peterthomassen: Validation: fixed.

We allow logins with email addresses and token; allowing `@` in subdomains creates a uniqueness conflict here. Also, I'd say that dynDNS users won't usually have `@` in their dyn hostnames. No one has asked for this; I proposed to revisit once it becomes a problem.

docs/dyndns/update-api.rst
peterthomassen: done (full domain name)

docs/dyndns/update-api.rst
peterthomassen: done

api/desecapi/models.py
peterthomassen: No, that code only runs when creating or deleting a domain (not when doing an IP update)

api/desecapi/models.py
peterthomassen: fixed (`models.query.QuerySet`). That's why I don't like annotations for now, as they become inconsistent easily as long as they aren't enforced.

peterthomassen:
> Looking at [the new implementation of `DynDNS12Update.get`](https://github.com/desec-io/desec-stack/pull/518/files#diff-ea17007e12d5348df0100d59f45cf5f26e8247e167e552de7be4fb8fe5db9adbR354-R384), I think it could be quite easily turned into a loop that supports updating many hostnames [...] If you don't have time to implement this, it's a different story. :)

I'd rather do it some other time. For reference, #187.

api/desecapi/models.py
nils-wisiol: You can enforce them by observing PyCharm's warnings

nils-wisiol (approved):
lgtm

PR #521, opened 2021-03-01
feat(api): support DNAME records

peterthomassen:
It looks like DNAME substitution (RFC 6672 Sec. 2.2) is not working (try dig on a subdomain of a DNAME name with a corresponding record under the target's domain; it does not show up in the response).

While that is so, this is just scribble.


peterthomassen:
It turns out pdns has complete support, so I went ahead.

There's a catch. Let's say we have two zones containing the following records:

```bind
# example.com
www.example.com.  IN  A      1.2.3.4

# example.net
example.net.      IN  DNAME  example.com.
```

If you query `A www.example.net`, you get the DNAME record (with `example.net` owner name), a CNAME pointing from `www.example.net.` to `www.example.com.`, and the `A` record for `www.example.com` (the latter only if we are authoritative for the target zone). The CNAME content depends on the question, so the record must be synthesized, and can't be signed. However, [that is correct behavior](https://tools.ietf.org/html/rfc6672#section-5.3.1).

The thing is that this special DNAME processing, including the CNAME synthesis, happens on the secondary, and it only happens with PowerDNS auth if `dname-processing=yes` is set in the secondary config. If that is not the case, no special processing occurs (you can only query the DNAME record directly at its owner name). The stack does no special processing (the DNAME just shows up in AXFR as usual).

So, we can't really test this without involving a secondary instance, right? How to proceed? Should we be doing this at all, given that nobody has asked for it yet?

Some things to check that PowerDNS is doing things correctly:
- Test loops
- Test long names (if a zone with a short apex has a DNAME record with a long target, and is queried with a long subdomain, the effective target may exceed the DNS name length or number-of-labels limitations),
- Test occlusion (records below DNAME are not allowed (analogous to record at CNAME)).

peterthomassen:
Also, I did not add e2e2 tests yet because I wasn't sure what that would mean exactly, given the situation described above.

peterthomassen:

peterthomassen:
Thinking about it again, I think our tests should only ensure that the DNAME can be queried directly. IMHO, any synthesis testing is out of scope for our tests. What do you think, @nils-wisiol?

nils-wisiol:
My suggestion would be to include desec-ns as a subrepo in desec-stack, so that it is available for testing. Development of desec-ns would still happen at desec-ns repo, and from time to time, the subrepo at desec-stack could be updated (which would trigger integration tests). With this setup, we could have better testing, including synthesis testing. However, I believe it to be incompatible with the VPN solution for replication, so this may be blocked by desec-io/desec-ns#40

peterthomassen:
While I think that's a good idea, I also would like to see DNAME support even before we have a way of extending integration tests (incl. synthesis) to it. That functionality is not a functionality of our software, but of other software, and is not in our testing domain.

(We also don't test secondaries' answers for other record types. If we did that, we could have catched https://github.com/PowerDNS/pdns/issues/10521, but I really think it's out of scope.)

nils-wisiol (approved):
indeed, this is not within our test scope

peterthomassen:
> * [ ]  Test loops

pdns gives SERVFAIL; log says "Abort CNAME chain resolution after 10 redirects, sending out servfail."

> * [ ]  Test long names (if a zone with a short apex has a DNAME record with a long target, and is queried with a long subdomain, the effective target may exceed the DNS name length or number-of-labels limitations),

pdns gives SERVFAIL; log says nothing. However, SERVFAIL is incorrect. --> https://github.com/PowerDNS/pdns/issues/10561

> * [ ]  Test occlusion (records below DNAME are not allowed (analogous to record at CNAME)).

pdns does not honor occlusion, but we should be doing it on model level (like CNAME exclusivity). https://github.com/desec-io/desec-stack/issues/557

PR #523, opened 2021-03-24
20210324 dnssec config

peterthomassen:


peterthomassen:
Here's a screenshot of the new feature matrix:

image

I think it catches the eye at the right places, but if it's too fidgety, I can remove some of the markup.

peterthomassen:
> I think it catches the eye at the right places, but if it's too fidgety, I can remove some of the markup.

The PR now contains a Vuetify update, which makes `<code>` elements gray instead of read. This leads to a big improvement in the above screenshot.

peterthomassen:

api/desecapi/pdns.py
```diff
@@ -90,11 +90,16 @@ def get_keys(domain):
     """
     Retrieves a dict representation of the DNSSEC key information
     """
+    def _filter_ds(key):
+        key['ds'] = [ds for ds in key['ds'] if int(ds.split()[2]) in [2, 4]]
```
nils-wisiol, line 94: consider putting that in settings.py or at least in a local/module variable so that it becomes really obvious what is done here

webapp/src/components/Field/Record/TLSA.vue
```diff
@@ -0,0 +1,29 @@
+<script>
+import { and, helpers, integer, between } from 'vuelidate/lib/validators';
+import Record from '../Record.vue';
+
+const hex = helpers.regex('hex', /^([0-9a-fA-F]\s*)*[0-9a-fA-F]$/);
+const trim = and(helpers.regex('trimBegin', /^[^\s]/), helpers.regex('trimEnd', /[^\s]$/));
+
+const MAX8 = 255;
+const int8 = between(0, MAX8);
```
nils-wisiol, line 9: chic

webapp/src/components/Field/Record/TLSA.vue
```diff
@@ -0,0 +1,29 @@
...
+      { label: 'Usage', validations: { integer, int8 } },
+      { label: 'Selector', validations: { integer, int8 } },
+      { label: 'Matching Type', validations: { integer, int8 } },
+      { label: 'Certificate Data', validations: { trim, hex } },
+    ],
+    errors: {
+      integer: 'Please enter an integer.',
+      int8: `0 ≤ … ≤ ${MAX8}`,
+      hex: 'Please use hexadecimal format.',
+      trim: 'Only internal whitespace allowed.',
```
nils-wisiol, line 25: I find "internal" hard to understand here

nils-wisiol (approved):
lgtm! A few nits. Review of webapp things was rather sloppy.

api/desecapi/pdns.py
peterthomassen: This piece of code will go away with pdns 4.5, and the behavior is then governed by PowerDNS' publish-cds setting (which we already use for CDS records). See https://github.com/PowerDNS/pdns/pull/10217

As this is not permanent, I prefer to keep as is

webapp/src/components/Field/Record/TLSA.vue
peterthomassen: As this would be the only change, I think it's not worth the effort now, and we should merge. (Besides, what would be a better word?)

peterthomassen:
trying to tigger the stuck build job

PR #526, opened 2021-04-12
20210411 domain throttling

peterthomassen:


api/desecapi/views.py
```diff
@@ -70,14 +70,19 @@ class DomainViewMixin:
...
         # noinspection PyUnresolvedReferences
-        super().initial(request, *args, **kwargs)
+        return self.request.user.domains.get(name=self.kwargs['name'])
+
+    def check_throttles(self, request):
         try:
-            # noinspection PyAttributeOutsideInit, PyUnresolvedReferences
-            self.domain = self.request.user.domains.get(name=self.kwargs['name'])
+            # noinspection PyStatementEffect
+            self.domain  # call the method behind the property
```
nils-wisiol, line 93: that's a weird usage of a property. Maybe convert it to a method?

peterthomassen: right. Did it differently now

api/desecapi/throttling.py
```diff
@@ -25,6 +25,11 @@ def allow_request(self, request, view):
         if self.rate is None:
             return True
 
+        # Amend scope with optional bucket
+        bucket = getattr(view, self.scope_attr + '_bucket', None)
+        if bucket is not None:
+            self.scope += ':' + bucket
```
nils-wisiol, line 31: why not let the view decide that

nils-wisiol (approved):
lgtm

PR #531, opened 2021-04-27
fix(api): properly return mandate reference for donations, fixes #530

### LINKED ISSUE #530: webapp: broken display of donation mandate references

peterthomassen:
The response is generated from serializer.data, which is based on only
request data until .save() is called.  After that, is contains a
representation of the instance created during the .save() call.

This commit moves the instance creation from the view to the .save()
call so that the .data logistics works properly (without actually
saving anything).


nils-wisiol (approved):
lgtm

PR #532, opened 2021-05-07
feat(webapp): add partner logos

peterthomassen:
In response to deSEC integration in Nextcloud VM.

This is how it looks like:
image


nils-wisiol (approved):
:rocket:

PR #537, opened 2021-05-16
fix(docs): Running Certbot headless requires different arguments now

logolive:
According to the errors and warnings returned by certbot
A) --manual-public-ip-logging-ok is depricated, and
B) You are now required to register by mail.

This Pull request replaces the relevant arguments in the documentation

### DIFF
--- a/docs/dyndns/lets-encrypt.rst
+++ b/docs/dyndns/lets-encrypt.rst
@@ -48,22 +48,24 @@ steps.
      certbot --manual --manual-auth-hook ./hook.sh --manual-cleanup-hook ./hook.sh \
          --preferred-challenges dns -d "YOURDOMAINNAME.dedyn.io" certonly
          
-   You can also use certbot to get wildcard certificates like so::
+   Please note that the hook script may wait up to two minutes to ensure that
+   the challenge was correctly published.
+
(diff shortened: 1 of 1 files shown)


peterthomassen:
Thank you for change a)!

The certbot docs still seem to indicate that providing an email address is optional. Do you have a reference for the claim that it is now required?

logolive:
Hey @peterthomassen :)
Running
`sudo certbot --manual --manual-auth-hook /bin/echo --manual-cleanup-hook /bin/echo     --preferred-challenges dns -d "YOURDOMAINNAME.dedyn.io" -n certonly`
results in the following error:
`You should register before running non-interactively, or provide --agree-tos and --email <email_address> flags.`

Adding just `--agree-tos` does so as well. So yes, at least for my installation of certbot 1.15.0 (Arch Linux Official Repositories, installed 5 min. ago) the mail address seems necessary. Using the above command, you should be able to reproduce the error.

I do not recall the address being validated however.
In any case, it seems wise to provide it, since LetsEncrypt sends out reminders for renewals.

peterthomassen (approved):
Thanks for explaining! I updated the commit accordingly, but also discovered an opt-out option that is now mentioned.

I took the opportunity and made some other updates to this docs section. :-)

PR #539, opened 2021-05-18
20210518 django32

peterthomassen:
Updating to Django 3.2 comes with two notable points:
- The default datatype for automatically generated primary key IDs is about to change. The old setting creates warnings. I changed the setting to the new recommendation and added a migration.
- Test data from `setUpTestData()` are now isolated between tests. This has exposed two tests that only succeeded due to side effects, when in fact there was a bug.

The two affected tests are `test_pdns_change_tracker.py::RRSetTestCase.test_update_type` and `.test_update_subname`. They change an RRset's `type` or `subname` attribute directly on the instance, then save it, and observe that no pdns backend request occurs. However, such requests do occur. In Django < 3.2, they did not, because a previous test would have deleted that RRset, so the two tests that are now failing used to be no-op. (I found when run individually, these tests also used to fail.)

The problem does not occur when trying to change these things via the API, as the serializer rejects that (and we have tests for that).

When changing the attributes directly on the RRset instance, my understanding of the current implementation is that a new RRset will be created with the new subname or type, but the old one will remain in place on pdns level, but not on the API level. If I read the code correctly, we will thus have an inconsistency with orphaned RRsets in pdns.

We should fix the model (ideally) or the change tracker to reject such changes. Before I make these changes -- do you agree, @nils-wisiol?


nils-wisiol:
I think the easiest way here is to adjust the model to now allow type and subname changes. Adding this functionality can be done later (and adding it now would only lead to dead code).

peterthomassen (commented):
The problem is that if you *do* have the new git version, then `git init` (without `-b`) will produce non-empty stderr, which will raise an exception in our code. The stderr output is:

```
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
```

The suggested setting to suppress the message does not seem appropriate, because a) the old git version doesn't support it, b) I don't think there is write permission in the global configuration location.

---

What matters is that the container image we're building works correctly. It is Alpine-based and does have an up-to-date git version. I don't think we should work around some issue with Ubuntu 20.04 here, as we're not using Ubuntu in this context. (If the code runs elsewhere during development, I'd say that the dev environment needs to be set up with the correct dependencies.) What do you think?

peterthomassen:
Oops. I meant to attach this comment to https://github.com/desec-io/desec-stack/pull/539/commits/bcac3b83102a18fadbd3362280697cb00cf26a93.

Addendum: The stderr message that is caused when using the new git version means that the tests won't pass inside the api container.

peterthomassen:
> I think the easiest way here is to adjust the model to now allow type and subname changes. Adding this functionality can be done later (and adding it now would only lead to dead code).

Done. (The model already allowed it, so I just dropped the broken tests, 8a377954519eedefc119e962b003c1549051bb38).

We still test the serializer's functionality that `subname` and `type` can't be changed through public interfaces (https://github.com/desec-io/desec-stack/blob/20210518_django32/api/desecapi/tests/test_rrsets.py#L795-L812), so that should be ok.

IMO, it only remains to drop 79c60ff8033d7b6fdceca123bbf72cd18260cbc7, and then this should be good to go.

peterthomassen:
I cherry-picked a few commits from #378: that PR is blocked, but the cherry-picked commits are still stylistic improvements, so I figured we should not loose them.

@nils-wisiol Can you take a look at this PR again if you get a chance? Also, do you agree we should drop 79c60ff8033d7b6fdceca123bbf72cd18260cbc7? (That would resolve the test failures.)

nils-wisiol:
> > I think the easiest way here is to adjust the model to now allow type and subname changes. Adding this functionality can be done later (and adding it now would only lead to dead code).
> 
> Done. (The model already allowed it, so I just dropped the broken tests, [8a37795](https://github.com/desec-io/desec-stack/commit/8a377954519eedefc119e962b003c1549051bb38)).

unfortunately a typo changed the meaning of my statement: I meant to suggest to adjust the model to *not* allow type and subname changes.

nils-wisiol:
> The problem is that if you _do_ have the new git version, then `git init` (without `-b`) will produce non-empty stderr, which will raise an exception in our code.

Yes let's drop 79c60ff for now.

README.md
```diff
@@ -351,7 +351,7 @@ While there are certainly many ways to get started hacking desec-stack, here is
         2. In Project: desec-stack › Project Interpreter, add a new interpreter. Choose "existing environment" and select `api/venv/bin/python3` from the project root.
         3. In Languages & Frameworks › Django, enable the Django support and set the Django project root to `api/`.
 
-    1. From the PyCharm menu, select Run › Edit Configurations and select the "Django tests" template from the list.
+    1. From the PyCharm menu, select Run › Edit Configurations click on "Edit configuration templates"; select the "Django tests" template from the list.
```
nils-wisiol, line 354: typo: "and" missing

docs/dns/rrsets.rst
```diff
@@ -646,21 +661,29 @@ Record types with priority field
...
+              --header "Authorization: Token {token}" \
+              --header "Content-Type: application/json" --data @- <<< \
+              '{"type": "TXT", "records": ["\"test value1\"","\"value2\""], "ttl": 3600}'
+
+    - Record values consisting of several token strings separated by
+      whitespace are permitted, for example:
+      ``["\"token 1 of record 1\" \"token 2 of record 1\"", "\"record 2\""]``
+
+      Token strings longer than 255 characters are automatically split into
+      several token strings.  (The maximum length is given by the DNS
```
nils-wisiol, line 679: clarify: *This* maximum length is given by

docs/dns/rrsets.rst
```diff
@@ -624,9 +640,8 @@ Record types with priority field
       internally managed SOA record).
 
       If you need redirect functionality at the zone apex, consider using the
-      ``HTTPS`` record type which serves exactly this purpose. Although new,
-      browser vendor support is under way (with Chrome planning to roll out
-      experimental support in February 2021).
+      ``HTTPS`` record type which serves exactly this purpose.  (Note that
```
nils-wisiol, line 643: you may want to put a date here

docs/dns/rrsets.rst
```diff
@@ -542,11 +548,21 @@ Special care needs to be taken with some types of records, as explained below.
...
-    disabled in our API.  If you attempt to create such RRsets, you will receive
-    a ``400 Bad Request`` response.
+``ALIAS``
+    Due to conflicts with the security guarantees we would like to give, this
+    record type is disabled in our API.  If you attempt to create such RRsets,
+    you will receive a ``400 Bad Request`` response.
+
+    If you need redirect functionality at the zone apex, consider using the
+    ``HTTPS`` record type which serves exactly this purpose.  (Note that this
+    record type is not yet supported in all browsers.)
```
nils-wisiol, line 558: you may want to put a date here

api/desecapi/models.py
```diff
@@ -99,7 +99,7 @@ def _limit_domains_default():
     is_active = models.BooleanField(default=True)
     is_admin = models.BooleanField(default=False)
     created = models.DateTimeField(auto_now_add=True)
-    limit_domains = models.IntegerField(default=_limit_domains_default.__func__, null=True, blank=True)
+    limit_domains = models.PositiveIntegerField(default=_limit_domains_default.__func__, null=True, blank=True)
```
nils-wisiol, line 102: Does this include zero? We may want to set `limit_domains = 0` one day.

nils-wisiol (approved):
minor nits; lgtm

docs/dns/rrsets.rst
peterthomassen: I couldn't find anything reliable

docs/dns/rrsets.rst
peterthomassen: I couldn't find anything reliable

api/desecapi/models.py
peterthomassen: yes, it does (https://docs.djangoproject.com/en/3.2/ref/models/fields/#positiveintegerfield)

nils-wisiol: Weird name for a field that allows zero. :nerd_face:

peterthomassen:
> > > I think the easiest way here is to adjust the model to now allow type and subname changes. Adding this functionality can be done later (and adding it now would only lead to dead code).
> > 
> > 
> > Done. (The model already allowed it, so I just dropped the broken tests, [8a37795](https://github.com/desec-io/desec-stack/commit/8a377954519eedefc119e962b003c1549051bb38)).
> 
> unfortunately a typo changed the meaning of my statement: I meant to suggest to adjust the model to _not_ allow type and subname changes.

Sure. I think the best way to go is a database trigger, but that's actually unrelated to the Django update. I created #553 and will get to it in a few days in a separate PR.

docs/dns/rrsets.rst
nils-wisiol: I meant: as of 06/2021, this is not yet supported in all browsers

README.md
peterthomassen: done

docs/dns/rrsets.rst
peterthomassen: done

docs/dns/rrsets.rst
peterthomassen: done

PR #542, opened 2021-06-03
Add notes about ddclient use and avoid use of ipv6

lulis:
Add notes to `ddclient` documentation:
- Clarify use of `usev6` (only on Debian) and point reference docs
- Add note to avoid auto-set IPv6 when unwanted


peterthomassen (approved):
Thanks for the docs improvement! It's much appreciated <3

PR #546, opened 2021-06-14
20210614 find zone

peterthomassen:


docs/dns/domains.rst
```diff
@@ -183,6 +183,51 @@ returns the domain object in the response body.  Otherwise, the return status
...
 
+Identifying the Responsible Domain for a DNS Name
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have several domains which share a DNS suffix (i.e. one domain is a
+parent of the other), it is sometimes necessary to find out which domain is
+responsible for a given DNS name.  (In DNS terminology, the responsible domain
+is also called the "authoritative zone".)
+
+The responsible domain for a given DNS query name (``qname``) can be retrieved
```
nils-wisiol, line 194: I think the correct terminology is "owner name", and, in contrast to "domain" vs "zone", I do not see an advantage in using the less accurate "query name" over "owner name", so I'd suggest to consider to change that.

nils-wisiol (approved):
lgtm + 1 nit

docs/dns/domains.rst
peterthomassen: I intentionally avoided that because "owner name" may, for a layperson, seem related to the domain owner, but isn't. Query name, on the other hand, plays nicely with "qname", which is used in the DNS spec and means "query domain name".

nils-wisiol: fair enough :+1:

PR #547, opened 2021-06-18
20210618 rr validation

peterthomassen:


peterthomassen:
For the record: pdns does not require names to be hostnames for the following record types, but they still cause an internal server error when passed a hostname containing e.g. `$` due to different escaping between pdns and dnspython: AFSDB, CNAME, KX, NAPTR, PTR, RP

I did not SVCB/HTTPS for now (specification is not final).

nils-wisiol (approved):
lgtm

I'm not 100% confident that there are no RRs in any production database which violate these rules. If they exist, they certainly belong to somehow broken zones.

PR #549, opened 2021-06-20
fix(api): clarify throttling algorithm, fixes #548

### LINKED ISSUE #548: doc: is not clear whether the rate limits are rolling windows or fixed times
The doc on rate limits (https://desec.readthedocs.io/en/latest/rate-limits.html) is not clear about what the “Rate” columns means.

“1 day” e.g. could mean from 24 hours ago until now (= the request time), which would be a rolling window as a rate limit.
“1 day” could e.g. however also mean that the rate limiting applies for the whole day once you are limited. If that would be the case, you maybe should/could/want to add the time zone, you use for doing that. Because at 00:00 the rate limit would then reset…

peterthomassen:


nils-wisiol (approved):
lgtm

PR #551, opened 2021-06-29
fix(api): don't use Celery 5.0.6 (broken)

peterthomassen:
due to https://github.com/celery/celery/issues/6829

Celery 5.0.6 release has been retracted from GitHub, but not from PyPI. Today's deployment of #547 thus caused deployment of the broken version. 

The impact is unknown (it appears that sign-up emails are delivered).

Lesson: Wait for post-merge CI to succeed before deploying.

### DIFF
--- a/api/requirements.txt
+++ b/api/requirements.txt
@@ -1,5 +1,5 @@
 captcha~=0.3.0
-celery~=5.0.2
+celery~=5.0.2,!=5.0.6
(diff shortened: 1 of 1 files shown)


nils-wisiol:
We should also consider to eventually modify our build pipeline to run precisely the tested builds.

nils-wisiol (approved):

PR #552, opened 2021-07-01
Security Headers

peterthomassen:


peterthomassen:
Regarding security.txt:
- Unfortunately, [the `Expires:` field is required](https://datatracker.ietf.org/doc/html/draft-foudil-securitytxt-12#section-3.5.5), so I set it to a large value to avoid maintenance.
- For now, I skipped the `Encryption:` field (used to publish OpenPGP keys to be used for security notifications). Do we want to publish PGP keys? The spec [allows various URI formats](https://datatracker.ietf.org/doc/html/draft-foudil-securitytxt-12#section-3.5.4) pointing to e.g. a web resource, an OPENPGPKEY record, or stating a fingerprint only. What do you think is best?

nils-wisiol (approved):
lgtm. No PGP key needed for now, too much maintenance for too little (possibly none ever) benefit.

PR #556, opened 2021-07-02
Fix wildcard domains in filter_qname

### LINKED ISSUE #555: api: wildcard domains don't work with IP update API
I noticed that the dynDNS/IP update API doesn't accept wildcard domains, I get a `badauth` when trying to specify a wildcard domain as username or `nohost` when specifying as GET parameter.

Skimming through [the code](https://github.com/desec-io/desec-stack/blob/b6f3d47bf4b725cd0547d1087439fcff6118d594/api/desecapi/views.py#L288), I'm not quite sure as to why. Do you think that's fixable? Is it by design? I'd be happy to try and fix it myself if you push me in the right direction.

Eisfunke:
Fixes #555

### DIFF
--- a/api/api/settings.py
+++ b/api/api/settings.py
@@ -109,7 +109,7 @@ REST_FRAMEWORK = {
         # ScopedRatesThrottle
         'account_management_active': ['3/min'],  # things with side effect, e.g. sending mail or zone creation on signup
         'account_management_passive': ['10/min'],  # things like GET'ing v/* or auth/* URLs, or creating/deleting tokens
-        'dyndns': ['1/min'],  # dynDNS updates; anything above 1/min is a client misconfiguration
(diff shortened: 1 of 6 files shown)


peterthomassen:
Thanks a lot, that looks good!

There was a small issue in your test, were you checked IP addresses at subnames covered by the wildcard. However, that matching is only done during DNS resolution; the tests see the API database only, and thus need to check the wildcard explicitly.

I squashed the two commits into one, removed some code duplication, and added tests to `test_filter_qname` and `test_filter_qname_invalid`. I think this is good to go!

PS: We currently have some issues with our end-to-end tests, which is why the CI/CD pipeline fails more often than it should. Sorry if that was confusing during. I'm working on fixing that.

nils-wisiol (approved):

PR #565, opened 2021-07-13
See if CI Pipeline works 

nils-wisiol:
Do not merge this.


nils-wisiol:
Pipeline shows two tests as expected. Everything good.

PR #566, opened 2021-07-13
20210713 pdns4.5

peterthomassen:


api/desecapi/models.py
```diff
@@ -502,7 +502,7 @@ def interval_label(self):
     'IPSECKEY',  # broken in pdns, https://github.com/PowerDNS/pdns/issues/10589 TODO enable with pdns auth > 4.5.0
     'KEY',  # Application use restricted by RFC 3445, DNSSEC use replaced by DNSKEY and handled automatically
     'WKS',  # General usage not recommended, "SHOULD NOT" be used in SMTP (RFC 1123)
-}
+} | {'NID', 'L32', 'L64', 'LP'}  # https://github.com/rthalley/dnspython/issues/674
```
nils-wisiol, line 505: why use the union operator `|` here?

api/desecapi/pdns.py
```diff
@@ -18,6 +18,9 @@
 
     # "additional" types, without obsolete ones
     'DHCID', 'DLV', 'EUI48', 'EUI64', 'IPSECKEY', 'KX', 'MINFO', 'MR', 'RKEY', 'WKS',
+
+    # https://doc.powerdns.com/authoritative/changelog/4.5.html#change-4.5.0-alpha1-New-Features
+    'NID', 'L32', 'L64', 'LP'
```
nils-wisiol, line 23: aren't those 'major' types? (This classification originally stems from the pdns docs.)

nils-wisiol (approved):
two nits

api/desecapi/models.py
peterthomassen: Just for grouping, because I expect that we will remove the four of them together. I was too lazy "sorting them into the list" properly. (It would have required one line per entry, so should I have four comments, too?)

api/desecapi/pdns.py
peterthomassen: No, [they aren't listed in the pdns docs](https://doc.powerdns.com/authoritative/appendices/types.html), but they are announced in the 4.5 release notes. So I made a new category here.

I just submitted [a PR for documentation](https://github.com/PowerDNS/pdns/pull/10607).

api/desecapi/models.py
nils-wisiol: I think it could be added as a single line in the first definition. But whatever :)

PR #571, opened 2021-09-20
20210920 tsig axfr

peterthomassen:
- Existing zones need their SOA values updated (manual process)
- Existing zones need to have the TSIG key assigned: `pdnsutil pdnsutil activate-tsig-key $zone default primary`


docker-compose.yml
```diff
@@ -191,12 +191,16 @@ services:
     init: true
     cap_add:
     - NET_ADMIN
+    ports:
+      - "53:53"
```
nils-wisiol: could be restricted to one or more explicitly given IP addresses

nils-wisiol:
[pdns response code: 422, body: {"error": "A TSIG key with the name 'default' does not exist"}](https://github.com/desec-io/desec-stack/runs/3661782347#step:5:90483)

docker-compose.yml
peterthomassen: desec-stack does not know the public-facing IP address

peterthomassen:
This is ready for review. The previous reasons for test failure have been fixed.

The current failure is due to a glitch in the MariaDB respository. (I'm going to file a report with them.)

peterthomassen:
For reference, https://github.com/MariaDB/mariadb-docker/issues/394

PR #588, opened 2022-01-05
feat(docs): remove certbot hook, refer to certbot plugin

nils-wisiol:
I've stripped the certbot instructions to, essentially, a link to the certbot plugin. I figure, it's better to have a single place to maintain them. Also, I've moved it out of the "dynDNS" section, as it has wider applicability. In absence of a better idea, I created a chapter called "Applications". (How about "Integrations"? Any idea is welcome here.)

See also desec-io/certbot-dns-desec#11


nils-wisiol:
The former failure in test run stems from an error where `ipaddress.ip_network` receives `None` as an argument.

I do not understand how this code ever worked:

1. [In line 24](https://github.com/desec-io/desec-stack/blob/main/test/e2e2/spec/test_dyndns.py#L24), `ipaddress.ip_network` is called and `network` is passed as sole argument.
2. [In line 49](https://github.com/desec-io/desec-stack/blob/main/test/e2e2/spec/test_dyndns.py#L49), `_ips_in_network` in line 24 is called and `expected_net` is passed as argument.
3. `expected_net` is filled with the `expected_ipv4` or `expected_ipv6` parameters, depending on the loop iteration.
4. `expected_ipv4` and `expected_ipv6` parameters are filled in [line 77](https://github.com/desec-io/desec-stack/blob/main/test/e2e2/spec/test_dyndns.py#L77) with gives unconditionally `None` to `expected_ipv4` in the second loop iteration.

From how I read this, the second loop iteration should unconditionally result in this `ValueError`:

```python
>>> import ipaddress
>>> ipaddress.ip_network(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/ipaddress.py", line 83, in ip_network
    raise ValueError('%r does not appear to be an IPv4 or IPv6 network' %
ValueError: None does not appear to be an IPv4 or IPv6 network
```

But it doesn't. What am I missing?

A new run of the test suite without modifications did not show any error.

docs/applications/lets-encrypt.rst
```diff
@@ -0,0 +1,24 @@
+TLS Certificates with Let's Encrypt
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+certbot with deSEC Plugin
+`````````````````````````
+deSEC supports the DNS challenge protocol to make it easy for you to
```
peterthomassen, line 6: Add "ACME" somewhere for SEO

docs/applications/lets-encrypt.rst
```diff
@@ -0,0 +1,24 @@
...
+We currently know of the following:
+
+- `acme.sh <https://github.com/Neilpang/acme.sh/wiki/dnsapi#71-use-desecio>`_
+- `cert-manager web hook <https://github.com/kmorning/cert-manager-webhook-desec>`_
+  (Kubernetes)
+- `lego <https://github.com/go-acme/lego>`_
+- `Posh-ACME <https://github.com/rmbolger/Posh-ACME/blob/main/Posh-ACME/Plugins/DeSEC-Readme.md>`_
+- `Terraform vancluever/acme <https://registry.terraform.io/providers/vancluever/acme/latest/docs/guides/dns-providers-desec>`_
+
+Our forum has `a more updated list <https://talk.desec.io/t/tools-implementing-desec/11>`_.
```
peterthomassen, line 24: `more comprehensive list of tools and integrations around deSEC`

docs/index.rst
```diff
@@ -40,6 +40,13 @@ of the ``curl`` call, like this::
    dns/rrsets
 
 
+.. toctree::
+   :maxdepth: 2
+   :caption: Applications
+
+   applications/lets-encrypt
```
peterthomassen, line 47: `Integrations` would be much better

docs/applications/lets-encrypt.rst
```diff
@@ -0,0 +1,24 @@
...
+`````````````````````````
+deSEC supports the DNS challenge protocol to make it easy for you to
+obtain wildcard certificates for your domain name easily from anywhere.
+All you need is `certbot <//certbot.eff.org/>`_, your credentials
+and our `certbot plugin <//pypi.org/project/certbot-dns-desec//>`_.
+
+
+Other ACME Clients
+``````````````````
+Beside certbot, there are other ACME clients that support deSEC out of the box.
```
peterthomassen, line 14: `Besides`

peterthomassen (changes requested):
small nits

docs/index.rst
peterthomassen: ... and I would move it below dynDNS (before API Summary)

docs/applications/lets-encrypt.rst
```diff
@@ -0,0 +1,24 @@
+TLS Certificates with Let's Encrypt
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+certbot with deSEC Plugin
+`````````````````````````
+deSEC supports the DNS challenge protocol to make it easy for you to
+obtain wildcard certificates for your domain name easily from anywhere.
+All you need is `certbot <//certbot.eff.org/>`_, your credentials
+and our `certbot plugin <//pypi.org/project/certbot-dns-desec//>`_.
```
peterthomassen, line 9: double `/` *at the end*

peterthomassen:
> I do not understand how this code ever worked:
> 
> 1. [In line 24](https://github.com/desec-io/desec-stack/blob/main/test/e2e2/spec/test_dyndns.py#L24), `ipaddress.ip_network` is called and `network` is passed as sole argument.

The full function is:
```python
    def _ips_in_network(ip_set, network):
        return all(ipaddress.ip_address(ip) in ipaddress.ip_network(network) for ip in ip_set)
```

So, if `network is None`, we get `ipaddress.ip_address(ip) in ipaddress.ip_network(None)` evaluated **for each** `ip` in `ip_set`. The test passes `network=None` only under circumstances where `ip_set` is expected to be empty.

More precisely: `ip_set` is the set of IPv4 or IPv6 addresses received from API, nslord or nsmaster, *after the update*, see [lines 47-40](https://github.com/desec-io/desec-stack/blob/main/test/e2e2/spec/test_dyndns.py#L47-L49). In some test cases, e.g. when removing the IPv4 address in an update call ([line 72](https://github.com/desec-io/desec-stack/blob/main/test/e2e2/spec/test_dyndns.py#L72)), we expect that there is no IPv4 address once the update is processed. We thus pass `network=None`, but it is never evaluated as `ip_set` is expected to be empty.

If there is a replication delay, `ip_set` may not be empty (still containing the old IP), and `ipaddress.ip_network(None)` gets evaluated and raises. This is correct behavior, although a bit obscure.

We could improve this by adding ``assert not (bool(ip_set) != network is None), "`ip_set` wasn't empty when expected"`` in the previous line. (Btw, check out how the inline backticks are working! (e.g. by quoting and looking at the quote)

docs/applications/lets-encrypt.rst
nils-wisiol: done

docs/applications/lets-encrypt.rst
nils-wisiol: done

docs/index.rst
nils-wisiol: done

docs/applications/lets-encrypt.rst
nils-wisiol: done

docs/applications/lets-encrypt.rst
nils-wisiol: done

nils-wisiol:
These changes will break [https://desec.readthedocs.io/en/latest/dyndns/lets-encrypt.html](https://desec.readthedocs.io/en/latest/dyndns/lets-encrypt.html)

nils-wisiol:
> We could improve this by adding `` assert bool(ip_set) != network is None, "`ip_set` wasn't empty when expected" `` in the previous line. (Btw, check out how the inline backticks are working! (e.g. by quoting and looking at the quote)

This line accepts only two cases:

1. `bool(ip_set) == False` and `network is None`, which is fine because `ipaddress.ip_network(network)` won't be called
2. `bool(ip_set) == True` and `network is not None`, which is fine because `ipaddress.ip_network(network)` will only be called on a value that is not `None`

However, this third case is seems also acceptable:

3. `bool(ip_set) == False` and `network is not None`, which is fine for either of above reasons.

The third case is not covered by `bool(ip_set) != network is None`. I included a statement covering all three cases in a fixup commit.

peterthomassen:
> 5\. `bool(ip_set) == False` and `network is not None`, which is fine for either of above reasons.

I didn't include it because it's unclear what that case would mean. We also don't use it currently; if it came up, it would be out of what was originally intended. We will now silently accept it. (But fine.)

nils-wisiol:
Ready to merge?

docs/integrations/lets-encrypt.rst
```diff
@@ -0,0 +1,24 @@
...
+We currently know of the following:
+
+- `acme.sh <https://github.com/Neilpang/acme.sh/wiki/dnsapi#71-use-desecio>`_
+- `cert-manager web hook <https://github.com/kmorning/cert-manager-webhook-desec>`_
+  (Kubernetes)
+- `lego <https://github.com/go-acme/lego>`_
+- `Posh-ACME <https://github.com/rmbolger/Posh-ACME/blob/main/Posh-ACME/Plugins/DeSEC-Readme.md>`_
+- `Terraform vancluever/acme <https://registry.terraform.io/providers/vancluever/acme/latest/docs/guides/dns-providers-desec>`_
+
+Our forum has `more comprehensive list of tools and integrations around deSEC <https://talk.desec.io/t/tools-implementing-desec/11>`_.
```
peterthomassen, line 24: `a` missing after `has`

peterthomassen (changes requested):
almost :)

peterthomassen (approved):

PR #591, opened 2022-01-12
20220112 email verified

peterthomassen:


nils-wisiol (approved):

peterthomassen:
The PR improves (hopefully) on the way authenticated actions are processed. `User.is_active` is now nullable, with the null value representing pending activation. Activated users have `is_active = True`, and locked ones have `is_active = False`, as before. Users which have not logged in and previously had `is_active = False` are pending activation (in the old scheme) and are migrated to "null".

Authenticated actions now only can be performed by active users (except account activation, which only requires that `is_active != False`, so null is fine). Locked users cannot perform any authenticated actions any longer.

The PR also adds `User.email_verified` which is set any time an auth action code is presented and passes the serializer validation. This means that the code is structurally valid and not expired (this includes cases where the action is not executed because the user is locked, but the code was otherwise valid).

We started verifying email addresses on Oct 23, 2019. Users which signed up on or after this date and have `is_active = True` or have logged in at least once have `email_verified` set to `created`. (The last case covers locked users, for which we set `is_active = False` manually at some point.)

peterthomassen:
rebased on top of #592, i.e. first commit doesn't actually belong to this PR

api/desecapi/migrations/0020_user_email_verified.py
```diff
@@ -0,0 +1,30 @@
...
+import datetime
+
+from django.db import migrations, models
+from django.db.models import F, Q
+
+
+def forwards_func(apps, schema_editor):
+    User = apps.get_model("desecapi", "User")
+    db_alias = schema_editor.connection.alias
+    filter = Q(is_active=True) | ~Q(last_login__isnull=True)
```
nils-wisiol, line 12: Is this the same as `Q(is_active=True) | Q(last_login__isnull=False)`?

api/desecapi/migrations/0020_user_email_verified.py
```diff
@@ -0,0 +1,30 @@
...
+    filter_kwargs = dict(created__date__gte=datetime.date(2019, 11, 1))
```
nils-wisiol, line 13: What's the difference between `filter` and `filter_kwargs`? Couldn't they be one expression? Are their conditions ANDed or ORed?

nils-wisiol (changes requested):
I've read everything before "feat(api): add User.outreach_preference" and left questions and comments on the code. Rest will follow later.

> feat(api): introduce auth action to verify email for legacy accounts
>
> This commit can be reverted once all email addresses have been
> verified.

I believe the migration file must be retained and reverted by another migration. Maybe instead implement (now or later) an action for re-verification of accounts that can be triggered if needed.

api/desecapi/authentication.py
```diff
@@ -156,9 +156,17 @@ class AuthenticatedBasicUserActionAuthentication(BaseAuthentication):
...
+        return self.authenticate_credentials(view.get_serializer_context())
+
+    def authenticate_credentials(self, context):
+        serializer = AuthenticatedBasicUserActionSerializer(data={}, context=context)
         serializer.is_valid(raise_exception=True)
-        return serializer.validated_data['user'], None
+
+        user = serializer.validated_data['user']
+        if user.is_active == False:  # don't catch None here!
+            raise exceptions.AuthenticationFailed('User inactive.')
```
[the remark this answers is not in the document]
peterthomassen: All out-of-the-box Django authenticators have this check.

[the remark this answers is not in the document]
nils-wisiol: Good point, but this PR also introduces a permission for this purpose, no?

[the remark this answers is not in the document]
peterthomassen: No. This authenticator fails if `user.is_active` is exactly `False`, i.e. it does not fail when activation is pending (`user.is_active is None`). The `IsActiveUser` permission on the other hand requires `user.is_active` to be truthy, i.e. active *and* activated.

A user that is not active has no business with the API whatsoever. A user which is pending or fully active has a different set of permissible activities (e.g. pending: performing the activation action but not DNS updates), so that's a permission.

api/desecapi/templates/emails/activate/content.txt
```diff
@@ -1,11 +1,16 @@
 Hi there,
 
-Thank you for registering with deSEC!
-
+Thank you for registering with deSEC!{% if domain is not None %} You are about to set up the
+following domain: {{ domain }}{% endif %}
+{% if domain is None %}
```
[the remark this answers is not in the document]
nils-wisiol: actually, both versions are okay

api/desecapi/serializers.py
```diff
@@ -728,36 +728,35 @@ class AuthenticatedActionSerializer(serializers.ModelSerializer):
...
+
     class Meta:
         model = models.AuthenticatedAction
         fields = ('state',)
 
     @classmethod
     def _pack_code(cls, data):
         payload = json.dumps(data).encode()
-        code = crypto.encrypt(payload, context='desecapi.serializers.AuthenticatedActionSerializer').decode()
+        code = crypto.encrypt(payload, context=cls._crypto_context).decode()
```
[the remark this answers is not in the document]
nils-wisiol: generic implementations of this class also need to pack/unpack codes of specific instances; `cls.__name__` would break this functionality

api/desecapi/models.py
```diff
@@ -953,6 +953,7 @@ class AuthenticatedUserAction(AuthenticatedBasicUserAction):
     Abstract AuthenticatedBasicUserAction, incorporating the user's id, email, password, and is_active flag into the
     Message Authentication Code state.
     """
+    __state_fields = None
```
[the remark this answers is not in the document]
nils-wisiol: added fixup commit with alternative proposal

api/desecapi/authentication.py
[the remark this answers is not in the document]
nils-wisiol: alright let's add some more docs and keep it this way

api/desecapi/models.py
```diff
@@ -192,6 +193,24 @@ def send_email(self, reason, context=None, recipient=None):
         metrics.get('desecapi_messages_queued').labels(reason, self.pk, lanes[reason]).observe(num_queued)
         return num_queued
 
+    def send_confirmation_email(self, reason, recipient=None, params=None, **kwargs):
```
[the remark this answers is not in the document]
nils-wisiol: Consider making the used interfaces and/or data structures more explicit/less anonymous. E.g., attach to class specific to the action type in question.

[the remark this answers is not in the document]
peterthomassen: also, outsource link generation to a template tag

api/desecapi/views.py
```diff
@@ -477,12 +477,10 @@ def perform_create(self, serializer):
         email = EmailMessage(subject_tmpl.render(context),
                              content_tmpl.render(context),
                              from_tmpl.render(context),
-                             ['donation@desec.io'],
-                             attachments=[
-                                 ('jameica-directdebit.xml',
-                                  attachment_tmpl.render(context),
-                                  'text/xml')
-                             ])
+                             ['support@desec.io'],
```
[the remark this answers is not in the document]
peterthomassen: done

api/desecapi/models.py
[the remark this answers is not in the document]
peterthomassen: done

api/desecapi/migrations/0020_user_email_verified.py
peterthomassen: yes, fixed

api/desecapi/migrations/0020_user_email_verified.py
peterthomassen: > Couldn't they be one expression?

No, because the Q objects (for doing OR) have to go to `filter()`'s `args`, and the date check has to go to `kwargs`.

api/desecapi/authentication.py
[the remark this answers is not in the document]
peterthomassen: done

api/desecapi/models.py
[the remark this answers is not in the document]
peterthomassen: done

api/desecapi/templates/emails/activate-account/content.txt
```diff
@@ -11,7 +11,7 @@ To create your account and finish the registration, please confirm you
 received this email by clicking on the following link (valid for {{ link_expiration_hours }}
 hours):
 {% endif %}
-{{ confirmation_link }}
+{{ confirmation_link.0 }}
```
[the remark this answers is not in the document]
peterthomassen: replaced by link generation template tag

peterthomassen:
> I've read everything before "feat(api): add User.outreach_preference" and left questions and comments on the code. Rest will follow later.

Addressed all your comments. The confirmation email sending is refactored in 65ed7a9d9d6e5b651e6d95bb801ce75e11360909. I have a few thoughts on that:
- `confirm-account` and `change-outreach-preference` confirmations still need to have their templates adapted
- In conventional DRF use, `save()` would call `update()` when `self.instance` is present (as is the case here), which requires having called `.is_valid()`; however, `.is_valid()` is not applicable here (as we don't initialize the serializer with a serialized instance). I still used `save()` because it makes clear that some kind of write operation happens. I'm happy to rename the method to something else though, if we don't want to break DRF style.
- Sending a confirmation email from a view now requires 1.) instantiating the action, 2.) passing it to the serializer, 3.) saving the serializer. Steps 2 and 3 are pure boilerplate, so we could give the action seralizer a class method that takes the arguments for 1.) and also executes 2.) and 3.). What do you think?
- I introduced the `action_user` property on the action serializers so that the `save()` method can be written identically both for individual and for list serializers, so that it could get outsourced to `AuthenticatedBasicUserActionMixin`.
  * Is that overengineered?
  * I called the property `user` initially, but that collided with the case where `user` is also the name of an action parameter (which is always the case at the moment). I can imagine similar problems to happen again in the future, without getting caught, with unintended side effects. -- Thus: Should be prefix attributes that are not action parameters with an underscore (`_reason`, `_validity_period`, `_user` (instead of `action_user`))? -- Those would then be accessed freely across the `AuthenticatedActionSerializer` inheritance hierarchy.
  * I also investigated moving the email target user to `context`, but it looked like that would cause more problems than it would solve.

> > feat(api): introduce auth action to verify email for legacy accounts
> > This commit can be reverted once all email addresses have been
> > verified.
> 
> I believe the migration file must be retained and reverted by another migration. Maybe instead implement (now or later) an action for re-verification of accounts that can be triggered if needed.

Good point. Regarding your suggestion, isn't that what this commit implements? (The only change being that we wouldn't revert it.)

peterthomassen:
rebased on top of main

peterthomassen:
Also, 67ee0d8a5d0237f491bcd56382af5d779df68f52 essentially makes 6719dc5d5ce851372967ae5d4cd07b4db07d5bd2 obsolete (both are in this PR). But I suspect it would be very messy to get rid of the latter ...

api/desecapi/models.py
```diff
@@ -981,6 +982,21 @@ def _state_fields(self):
...
 
+class AuthenticatedChangeOutreachPreferenceUserAction(AuthenticatedEmailUserAction):
+    outreach_preference = models.BooleanField()
+
+    class Meta:
+        managed = False
+
+    @property
+    def _state_fields(self):
+        return super()._state_fields + [self.user.outreach_preference] + [self.outreach_preference]  # both needed?
```
peterthomassen, line 974: code should not get invalidated just because it has already been applied (idempotent)

api/desecapi/templates/emails/change-outreach-preference/content.txt
```diff
@@ -0,0 +1,7 @@
+Hi there,
+
+Stay secure,
```
peterthomassen, line 3: add management command for adding content

api/desecapi/serializers.py
```diff
@@ -760,7 +760,7 @@ def to_representation(self, instance: models.AuthenticatedAction):
 
     def to_internal_value(self, data):
         # calculate code TTL
-        validity_period = self.context.get('validity_period', self.validity_period)
+        validity_period = self.context.get('validity_period', self.validity_period)  # TODO understand why necessary
```
peterthomassen, line 763: to make sure that serializer in authentication.py gets the right validity period

Improve comment; also, try if fallback is needed or subscript suffices

api/desecapi/serializers.py
```diff
@@ -779,10 +779,11 @@ def to_internal_value(self, data):
             raise serializers.ValidationError({api_settings.NON_FIELD_ERRORS_KEY: msg})
 
         # add extra fields added by the user
-        unpacked_data.update(**data)
+        data = {**data, **unpacked_data}
```
nils-wisiol, line 782: this fixes a problem, add test

api/desecapi/serializers.py
```diff
@@ -793,17 +793,43 @@ def save(self, **kwargs):
         raise ValueError
 
 
-class AuthenticatedBasicUserActionSerializer(AuthenticatedActionSerializer):
+class AuthenticatedBasicUserActionMixin():
+    def save(self, recipient=None):
+        context = {**self.context, 'action_serializer': self}
+        return self.action_user.send_email(self.reason, context=context, recipient=recipient)
```
nils-wisiol, line 799: consider moving this to models.py

nils-wisiol:
>     * In conventional DRF use, `save()` would call `update()` when `self.instance` is present (as is the case here), which requires having called `.is_valid()`; however, `.is_valid()` is not applicable here (as we don't initialize the serializer with a serialized instance). I still used `save()` because it makes clear that some kind of write operation happens. I'm happy to rename the method to something else though, if we don't want to break DRF style.

I'm fine with save.

> 
>     * Sending a confirmation email from a view now requires 1.) instantiating the action, 2.) passing it to the serializer, 3.) saving the serializer. Steps 2 and 3 are pure boilerplate, so we could give the action seralizer a class method that takes the arguments for 1.) and also executes 2.) and 3.). What do you think?

I like that.

> 
>     * I introduced the `action_user` property on the action serializers so that the `save()` method can be written identically both for individual and for list serializers, so that it could get outsourced to `AuthenticatedBasicUserActionMixin`.
>       
>       * Is that overengineered?
>       * I called the property `user` initially, but that collided with the case where `user` is also the name of an action parameter (which is always the case at the moment). I can imagine similar problems to happen again in the future, without getting caught, with unintended side effects. -- Thus: Should we prefix attributes that are not action parameters with an underscore (`_reason`, `_validity_period`, `_user` (instead of `action_user`))? -- Those would then be accessed freely across the `AuthenticatedActionSerializer` inheritance hierarchy.
I think action_user is fine.
> 
> > > feat(api): introduce auth action to verify email for legacy accounts
> > > This commit can be reverted once all email addresses have been
> > > verified.
> > 
> > 
> > I believe the migration file must be retained and reverted by another migration. Maybe instead implement (now or later) an action for re-verification of accounts that can be triggered if needed.
> 
> Good point. Regarding your suggestion, isn't that what this commit implements? (The only change being that we wouldn't revert it.)

:+1:

nils-wisiol (approved):

api/desecapi/serializers.py
peterthomassen: done

webapp/src/views/SignUp.vue
```diff
@@ -248,13 +248,23 @@
...
         try {
           this.captcha = (await HTTP.post('captcha/', {kind: this.captcha_kind})).data;
-          if(focus) {
-            this.$refs.captchaField.focus()
+        } catch (error) {
+          if (error.response && error.response.status == 429) {
+            this.captcha = null;
+            let retryAfter = parseInt(error.response.headers['retry-after']);
+            this.captcha_errors = [`Too many attempts. Please try again in ${retryAfter} seconds.`];
+            setTimeout(() => this.captcha || this.getCaptcha(), 1000 * retryAfter);
```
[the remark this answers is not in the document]
peterthomassen: moved to #616

api/desecapi/models.py
peterthomassen: done (removed property)

api/desecapi/templates/emails/change-outreach-preference/content.txt
peterthomassen: done! see c2ad5d9d0076932f0da76596302a4db173d7843a

peterthomassen: done! see c2ad5d9d0076932f0da76596302a4db173d7843a @nils-wisiol

api/desecapi/serializers.py
peterthomassen: I thought about this, and I don't see how it could be done:

- `send_email` is already in models.py (belongs to `User`)
- `save()` has to remain some way or the other, as it is used for the part of the serializer that changes state (here: sends email). See also https://github.com/desec-io/desec-stack/pull/591#issuecomment-1123577349 and https://github.com/desec-io/desec-stack/pull/591#issuecomment-1163349605
- What remains is that `save()` could be calling `save()` on the corresponding `AuthenticatedAction` instance, which would have to be instantiated first. -- But with which arguments? (I'm not sure if `self.Meta.model(**self.data).save()` would always be correct.)
- What's more, sending the email in `AuthenticatedAction.save()` wouldn't work when we have a `ListSerializer` (e.g. for multiple domain renewals), where we only want to send one email (but there would be multiple `AuthenticatedAction` instances -- if we call `save()` on all of them, we'd get multiple emails).

All in all, I think it's best to leave this as is. It seems rather natural that the serializer coalesces any `many=True` actions, and then does what ever to initiate the state change.

api/desecapi/serializers.py
peterthomassen: done, dca593d647678e8e38bc1677f11e4fc97d4405ab

peterthomassen:
I believe all comments have been addressed! The outreach management command is now there and covers all cases we need (I think; let's talk about how it does that). I think it's ready!

peterthomassen:
Alright, this time ready for real. Did some minor tweaks here and there, added some unrelated bug fixes, and put all the commits in right order.

api/desecapi/management/commands/outreach-email.py
```diff
@@ -0,0 +1,62 @@
...
+        if isinstance(backend, DjangoTemplates):
+            return backend
+    raise ImproperlyConfigured("No DjangoTemplates backend is configured.")
+
+
+class Command(BaseCommand):
+    help = 'Reach out to users with an email. Takes email template on stdin.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('email', nargs='*', help='User(s) to contact, identified by their email addresses')
```
nils-wisiol, line 25: Add to help: Defaults to everyone with outreach preference

nils-wisiol (approved):
This lgtm!

nils-wisiol:
Note the one comment above. While at it, maybe add an example for how to use this to README?

api/desecapi/management/commands/outreach-email.py
peterthomassen: done

PR #592, opened 2022-01-18
chore(api),feat(api): dnspython upgrade, support L32/L64/LP/NID

peterthomassen:


peterthomassen:
ready for review.

peterthomassen:
Remaining to dos: fix test failures (seems straightforward), and probably should add e2e tests for record types introduced in 91e84442d44abaa2450c39871274ded4dd8b888a.

peterthomassen:
all done!

api/desecapi/dns.py
```diff
@@ -25,6 +25,15 @@ def _strip_quotes_decorator(func):
 dns.rdtypes.svcbbase.PortParam.to_text = _strip_quotes_decorator(dns.rdtypes.svcbbase.PortParam.to_text)
 
 
+@dns.immutable.immutable
+class CERT(dns.rdtypes.ANY.CERT.CERT):
```
nils-wisiol, line 29: how is this different from base class? add short description

test/e2e2/spec/test_api_rr.py
```diff
@@ -40,7 +40,7 @@ def generate_params(dict_value_lists_by_type: dict) -> List[Tuple[str, str]]:
         '6454 8 1 24396e17e36d031f71c354b06a979a67a01f503e',
         '0 0 0 00',
     ],
-    'CERT': ['6 0 0 sadfdQ=='],
+    'CERT': [],  # handled in non-canonical, because apply presentation format conversion on the stack (but not here)
```
nils-wisiol, line 43: Add a canonical example. (I do not understand if/why there can't be one.)

nils-wisiol (approved):

api/desecapi/dns.py
peterthomassen: upstream implementation calls _ctype_to_text. Added comment

test/e2e2/spec/test_api_rr.py
peterthomassen: It's not possible, because we have an override for the presentation format conversion of the CERT rrtype in dns.py on the stack (see above), but not in the testing container. We could only do this if we also mock dnspython in the test env. Not sure if that makes sense.

The non-canonical tests cover this type too, so I feel nothing is really missing.

nils-wisiol: alright :+1:

nils-wisiol:
this lgtm

PR #593, opened 2022-01-25
feat(nsmaster): do not sign outgoing NOTIFYs

peterthomassen:
Our secondaries will have different TSIG key configurations, and
nsmaster currently cannot pick the right key.  As a result,
secondaries would discard (some) NOTIFYs.

Outgoing AXFRs can still use TSIG, as the AXFR query contains the
required parameters.

Related: https://github.com/PowerDNS/pdns/issues/10867


nils-wisiol (approved):
lgtm

PR #595, opened 2022-02-07
fix(api): enforce (non-)apex requirements from DNSKEY, DS

peterthomassen:
Related: https://github.com/PowerDNS/pdns/pull/10959

This we deployed PowerDNS 4.6.0, non-compliant RRsets are rejected by pdns. Let's reject them earlier to get a better error message.

### DIFF
--- a/api/desecapi/models.py
+++ b/api/desecapi/models.py
@@ -209,7 +209,7 @@ validate_domain_name = [
 class DomainManager(Manager):
     def filter_qname(self, qname: str, **kwargs) -> models.query.QuerySet:
         try:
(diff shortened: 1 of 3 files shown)


nils-wisiol (approved):

nils-wisiol:
This looks good to me. Do we need similar checks for DNAME? [RFC6672](https://datatracker.ietf.org/doc/html/rfc6672) says:

> DNAME RRs MUST NOT appear at the same owner name as an NS RR unless
> the owner name is the zone apex; if it is not the zone apex, then the
> NS RR signifies a delegation point, and the DNAME RR must in that
> case appear below the zone cut at the zone apex of the child zone.

PR #602, opened 2022-03-15
fix(gitzones): avoid deadlock when waiting for git diff

nils-wisiol:
Using pipes for stdout, stderr can cause deadlocks, see
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

Also, the old version used busy waiting. Solution used now is
recommended by Python docs.


peterthomassen (approved):
lgtm.

Would you mind cherry-picking d547c1eb99e3db7db805b89c6cd09ef7b97c47b9 and a83a9d6157120c894995232e6d9e1cbf9b932595?

nils-wisiol:
I've also added a fix for the PSL upgrade. Please confirm

peterthomassen (approved):
Confirmed.

Fyi, full psl-dns upgrade adaptations are here: https://github.com/desec-io/desec-stack/pull/597/commits/93b67a7090ade6c699c5826e3a99f3a440bc09b2 No need to apply them now.

nils-wisiol:
93b67a7090ade6c699c5826e3a99f3a440bc09b2 will automatically pull psl 1.2 when released and hence doesn't solve the problem permanently. I'm voting to extend 93b67a7090ade6c699c5826e3a99f3a440bc09b2 to set the requirement to ~=1.2.0. But IMHO that not urgent, so let's merge this first. :-)

PR #611, opened 2022-07-15
20220629 webapp image

peterthomassen:
This restores webapp deployability, which was recently broken.

In some cases, the `webapp` stage of the `www` Dockerfile is built multiple times. I'd like to leave optimization of this for later, because deployability trumps build speed.

The api dependency bumps are mainly in order to update Python and Django, to fix some recent vulnerabilities (which don't seem to affect us, though).


nils-wisiol (approved):
lgtm :+1:

PR #612, opened 2022-07-25
Introduce Management Command for Abuse Cases

nils-wisiol:
Introduces the `stop-abuse` command, which takes a list of domain names and:

1. removes all contents of the domains under these names except for `NS` records at the apox, and
2. disables all user accounts associated with any of the given domain names

### DIFF
--- a/api/desecapi/management/commands/stop-abuse.py
+++ b/api/desecapi/management/commands/stop-abuse.py
@@ -1,0 +1,21 @@
+from django.core.management import BaseCommand
+
+from desecapi.models import RRset, Domain, User
+from desecapi.pdns_change_tracker import PDNSChangeTracker
+
+
+class Command(BaseCommand):
+    help = 'Removes all DNS records from the given domain and suspends the associated user'
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain-name', nargs='*', help='Domain(s) to remove all DNS records from')
+
+    def handle(self, *args, **options):
+        with PDNSChangeTracker():
+            domains = Domain.objects.filter(name__in=options['domain-name'])
+            users = User.objects.filter(domains__name__in=options['domain-name'])
+            rrsets = RRset.objects.filter(domain__name__in=options['domain-name']).exclude(type='NS', subname='')
+            print(f'Deleting {rrsets.count()} RRset(s) from {domains.count()} domain(s); '
+                  f'disabling {users.count()} associated user account(s).')
+            rrsets.delete()
+        users.update(is_active=False)
--- a/api/desecapi/management/commands/stop-abuse.py
+++ b/api/desecapi/management/commands/stop-abuse.py
@@ -1,21 +1,46 @@
 from django.core.management import BaseCommand
+from django.db.models import Q
 
+from api import settings
 from desecapi.models import RRset, Domain, User
 from desecapi.pdns_change_tracker import PDNSChangeTracker
 
 
 class Command(BaseCommand):
-    help = 'Removes all DNS records from the given domain and suspends the associated user'
+    help = 'Removes all DNS records from domains given either by name or by email address of their owner. ' \
+           'Locks all implicated user accounts.'
 
     def add_arguments(self, parser):
-        parser.add_argument('domain-name', nargs='*', help='Domain(s) to remove all DNS records from')
+        parser.add_argument('names', nargs='*',
+                            help='Domain(s) and User(s) to truncate and disable identified by name and email addresses')
 
     def handle(self, *args, **options):
         with PDNSChangeTracker():
-            domains = Domain.objects.filter(name__in=options['domain-name'])
-            users = User.objects.filter(domains__name__in=options['domain-name'])
-            rrsets = RRset.objects.filter(domain__name__in=options['domain-name']).exclude(type='NS', subname='')
+            # domains to truncate: all domains given and all domains belonging to a user given
+            domains = Domain.objects.filter(
+                Q(name__in=options['names']) |
+                Q(owner__email__in=options['names'])
+            )
+
+            # users to lock: all associated with any of the domains and all given
+            users = User.objects.filter(
+                Q(domains__name__in=options['names']) |
+                Q(email__in=options['names'])
+            )
+
+            # rrsets to delete: all belonging to (all domains given and all domains belonging to a user given)
+            rrsets = RRset.objects.filter(
+                Q(domain__name__in=options['names']) |
+                Q(domain__owner__email__in=options['names'])
+            )
+
             print(f'Deleting {rrsets.count()} RRset(s) from {domains.count()} domain(s); '
                   f'disabling {users.count()} associated user account(s).')
+
+            # delete rrsets and create default NS records
             rrsets.delete()
+            for d in domains:
+                RRset.objects.create(domain=d, subname='', type='NS', ttl=3600, contents=settings.DEFAULT_NS)
+
+        # lock users
         users.update(is_active=False)
--- a/api/desecapi/tests/test_stop_abuse.py
+++ b/api/desecapi/tests/test_stop_abuse.py
@@ -1,0 +1,57 @@
+from django.core import management
+
+from desecapi import models
+from desecapi.tests.base import DomainOwnerTestCase
+
+
+class StopAbuseCommandTest(DomainOwnerTestCase):
+
+    @classmethod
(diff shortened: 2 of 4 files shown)


api/desecapi/tests/test_stop_abuse.py
```diff
@@ -0,0 +1,54 @@
...
+        super().setUpTestDataWithPdns()
+        cls.create_rr_set(cls.my_domains[1], ['127.0.0.1', '127.0.1.1'], type='A', ttl=123)
+        cls.create_rr_set(cls.other_domains[1], ['40.1.1.1', '40.2.2.2'], type='A', ttl=456)
+        for d in cls.my_domains + cls.other_domains:
+            cls.create_rr_set(d, ['ns1.example.', 'ns2.example.'], type='NS', ttl=456)
+            cls.create_rr_set(d, ['ns1.example.', 'ns2.example.'], type='NS', ttl=456, subname='subname')
+            cls.create_rr_set(d, ['"foo"'], type='TXT', ttl=456)
+
+    def test_noop(self):
+        # test implicit by absence assertPdnsRequests
```
peterthomassen, line 20: Sure? (Otherwise, `assertPdnsNoRequestsBut` with empty argument.)

api/desecapi/tests/test_stop_abuse.py
```diff
@@ -0,0 +1,54 @@
...
+    def test_remove_rrsets(self):
+        with self.assertPdnsRequests(self.requests_desec_rr_sets_update(name=self.my_domain.name)):
+            management.call_command('stop-abuse', self.my_domain)
+        self.assertEqual(models.RRset.objects.filter(domain__name=self.my_domain.name).count(), 1)  # only NS left
+
+    def test_disable_user(self):
+        with self.assertPdnsRequests(self.requests_desec_rr_sets_update(name=self.my_domain.name)):
+            management.call_command('stop-abuse', self.my_domain)
+        self.owner.refresh_from_db()
+        self.assertEqual(self.owner.is_active, False)
```
peterthomassen, line 32: I'd have made that one test, but ok

api/desecapi/tests/test_stop_abuse.py
```diff
@@ -0,0 +1,54 @@
...
+        with self.assertPdnsRequests(self.requests_desec_rr_sets_update(name=self.my_domain.name)):
+            management.call_command('stop-abuse', self.my_domain)
+        self.assertGreater(models.RRset.objects.filter(domain__name=self.my_domains[1].name).count(), 1)
+
+    def test_only_disable_owner(self):
+        with self.assertPdnsRequests(self.requests_desec_rr_sets_update(name=self.my_domain.name)):
+            management.call_command('stop-abuse', self.my_domain)
+        self.assertEqual(self.other_domain.owner.is_active, True)
+
+    def test_only_disable_owners(self):
```
peterthomassen, line 44: This test does not stand up to the "only" in its name

api/desecapi/management/commands/stop-abuse.py
```diff
@@ -0,0 +1,21 @@
...
+        parser.add_argument('domain-name', nargs='*', help='Domain(s) to remove all DNS records from')
+
+    def handle(self, *args, **options):
+        with PDNSChangeTracker():
+            domain_f = Domain.objects.filter(name__in=options['domain-name'])
+            user_f = User.objects.filter(domains__name__in=options['domain-name'])
+            rrset_f = RRset.objects.filter(domain__name__in=options['domain-name']).exclude(type='NS', subname='')
+            print(f'Deleting {rrset_f.count()} RRset(s) from {domain_f.count()} domain(s); '
+                  f'disabling {user_f.count()} associated user account(s).')
+            rrset_f.delete()
```
peterthomassen, line 20: This needs to be done by domain, as pdns (change tracker) doesn't support multi-domain updates, right?

api/desecapi/management/commands/stop-abuse.py
```diff
@@ -0,0 +1,21 @@
...
+    help = 'Removes all DNS records from the given domain and suspends the associated user'
+
+    def add_arguments(self, parser):
+        parser.add_argument('domain-name', nargs='*', help='Domain(s) to remove all DNS records from')
+
+    def handle(self, *args, **options):
+        with PDNSChangeTracker():
+            domain_f = Domain.objects.filter(name__in=options['domain-name'])
+            user_f = User.objects.filter(domains__name__in=options['domain-name'])
+            rrset_f = RRset.objects.filter(domain__name__in=options['domain-name']).exclude(type='NS', subname='')
```
peterthomassen, line 17: `domains__in=domain_f` (and `domain__in=domain_f`)?

Also, I find `_f` confusing. It looks like a file handle. But can stay that way :) No need to spend time on.

peterthomassen (changes requested):
The only thing potentially holding up is the multi-domain update. Otherwise, lgtm!

api/desecapi/tests/test_stop_abuse.py
nils-wisiol: done

api/desecapi/tests/test_stop_abuse.py
nils-wisiol: I did two because of the function name 🤷

api/desecapi/tests/test_stop_abuse.py
nils-wisiol: Yes, sure, by virtue of having `DomainOwnerTestCase` as parent class

api/desecapi/management/commands/stop-abuse.py
nils-wisiol: PDNSChangeTracker supports updates across multiple domains, but without guaranteeing atomic updates (multiple pdns API requests are used). Check [this loop](https://github.com/desec-io/desec-stack/blob/main/api/desecapi/pdns_change_tracker.py#L289). I'm unsure why the loop is only over domain additions and RRset creations; will look into further.

nils-wisiol: PDNSChangeTracker gurantees that the key sets of additions, modifications, and deletions are always equal, so the loop is indeed over all additions, modifications, and deletions, even though it looks like it's only over additions.

nils-wisiol:
Any thoughts on the name of the command? Maybe we want an extension that empties all zones of the associated user(s)?

api/desecapi/management/commands/stop-abuse.py
nils-wisiol: done

peterthomassen:
> Should we remove or reset the NS RRsets at the apex?

No preference. Reset would be clean, in a way.

peterthomassen:
> Any thoughts on the name of the command?

I think the name is catchy, but not very accurate :) `purge-zone-data`?

We could make `--lock-owner` a flag, so that we don't have to consider that aspect in the command name itself. Also, would come handy when cleaning up a test zone or something. Perhaps, for the purpose, it should default to false.

> Maybe we want an extension that empties all zones of the associated user(s)?

Perhaps. Although then the user would have to be the argument. Or would that apply to all users implicated by the domains passed in the arguments?

Also, we have the functionality implicitly: we can always fetch the domain list of the user, and pass the list of domains to the command (even when run before, i.e. after disabling the user, this can still happen). IMO that suffices for now.

nils-wisiol:
I've kept the name but now allow the command to take domain names and email addresses in mixed fashion as arguments. It will now:

1. Truncate all domains given on the command line
2. Truncate all domains of users given on the command line
3. Lock all users implicated with any given domain name
4. Lock all users given on the command line

api/desecapi/tests/test_stop_abuse.py
```diff
@@ -20,31 +21,68 @@ def test_noop(self):
...
+            expect_order=False,
+        ):
+            management.call_command('stop-abuse', self.owner.email)
+        self.assertEqual(models.RRset.objects.filter(domain__name=self.my_domain.name).count(), 1)  # only NS left
+        self.assertEqual(
+            set(models.RR.objects.filter(rrset__domain__name=self.my_domain.name).values_list('content', flat=True)),
+            set(settings.DEFAULT_NS),
+        )
+
+    def test_disable_user_by_name(self):
```
peterthomassen, line 45: by domain name

peterthomassen (approved):
lgtm, just one nit!

PR #615, opened 2022-07-26
feat(api): command for changing minimum TTL and domains limits

nils-wisiol:
Introduces a command to change limitations to user accounts and domains. Exact CLI is totally up for discussion, this is just a first idea. Open questions are how this could be extended to transfer domains between accounts, how to change other account details like email addresses, or trigger password resets, and so on.

Change number of domains limit for user@example.com to 50 domains:

    manage.py limits domains user@example.com 50

Change minimum TTL for example.com to 600:

    manage.py limits ttl example.com 600

Existing domains/records that are in violation of the newly introduced limit remain untouched.

### DIFF
--- a/api/desecapi/management/commands/limit.py
+++ b/api/desecapi/management/commands/limit.py
@@ -1,0 +1,38 @@
+from django.core.management import BaseCommand, CommandError
+from django.db.models import Q
+
+from api import settings
+from desecapi.models import RRset, Domain, User
+from desecapi.pdns_change_tracker import PDNSChangeTracker
+
+
+class Command(BaseCommand):
+    help = 'Sets/updates limits for users and domains.'
+
+    def add_arguments(self, parser):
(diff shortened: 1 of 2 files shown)


peterthomassen (approved):
lgtm.

Regarding the CLI interface, perhaps it would be cool to have the interface less implicit, in particular less dependent on the order of arguments: It may be hard to remember whether it should be `manage.py limits ttl example.com 600` or `manage.py limits example.com ttl 600`.

How about something like `manage.py limits --domain=example.com --minimum_ttl=600`? Similar with `--user` etc. Perhaps "only one of `--domain` and `--user` may be given", oder so.

nils-wisiol:
I agree that this would be nicer, but I think it's not worth the implementation effort. Suggestion is to keep it as-is and update/improve if we deem it too complicated/unclear.

peterthomassen:
sounds good.

PR #616, opened 2022-07-29
feat(webapp): sign-up: improve handling of response status 429

peterthomassen:


nils-wisiol:
This should be handled in the error handler in `utils.js`

peterthomassen:
fixed in 3b65917ea2ceb56eb3298d7d4673857a33fcb02a

PR #617, opened 2022-08-11
20220811 remove gitzones

peterthomassen:
This removes Nils' implementation of gitzones-based replication. It had been a lot of work and almost a 1000 lines, but we found that git repo maintenance is not reliable enough in the way we have it set up, and we don't see a straight way out. Too bad! 😢


peterthomassen:
FWIW, I left the improved formatting in api/api/celery.py as well as the faketime stuff in various images. Both may come in handy later.

nils-wisiol:
Oh man, this is hard to let go.

The celery heartbeat thing was a pain to figure out, and we might have uses for it in the future (e.g. collecting delegation status of domains).

peterthomassen:
Yes, I figured :-(

I was thinking the same about Celery heartbeat, but as we don't use it now but comes with extra infrastructure that are not used (in particular, the `celerybeat` volume), I thought it's better to remove it than effectively have it linger unmaintained, and later pose a hazard for something we don't foresee now. As we have this in the commit history, we won't have to reinvent it; it will be easy to restore when needed. What do you think?

peterthomassen:
I fixed the tests, which are broken due to an unrelated problem. (We recently had fixed waiting for Postgres dbapi, but the fix did not apply to one of the GitHub action jobs.) Should work now. :)

PR #618, opened 2022-08-12
chore(api): bump Python to 3.11

peterthomassen:
Preparation for Python 3.11 release (due ~today).


nils-wisiol:
The code looks good to me. However, I'm concerned that it will make development more difficult as I do not have 3.11 locally available. That puts me at risk in running into problems that only occur with 3.11 (like the changes you made); or alternatively spend much time on upgrading my local things.

nils-wisiol:
In particular, I'm concerned about running the api tests locally (like described in README).

peterthomassen:
yeah, same here. It's a "speculative" PR that I just made as I came across things that may be incompatible as I was going over the Python 3.11 changelog. I thought it's better to prepare the upgrade now, than to have to re-read the whole thing.

So, let's just have it sit here until we feel like applying it.

nils-wisiol:
This PR should extend GitHub action such that it runs with 3.11

peterthomassen:
> This PR should extend GitHub action such that it runs with 3.11

Does this require touching the GitHub action? (It seems to me that updating the test container's base image suffices; I've made this change.)

peterthomassen:
@nils-wisiol I think this is ready now!

nils-wisiol (approved):
lgtm

PR #621, opened 2022-08-16
fix(docs): typo

Rotzbua:


peterthomassen:
The repository is set up that a merge requires a successful test run. However, the test run is currently broken due to an external dependency that isn't available (Docker Hub problem or something). I'll rerun it later.

Rotzbua:
The `readthedocs.org` build was successful.

If the `docs` folder is not part of the docker build I would recommend to exclude PRs which only change something in this folder.

peterthomassen:
The test failed again, but that's because the test had a bug which is fixed in another PR (#617). I have moved your commit there. Thank you for your contribution!

peterthomassen:
> The `readthedocs.org` build was successful.
> 
> If the `docs` folder is not part of the docker build I would recommend to exclude PRs which only change something in this folder.

Yeah I agree. The "dumb test structure" just wasn't painful enough so far as to try to sort it out. But you are welcome to ;-)

PR #623, opened 2022-08-22
20220821 prepare 2fa

### LINKED ISSUE #273: api: consider adding e.g. User.action_counter to force invalidation of auth action codes
Otherwise, codes may become valid again if the state is the same again (e.g. by changing an email address back and forth).

Instead of an action counter, we could add User.last_modified or similar.

peterthomassen:
2FA requires that we fix #273, which this PR does. The rest are casual fixes.


api/desecapi/serializers.py
```diff
@@ -284,6 +284,12 @@ def to_internal_value(self, data):
             if not isinstance(item, dict):
                 self.fail('invalid', datatype=type(item).__name__)
             s, t = self._key(item)  # subname, type
+            if not (isinstance(s, str) or s is None):
+                errors[idx].update(subname=f"Expected a string, but got {type(s).__name__}.")
+            if not (isinstance(t, str) or t is None):
+                errors[idx].update(type=f"Expected a string, but got {type(t).__name__}.")
+            if errors[idx]:
+                continue
```
nils-wisiol, line 293: error logic is different from what we have in line 285. If `item` is not a `dict`, then `self.fail` is called. However, for the 'new' error types, `errors[idx]` is updated and the loop is continued. 

(1) Make sure the different behavior is okay/makes sense.
(2) Update comment in line 283.

nils-wisiol: Okay, after seeing the next commit, (1) is done; but (2) should be addressed :-)

nils-wisiol (approved):
nits only!

api/desecapi/management/commands/outreach-email.py
```diff
@@ -54,7 +54,7 @@ def handle(self, *args, **options):
         if options['email']:
             users = User.objects.filter(email__in=options['email'])
         elif content:
-            users = User.objects.filter(outreach_preference=True)
+            users = User.objects.exclude(is_active=False).filter(outreach_preference=True)
```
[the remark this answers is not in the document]
peterthomassen: done

api/desecapi/serializers.py
peterthomassen: done

PR #626, opened 2022-08-24
change(webapp): change to newer font library

Rotzbua:
* `roboto-fontface` is archived
* `fontsource/roboto` provides unicode-range -> smaller font file
* `fontsource/roboto` supports graduated font weight

### DIFF
--- a/www/webapp/.browserslistrc
+++ b/www/webapp/.browserslistrc
@@ -1,2 +1,4 @@
-> 1%
+> 0.5%
 last 2 versions
+Firefox ESR
+not dead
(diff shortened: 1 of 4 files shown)


peterthomassen (approved):
lgtm, thanks!

It's really appreciated that you are making these improvements. The frontend actually is our weak spot (we're not too experienced with JavaScript and JS frameworks), so if you find more stuff that you feel like working on, you are definitely more than welcome! :) (A currently open PR that I haven't gotten around to continue is #616, which improves handling of captcha rate limiting (if you hit the captcha refresh button a lot)).

PR #628, opened 2022-08-24
feat(webapp): update browserslist - just build for active browsers

Rotzbua:
Reduces size of CSS.

Current coverage and simple overview of used `browserslist`:
https://browsersl.ist/?q=%3E+0.5%25%2Clast+2+versions%2CFirefox+ESR%2Cnot+dead

### DIFF
--- a/www/webapp/.browserslistrc
+++ b/www/webapp/.browserslistrc
@@ -1,2 +1,4 @@
-> 1%
+> 0.5%
 last 2 versions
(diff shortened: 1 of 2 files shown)


peterthomassen (approved):
lgtm, thanks!

(For reference, https://browsersl.ist/ says that `defaults` has exactly the rules that @Rotzbua put into `.browserslistrc`.)

PR #629, opened 2022-08-25
fix(webapp): fix CSP header blocking inline font data

Rotzbua:
Webpack inserts small files into the code to minimize the number of requests. This is blocked by the past CSP rule in the browser.

### DIFF
--- a/test/e2e2/spec/test_www.py
+++ b/test/e2e2/spec/test_www.py
@@ -89,7 +89,7 @@ def test_security_headers(api_anon):
     api_anon.headers = {}
     expected_headers = {
         'Strict-Transport-Security': 'max-age=31536000; includeSubdomains; preload',
-        'Content-Security-Policy': "default-src 'self'; frame-src 'none'; connect-src 'self'; font-src 'self'; "
+        'Content-Security-Policy': "default-src 'self'; frame-src 'none'; connect-src 'self'; font-src 'self' data:; "
                                    "img-src 'self' data:; media-src data:; script-src 'self' 'unsafe-eval'; "
                                    "style-src 'self' 'unsafe-inline'; base-uri 'self'; frame-ancestors 'none'; "
                                    "block-all-mixed-content; form-action 'none';",
--- a/www/conf/sites-available/90-desec.static.location
+++ b/www/conf/sites-available/90-desec.static.location
@@ -3,7 +3,7 @@
 #####
 location / {
     add_header Strict-Transport-Security "max-age=31536000; includeSubdomains; preload" always;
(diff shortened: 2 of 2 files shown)


peterthomassen:
I'd like to understand this issue better. Indeed, I can see in Chrome's/Firefox's console that the `data:` URLs are blocked by CSP. I took one of the `data:` URLs, decoded the string and found a Roboto font file. According to other software (e.g. LibreOffice), I don't have that font installed on my system. Still, although it is not installed and the `data:` URL is blocked, it is still used on the web page.

I confirmed that by using the inspector to remove `Roboto` from the `font-face:` CSS property on an element that was using the font. As a result, the rendered font changed to the default `sans-serif` font, which is or looks similar to Verdana.

So I'm confused:
- Why is the font available although it is not installed on the system and the `data:` URL is discarded?
- If this is because it is shipped with browsers: are there any relevant browser that don't ship it? If not, we don't need to ship it.

Rotzbua:
After a second thought, I don't think this is the most targeted solution for a fast website. The new PR addresses the cause of the behavior.

Rotzbua:
To explain what is going on: 

My PR https://github.com/desec-io/desec-stack/pull/626 introduces a new font library.

**The change is:**

The old library uses **one** font file for every font weight (light, normal,...) including all possible charsets (latin, greek, cyrillic, ...) which result in a big font file (~65kb).

The new library has **split** font files for every charset for every font wheight. This results in smaller font size (~15kb).
The browser knows which chars are used on the site and then loads on demand according to the unicode-range the required font.
In the normal case is that just latin charset is used, so only roboto-latin font is loaded.  The advantage is that the traffic is lower and the webpage loads faster on slow networks.

Now there are some smaller font files (<10kb) and here is the problem. 
The build process tries to avoid small files, because those small files may slows down the webpage due connection overhead (more a problem of old http1.1). 
For this reason small files are converted to a dataurl and are shipped within the css file.

In the case of fonts, this is not desirable because the rarely or never used fonts (greek, cyrillic) are included in the CSS that is always delivered. Thus we lose the optimization that just rarely used fonts are loaded only on demand.

peterthomassen:
> * Why is the font available although it is not installed on the system and the `data:` URL is discarded?

I found the answer: It's because in addition to the `data:` URL, there is also a HTTP request to load the full font file. When `data:` fails due to CSP, the font will still be eventually available. -- My confusion arose from my (wrong) assumption that the `data:` URL replaces the network URL, but in fact there are both.

We did some measurements and found that in common setups, the delay between the CSS-based inline rendering and the non-inline rendering based on the network request only is easily on the level of seconds. We thus think that it's better to fix the CSP than to disable inlining. -- Would you mind reopening this PR?

Rotzbua:
The other PR improves the pagespeed. This PR just avoids problems with CSP and inline font. I would also recommend to merge the other PR.

peterthomassen:
> The other PR improves the pagespeed.

I imagine that the total loading time is reduced because the `data:`-encoded stuff is not transmitted, so the size goes down. Is that the improvement you meant, or something else?

PR #630, opened 2022-08-26
fix(webapp): disable inline font

Rotzbua:
Disable inline font to improve FCP (first contentful paint).
Disable to avoid violating CSP rules.

Obsoletes #629

### DIFF
--- a/www/webapp/vue.config.js
+++ b/www/webapp/vue.config.js
@@ -4,4 +4,13 @@ module.exports = {
       allowedHosts: 'all',
     },
   },
+  chainWebpack: config => {
+    config.module
+      .rule('fonts')
+      .set('parser', {
+        dataUrlCondition: {
+          maxSize: 0 // Disable inline font to improve FCP (first contentful paint).
+        }
+      });
+  },
 };


peterthomassen:
Thanks, - I'm not opposed at all to relax the CSP rules, so perhaps that was a misunderstanding. My point in the other PR was that I did not understand what was going on. Do you know?

Rotzbua:
@peterthomassen The explanation is a bit longer, so I tried to avoid it ;)
Here https://github.com/desec-io/desec-stack/pull/629#issuecomment-1228620129 is my explanation I hope it is understandable.

peterthomassen:
Thanks for your explanations and the new PR! :) We looked deeper and would like to close this one in favor of #629.

peterthomassen:
ok, let's talk about this. - I'll only get around to take care of it next week, though.

Rotzbua:
## Performance test

### Setup
* Tested with `vue-cli-service serve --mode production` (gzip http compression enabled by default) 
* Browser Chrome 103

### Results

Lighthouse performance results:

| Metric                   | main    | with pr |
|--------------------------|---------|---------|
| First Contentful   Paint | 5.4 s   | 5.0 s   |
| Time to Interactive      | 5.6 s   | 5.3 s   |
| Speed Index              | 5.4 s   | 5.0 s   |
| Total Blocking Time      | 150 ms  | 160 ms  |
| Largest Contentful Paint | 5.4 s   | 5.0 s   |
| Cumulative Layout Shift  | 0.003   | 0.003   |
|                          |         |         |
| requests                 | 11      | 11      |
| traffic                  | 767 KiB | 696 KiB |

peterthomassen:
We made some improvements as described in https://github.com/desec-io/desec-stack/issues/631#issuecomment-1273192547, but have not deployed this PR yet.

peterthomassen:
I'm still torn between whether a tiny wait is a good trade-off for font flickering, and I'd like to understand the issue better.

Here's where I'm lost:
- Your measurements (https://github.com/desec-io/desec-stack/pull/630#issuecomment-1236191388) indicate that disabling font inlining gains about 0.4s to first contentful paint.
- However, even with inline fonts currently enabled, I do see the font being replaced shortly after first contentful paint, so there is flickering *despite* of font inlining.

That indicates that the inline font is actually not used for the first contentful paint. This can be either
- because the CSS is not loaded -- in which case the FCP time should not be affected at all,
- or because the browser doesn't use it for another reason, in which case the question is why the technology exists at all.

Are you able to shed some light on these questions?

Rotzbua:
> Your measurements (fix(webapp): disable inline font #630 (comment)) indicate that disabling font inlining gains about 0.4s to first contentful paint.

The time reduction comes from the traffic reduction from 767 KiB to 696 KiB.

> However, even with inline fonts currently enabled, I do see the font being replaced shortly after first contentful paint, so there is flickering despite of font inlining.

There is no explicit `font inline` function. It is the generic webpack function: inline tiny files to reduce request to server.

> That indicates that the inline font is actually not used for the first contentful paint.

Yes, you are right.

>  or because the browser doesn't use it for another reason, 

The reason is that not used fonts char-set (greek-ext)  are inlined.

>  in which case the question is why the technology exists at all.

Because webpack makes the decision to inline only by the file size.

peterthomassen:
> > or because the browser doesn't use it for another reason,
> 
> The reason is that not used fonts char-set (greek-ext) are inlined.

I'm rephrasing what I understood: Even with inlining, there is font flickering in the browser because the inlined font is not used (greek-ext), and the fonts that *are* used are *not* inlined. If that's not correct, please correct me.

If it's correct, we should of course not be inlining the font that is not used (move it to a separate request, or ideally remove it entirely), and instead inline the fonts that are used, to get rid of the flickering. Do you agree with this?

Rotzbua:
> I'm rephrasing what I understood: Even with inlining, there is font flickering in the browser because the inlined font is not used (greek-ext charset), and the fonts that are used are not inlined. If that's not correct, please correct me.

I think we are talking at cross purposes. The inline does not directly have anything to do with the font flickering.
The inline technique is to save traffic and requests. This is helps for http1.x which has limited concurrent requests.
Also it saves some unnecessary traffic because for file sizes <2kByte the additional http header are about 1kByte. So there is more overhead and content.

The font flicker is caused by the the css property `font-display: swap`.
This enables the browser to start rendering the page as fast as possible. The browser starts rendering the page even if the custom font is not loaded yet. So the webpage is quickly shown with the minimal amount of traffic (good at low bandwidth).
Later if the custom font is loaded the website is rerendered with the custom font. This causes the flicker.

You can find multiple pro con discussions online about this topic.

One solution is to preload the initial shown fonts with the ` <link rel="preload"`. (Not sure how to do this.)
The other solution is to sacrifice the fast rendering and switch back to the default behavior  `font-display: block`. 

To get back to your question: I do not know what impact the inlining of the font has, but the two above solutions are the common and straight forward options.

PR #632, opened 2022-08-26
Reduce Confusion of Token IDs and Values/Secrets

nils-wisiol:


nils-wisiol:
image

www/webapp/src/views/CrudList.vue
```diff
@@ -197,6 +197,7 @@
                 :key="id"
                 :readonly="column.readonly"
                 :disabled="$store.getters.working || itemIsReadOnly(itemFieldProps.item)"
+                :placeholder="column.placeholder"
```
peterthomassen, line 200: This will also get accessed for column types that don't have that attribute. (You added it to `GenericText` only.) 

I invented `fieldProps` to deal with this case ;-) I'll comment in `TokenList.vue` how that works, but you can remove the line here.

www/webapp/src/views/TokenList.vue
```diff
@@ -110,6 +110,17 @@ export default {
...
+          value: {
+            name: 'item.value',
+            text: 'Token',
+            align: 'left',
+            sortable: false,
+            value: 'value',
+            readonly: true,
+            datatype: 'GenericText',
+            searchable: false,
+            placeholder: "(only displayed once)",
```
peterthomassen, line 113: This line should be replaced by
```javascript
fieldProps: () => ({ placeholder: '(only displayed once)' }),
```

www/webapp/src/views/TokenList.vue
```diff
@@ -18,8 +18,8 @@ export default {
         texts: {
           banner: () => ('<strong>New feature:</strong> You can now configure your tokens for finer access control. Check out the advanced settings switch above!'),
           create: () => ('<p>You can create a new API token here. The token is displayed after submitting this form.</p>'),
-          createBottom: () => ('<p><strong>Warning:</strong> Be sure to protect your tokens appropriately! Knowledge of an API token allows performing actions on your behalf.</p>'),
-          createSuccess: (item) => `Your new token is: <code>${item.token}</code><br />It is only displayed once.`,
+          createBottom: () => ('<p><strong>Warning:</strong> Be sure to protect your token secrets appropriately! Knowledge of an API token allows performing actions on your behalf.</p>'),
+          createSuccess: (item) => `The secret of your new token is: <code>${item.token}</code><br />It is only displayed once.`,
```
peterthomassen, line 22: `Your new token's secret value is:`

docs/dyndns/configure.rst
```diff
@@ -36,11 +36,8 @@ following details in your router configuration:
 - Hostname (same as your username)
 - Token secret (long random string for authentication, displayed after sign-up)
 
-**Advanced API users:** The dynDNS token technically is a regular API token.
-You can also use the token to make requests to our REST API. (Currently, all
-tokens are equally powerful, i.e. a token used for dynDNS updates can also be
-used to perform other kinds of API operations. Token scoping is on our
-roadmap.)
+**Advanced API users:** The dynDNS token technically is a regular API token
+with permissions restricted to DNS management (but not account management).
```
peterthomassen, line 40: These are the tokens that are generated for immediate display when activating an account that includes a dynDNS registration, right?

In that case, I don't think this is true. Or do you mean something else?

peterthomassen (changes requested):

docs/dyndns/configure.rst
nils-wisiol: When an account is activated using `AuthenticatedActivateUserActionView` (`api/v1/v/activate-account/`) and a locally registrable domain is given, a token object is returned in the response. [This token is created using `Token.objects.create(user=domain.owner, name="dyndns")`](https://github.com/desec-io/desec-stack/blob/830e9ba5724ecb6f0137b33118c3d1101c102876/api/desecapi/views/authenticated_actions.py#L167), i.e. the default value of `False` for `perm_manage_tokens` applies.

This makes sense, as we expect this token to end up in some router's storage; why should it be able to manage the deSEC account?

What is a bit awkward is that this business is done by the view class, rather than the action itself.

www/webapp/src/views/TokenList.vue
nils-wisiol: done

www/webapp/src/views/TokenList.vue
nils-wisiol: fixed

www/webapp/src/views/CrudList.vue
nils-wisiol: fixed

docs/dyndns/configure.rst
peterthomassen: > i.e. the default value of `False` for `perm_manage_tokens` applies.

Right! I thought the default was `True`. My bad!

peterthomassen:
rebased, going to merge

PR #633, opened 2022-08-27
feat(): larger common prefixes of Dockerfiles to speed up building

nils-wisiol:
This PR unifies the procedures of some similar Dockerfiles that we have such that they have more common steps, which I hope can reduce the build time and total image size.

This results in some images having unnecessary installations, in particular:

- nslord: iptables iproute2 pdns-backend-pgsql postgresql-client-10
- nsmaster: faketime pdns-backend-mysql

The last 15 builds on GitHub took 478s on average (std: 53s). Let's see how long it takes in this PR. (Data: 6:37 9:07 8:23 7:03 7:26 7:19 7:24 7:08 7:22 7:18 9:11 8:56 8:32 9:32 7:53)

Alternatives to this PR are:

1. Wait (possibly) longer for builds/deployments
2. More docker image layers
  - removing apt package lists requires longer build time for repeated `apt update`
  - not removing apt package lists requires larger layer sizes

### DIFF
--- a/dbapi/Dockerfile
+++ b/dbapi/Dockerfile
@@ -2,13 +2,13 @@ FROM postgres:13-alpine
 
 RUN apk add --no-cache pwgen
 
-ADD docker-entrypoint-initdb.d /docker-entrypoint-initdb.d
-
 USER postgres
 
 # mountable storage
 VOLUME /var/lib/postgresql/data
 
-COPY entrypoint-wrapper.sh /usr/local/bin/
 ENTRYPOINT ["entrypoint-wrapper.sh"]
 CMD ["postgres"]
+
+COPY entrypoint-wrapper.sh /usr/local/bin/
+ADD docker-entrypoint-initdb.d /docker-entrypoint-initdb.d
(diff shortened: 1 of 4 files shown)


nils-wisiol:
Building the new version took 430s.

nils-wisiol:
ping @peterthomassen

peterthomassen (commented):
db* looks cool!

I don't think we should be adding packages to one image that's only needed in the other. The identical part of Dockerfiles should IMO only be the part that both need.

How about restructuring to get rid of some redundant things (like apt-get update). This should still give most of the perfomance gain:
- common: apt-get update, install dnsutils net-tools dirmngr gnupg faketime, pdns-server
- individual:
  - nslord: apt-get install pdns-backend-mysql && apt-get clean && rm -rf ...
  - nsmaster: apt-get install iptables pdns-backend-pgsql postgresql-client-10 && apt-get clean && rm -rf ...

(If installing pdns-server in the first already installs a backend dependecy, then perhaps it has to be moved to the individual stage, too.)

peterthomassen (approved):
lgtm

Issue #631, opened 2022-08-26
Webapp: Pagespeed improvements


Rotzbua:
The pagespeed of the webpage (webapp) could be improved.

## Problem

Currently the reported First Contentful Paint for mobile devices is 10.2 s which is very high.

https://pagespeed.web.dev/report?url=https%3A%2F%2Fdesec.io%2F&form_factor=mobile

## Suggestions by report

* Eliminate render-blocking resources
* Enable text compression
* Reduce unused CSS
* Ensure text remains visible during webfont load

## Comments and possible solutions

Eliminate render-blocking resources:
 * enabling text compression also solves this problem
 * no explicit solution recommendable

Enable text compression:
* can be enabled in the nginx config
* current nginx supports gzip and brotli (better gzip for modern browsers)
* just for for static content, not for API because of security problem BREACH
* biggest impact: **halves the loading time**

Reduce unused CSS: 
* ~90kb by PR #630

Ensure text remains visible during webfont load:
* no action required, because for font can not be improved
* for font `roboto` improvement was done in #626

## Current pagespeed reports
![grafik](https://user-images.githubusercontent.com/7337347/186988381-27924b3e-8baa-485c-9ca3-82b0ef68326f.png)
![grafik](https://user-images.githubusercontent.com/7337347/186988417-b6ae0a26-af22-4afd-867a-3a412b67a442.png)

peterthomassen:
We made a few improvements, in particular adding compression for all static files (unfortunately, the previous deployment only had it enabled for text/html due to nginx config defaults), better webpack chunking, and also client-side caching (doesn't matter on first visit). Metrics from above (which I could reproduce now are):

![image](https://user-images.githubusercontent.com/4242683/194858616-78e6f380-f76e-4ca8-bf3c-bd2f3b71c3a8.png)
![image](https://user-images.githubusercontent.com/4242683/194858806-a088a0a0-85a7-4968-9f57-14675e1a0ef1.png)

We have deployed #630 yet because I'm still torn between whether a tiny wait is a good trade-off for font flickering. So far for the moment.

PR #634, opened 2022-08-27
www: Enable compression for static files

nils-wisiol:
This PR

1. disables redirects from www.dedyn.$DESECSTACK_DOMAIN and dedyn.$DESECSTACK_DOMAIN to desec.$DESECSTACK_DOMAIN, and
2. enables compression for serving static files.

Rationale:

1. These redirects could associate deSEC with malicious content hosted on dedyn domains. Also, the redirects will become a problem should we ever have more than one local public suffix domain -- each one would require separate certificates and nginx config.
2. Speed up webapp loading @Rotzbua #631 I have not enabled compression for API responses out of an unsubstantiated fear that this may break clients.

Merge #408 first.

### DIFF
--- a/test/e2e2/spec/test_www.py
+++ b/test/e2e2/spec/test_www.py
@@ -101,3 +101,23 @@ def test_security_headers(api_anon):
     response = api_anon.get(https_url)
     for k, v in expected_headers.items():
         assert response.headers[k] == v
+
+
+def test_compression(api_anon):
+    # compression for supporting clients on static files
+    assert api_anon.get(
+        f'https://desec.{os.environ["DESECSTACK_DOMAIN"]}/',
+        headers={'Accept-Encoding': 'Accept-Encoding: gzip, deflate'},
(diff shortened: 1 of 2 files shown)


peterthomassen:
I like the static compression!

---

> * These redirects could associate deSEC with malicious content hosted on dedyn domains.

Do we know to which extent this is a problem?

I'm afraid that we loose people who confuse "dedyn" and "desec", and try to reach our site. Here's the number of GET requests on http(s)://dedyn.io/ in the last 7 days: 80, 53, 94, 114, 72, 80, 71 (avg: 81)

> Also, the redirects will become a problem should we ever have more than one local public suffix domain -- each one would require separate certificates and nginx config.

But now, it's not a problem.

nils-wisiol:
Alright, let's drop the disabled redirect for now. This is blocked by #408 anyways so I didn't push anything yet.

nils-wisiol:
Reduced to just compression.

peterthomassen (approved):
lgtm!

PR #635, opened 2022-08-30
feat(www): do not redirect from dedyn names to desec

nils-wisiol:
Mainly for keeping the code around and documenting the decision process. See also #634


peterthomassen:
no planned, but code still around when closed

PR #637, opened 2022-08-31
fix(api): update credentials_changed when (unused) TOTP is created

peterthomassen:
Previously, credentials_changed was only updated when the factor was
first verified. As a result, when an activation link was clicked
twice, the state was unchanged, so the action would attempt to create
the factor again, causing a uniqueness error.

Our `exception_handler` (exception_handlers.py) correctly turned this
into a 409 response, but the ugly database message was exposed, and
an admin email was triggered.

With this commit, the auth action state is invalidated immediately
when the TOTP factor is created.

### DIFF
--- a/api/desecapi/models/users.py
+++ b/api/desecapi/models/users.py
@@ -112,6 +112,11 @@ class User(ExportModelOperationsMixin("User"), AbstractBaseUser):
         logger.warning(f"User {pk} deleted")
         return ret
 
+    def save(self, *args, **kwargs):
+        if kwargs.pop("credentials_changed", False):
+            self.credentials_changed = timezone.now()
+        super().save(*args, **kwargs)
+
     def send_email(
         self, reason, context=None, recipient=None, subject=None, template=None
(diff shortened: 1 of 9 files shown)


api/desecapi/models/users.py
```diff
@@ -112,6 +112,11 @@ def delete(self, *args, **kwargs):
         logger.warning(f"User {pk} deleted")
         return ret
 
+    def save(self, *args, **kwargs):
+        if kwargs.pop("credentials_changed", False):
```
nils-wisiol, line 116: Why not declare this parameter explicitly in the function definition?

nils-wisiol (approved):
LGTM, but the code duplication in the webapp's totp code should be reduced.

peterthomassen:
> LGTM, but the code duplication in the webapp's totp code should be reduced.

Yeah -- planning to fuse the different confirmation flows when adding FIDO2 and two-stage auth actions for MFA.

PR #642, opened 2022-10-07
Multi-IP updates; base58 tokens

### LINKED ISSUE #575: dynDNS: Support multi-IP update in myip= query parameter
Title says it all. Example log:

```
Oct 13 21:49:38 digga desec/www[25547]: update.dedyn.io:443 79.231.19.56 - cmvdo.dedyn.io [13/Oct/2021:19:49:38 +0000] "GET /nic/update?hostname=[...]&myip=2003:[...]:204d,79.[...].56 HTTP/1.0" 400 179 "-" "Deutsche Telekom - Speedport Smart 4 Typ B - 010146.1.1.000.0" TLSv1.3/TLS_AES_256_GCM_SHA384
```

peterthomassen:
Missing:
- [x] final implementation
- [x] tests
- [x] docs

`_find_ip()` currently raises a `ValueError` when contradicting parameters are given (like `&myip=1.2.3.4,,preserve`). Although this could be converted into a DRF exception, it would be nicer to make this part of some serializer validation error. The problem is that the serializer currently used is not aware of the `""` and `"preserve"` special values.

How do proceed? Should we move the whole query parameter handling to a new dynDNS serializer? (In that case, perhaps it's better to abandon this for now. I thought it's a very small change.) @nils-wisiol

### DIFF
--- a/api/desecapi/views/dyndns.py
+++ b/api/desecapi/views/dyndns.py
@@ -36,23 +36,34 @@ class DynDNS12UpdateView(generics.GenericAPIView):
     def throttle_scope_bucket(self):
         return self.domain.name
 
-    def _find_ip(self, params, separator):
+    def _find_ip(self, param_keys, separator):
         # Check URL parameters
-        for p in params:
+        for param_key in param_keys:
             try:
-                param = self.request.query_params[p]
+                params = {
+                    param.strip()
+                    for param in self.request.query_params[param_key].split(",")
+                    if separator in param or param.strip() in ("", "preserve")
+                }
             except KeyError:
                 continue
-            if separator in param or param in ("", "preserve"):
-                return param
+            if len(params) > 1 and params & {"", "preserve"}:
+                raise ValidationError(
+                    detail={
+                        "detail": f'IP parameter "{param_key}" cannot have addresses and "preserve" at the same time.',
+                        "code": "inconsistent-parameter",
+                    }
+                )
+            if params:
+                return [] if "" in params else list(params)
 
         # Check remote IP address
         client_ip = self.request.META.get("REMOTE_ADDR")
         if separator in client_ip:
-            return client_ip
+            return [client_ip]
 
         # give up
-        return None
+        return []
 
     @cached_property
     def qname(self):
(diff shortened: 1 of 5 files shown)


nils-wisiol:
Why not use an exception that will eventually be converted into a 400 status code? Integrating this code better into the DRF structure shouldn't stop us from implementing this feature.

Alternatively, `""` or `"preserve"` could be ignored if other values are given.

api/desecapi/views/dyndns.py
```diff
@@ -36,23 +36,34 @@ class DynDNS12UpdateView(generics.GenericAPIView):
...
+                    if separator in param or param.strip() in ("", "preserve")
+                }
             except KeyError:
                 continue
-            if separator in param or param in ("", "preserve"):
-                return param
+            if len(params) > 1 and params & {"", "preserve"}:
+                raise ValidationError(
+                    detail={
+                        "detail": f'IP parameter "{param_key}" has contradictory values.',
```
nils-wisiol, line 53: Be specific, e.g.: 'IP parameter ... cannot define values and "preserve" at the same time.'

docs/auth/tokens.rst
```diff
@@ -396,14 +396,28 @@ Security Considerations
...
+base58 encoding.
+It is based on a case-sensitive alphanumeric alphabet excluding the characters
+``lIO0`` (hence comprising only the symbols ``a-k``, ``m-z``, ``A-H``,
+``J-N``, ``P-Z``, and ``1-9``).
+This encoding is optimized for maximum clarity and usability:
+Exclusion of certain letters minimizes visual ambiguity, while the restriction
+to alphanumeric symbols allows easy selection (double-click) and input, and
+helps avoid line breaks during display.
+
+Before October 2022, tokens encoded at 21-byte secret using 28 characters in a
```
nils-wisiol, line 414: November :grimacing:

nils-wisiol (approved):
only nits; lgtm!

api/desecapi/views/dyndns.py
peterthomassen: done

docs/auth/tokens.rst
peterthomassen: December xD

PR #645, opened 2022-11-27
Passwordless Authentication

nils-wisiol:
Provide users the option to login using their mailbox. We send an authenticated action that will resolve into a login token.


nils-wisiol:
I've adjusted the API to allow passwordless login via an authenticated action. Webapp modifications are sketched out, but not fully functional yet. 2FA testing needs to be done for the API and for the webapp.

- [ ] 2FA tests for API
- [ ] redirects for webapp
- [ ] webapp: test 2FA support with passwordless login

@peterthomassen let me know what you think! :-)

peterthomassen (commented):
Looked over it, looks nice! Left some comments, but didn't do a real code review.

But why are we doing this? Would you like to have the feature yourself, or are people asking for it? -- The "real" way for passwordless login would be with FIDO2. Could the (potential) demand be met this way?

peterthomassen:
> Don't want to intrude in the work ticket

You are not introducing at all. Community participation is highly welcomed!

> 1. If there are problems with mail servers (your's, the user's etc.) login is broken. Sure, if I can use user/pass as a fallback OK, but why using passwordless login then in the first place if that's my fallback? I had an account that wasn't able to function any other way as to request a "password/token" to login and every damn time that I needed it FAST it had problems with delayed mail and I sat there waiting and waiting for the mail to come in. With mail forwarding services and Anti-Spam measures, that's always one step away from not working (mail was denied with a 500 from the mail server or anything along those lines)

This is especially true as people often run their mail server domain's DNS at deSEC. If the DNS config is broken and logging in requires email access, that's a new failure mode.

> 2. The mentioned mail spamming. That strikes me a bit like SMS/2FA access spam strategies, that would open an account to breaches. And spamming him could sent the mail to black/spam lists.

The login token will be issued to whoever clicks the link, not to the person who issued the email. This is different from 2FA spamming, were the account owner unwittingly approves of a transaction they did not create.

However, one could require another mitigation, e.g. "clicking the login link in the email" requires entering a code that is displayed when trigger the log-in email (i.e. you go to the log-in form, submit email address without password; as a result, an email is sent *and* a code is displayed; you click the link in the email, a form appears in which you have to enter the earlier code, and only if that's right you're logged in).

I share the spam listing concern, but I'd also like to observe that the same issue exists with the password reset functionality.

> Also as an afterthought: relying on the user's mailbox security somehow strikes me as not better / passwordless login but simply using "the same password as email" again. So if a user's login info for his mailbox falls into wrong hands, his desec account is compromised, too.

That's already true for the password reset mechanism. The password adds no layer of security over your mailbox, whether log-in via link is supported or not. To avoid that, you'll have to enable MFA.

I'm mainly worried whether the extra complexity and chance of bugs is worth the benefit.

nils-wisiol:
This is not to replace password-based login or 2FA/MFA.

Because deSEC does not require a password to create an account, we have many accounts that do not have a password. For these accounts, if the user wants to use the web-based UI, they first have to reset their password and can then login. This makes things more complicated than they need to be, especially if it's (thought to be) a one-off thing or if this users was advised by deSEC support to go to the web UI.

Unavailability arguments above apply equally to password reset, and generally we won't make it less available by adding more login methods. What we really should be talking is if this creates an additional attack surface. (I think it does not, simply because obtaining a password reset link is equally hard and essentially has the same effect.)

That being said, I do not have hard data on user needs and just saw somewhere that you can just leave the password field empty to get a login link, which I found neat.

My suggestion towards the password reset / login link spam would be to require a captcha and/or to have an additional field at the user that records the last time we sent an email and with that implement some very basic rate limiting. (But only if this ever becomes a problem.)

Doctormap1 (approved):

PR #647, opened 2023-01-04
fix(webapp): use strict equality

Rotzbua:
Use strict equality for type safety.

### DIFF
--- a/www/webapp/src/components/GenericActionHandler.vue
+++ b/www/webapp/src/components/GenericActionHandler.vue
@@ -20,7 +20,7 @@
       auto_submit: false,
     }),
     async created() {
-      this.auto_submit = this.auto_submit || this.$options.name == 'GenericActionHandler'
+      this.auto_submit = this.auto_submit || this.$options.name === 'GenericActionHandler'
       if(this.auto_submit) {
         this.$emit('autosubmit')
       }
--- a/www/webapp/src/router/index.js
+++ b/www/webapp/src/router/index.js
@@ -179,7 +179,7 @@ router.beforeEach((to, from, next) => {
   } else {
     if (store.state.authenticated) {
       // Log in state was present, but not needed for the current page
-      if (recovered && to.name == 'home') {
+      if (recovered && to.name === 'home') {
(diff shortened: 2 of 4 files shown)


www/webapp/src/components/Field/Record/DNSKEY.vue
```diff
@@ -10,7 +10,7 @@ const int8 = between(0, MAX8);
 const MAX16 = 65535;
 const int16 = between(0, MAX16);
 
-const equals3 = (value) => !value || value == 3;
+const equals3 = (value) => !value || value === 3;
```
peterthomassen, line 13: Is this correct? -- The input field value is expected to be a string, so `=== 3` should always be false.

peterthomassen (changes requested):
I think one of the changes causes problems (see comment at the respective code location).

All other changes are about comparisons where operand types are not under user control (e.g. the `.length` attribute of an array), so type safety is implicit because the operand is type-safe.

While I see that this change could be made, I'm not sure this is worth merging. -- I'm happy to be educated about the actual impact, or about a recent development where JavaScript folks feel that these changes are important.

Also, @nils-wisiol what do you think?

Rotzbua:
Thanks for the feedback. I reduced the changes to just type `String`. As you pointed out the compare with integer need way more investigation.

nils-wisiol:
Thanks for this work @Rotzbua! 

I believe less linter warnings are good, also because we do not have any automated tests for the webapp. Maintaining good code style will (a) produce less error-prone code and (b) unclog the linter output to see things that actually matter.

nils-wisiol:
Is there a way to integrate the linter in out GitHub actions? I do not think we can require zero warnings as of now. So ideally, PRs would be required to not produce more warnings than the code in main :)

nils-wisiol (approved):

nils-wisiol:
@peterthomassen this needs your approval for merging

peterthomassen (approved):

Rotzbua:
@nils-wisiol Unfortunately I don't know any tool which fulfills this. Otherwise I would have already created a PR ;)

PR #648, opened 2023-01-04
chore(webapp): bump dependencies

Rotzbua:
The eslint bump introduces new rules which throw errors. Changed them temporary to warning because they do not cause problems.


peterthomassen (approved):
lgtm, thanks!

PR #652, opened 2023-01-17
feat(CI): add dependabot to create update PRs for GH action and `npm`&`pip` dependencies

Rotzbua:
Let the bot do the chore of checking for updates.


peterthomassen (changes requested):
Good idea!

I think there should be several `package-ecosystem` directives, such as for `pip` (see https://github.com/rthalley/dnspython/blob/master/.github/dependabot.yml) and `npm`. We also should change to daily checks.

Rotzbua:
>  think there should be several package-ecosystem directives, such as for pip

I excluded them on purpose. One reason is that both systems `pip` and `npm` offer an useful and easy to use local dependency management. As far as I know there is no simple `gh update actions`.

Dependabot can also be very annoying as it creates a new PR for every single dependency. This can tie up more resources than simply updating the dependencies locally once a month and then committing.

But to be honest, I'm still missing experience which `pip` and `npm` settings of Dependabot generate a real value for development.

>  We also should change to daily checks.

Github Actions do not change very often. So I think a monthly check is sufficient and will not waste resources like an daily check.

nils-wisiol:
I like this idea, but believe that there is little added value for checking our GitHub Action dependencies. Skimming through the docs and our actions, I think the [only thing that will be monitored is the checkout action](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) (currently version 2).

I would love to see weekly checking of npm and pypi dependencies via dependabot, though. Is this something you would consider doing in this PR, @Rotzbua ?

Rotzbua:
Result will be:
https://github.com/Rotzbua/desec-stack-test/pulls

To clearify `dependabot.yml` does only configure the *version updates*.
The *Dependabot alerts* and the *security updates* must be enabled in the settings and can not be customized.
https://github.com/desec-io/desec-stack/security
The alerts are only visible for maintainers.

peterthomassen (approved):

peterthomassen:
Thanks! Can you please sign your commit?

You might want to set `git config commit.gpgsign true` to have this enabled permanently.

Rotzbua:
It was signed, but I rebased the branch with the Github function and it seems that it removes the signature. I will sign them later.

Rotzbua:
@peterthomassen Rebased and signed 🚀

Issue #246, opened 2019-09-26
Enforce IPv4 For Dyn Update?


nils-wisiol:
In a dual stack setup, how can I use the update.dedyn.io "dyndns12" API to reliably update the dyn DNS IP address to my v4 address? We provide this functionality for IPv6 via the `update6.dedyn.io` hostname, however, it appears there is no `update4.dedyn.io`. Is there a good reason we don't offer that?

fabricionaweb:
If you are using cURL you can enforce the request to be ipv4 using the `--ipv4` flag. The `update.dedyn.io` will get your ipv4 even if you have ipv6 ip.

PR #653, opened 2023-01-20
feat(docs):  add information and example to enforce cURL to use ipv4

fabricionaweb:
It has happened to me, and I notice it might have happened to others already #246 

I have an active IPv6 connection and when trying to update my record in a single call to `update.dedyn.io` (without passing my current ip as param), it assumes my IPv6 and it turns the record into `AAAA`. 

Instead of passing my current ip as param, we can pass the [curl option --ipv4 or --no-ipv6](https://curl.se/docs/manpage.html#-4) to make use of IPv4 only

Im proposing to add some information and an example about it to the docs.


peterthomassen (approved):
Thanks! I made some minor adjustments.

PR #656, opened 2023-01-26
refactor(webapp): migrate store from `vuex` to `pinia`

Rotzbua:
Changed:

* refactor init order so pinia is available in router

Checklist:

* [x] add pinia
* [x] remove vuex
* [x] compiles
* [x] first run test in browser
* [x] extended testing in browser (login, logout, auth)


Rotzbua:
Could not observe problems. Since it is a core change it would be nice if somebody also test it in browser.

Rotzbua:
Solved rebase conflict.

peterthomassen (commented):
Looks pretty good! Don't see any problems, but will have to test the various state variables before merging.

One question: Is it not possible to pass around the return value of `useUserStore()` across components, instead of re-instantiating it everywhere? (And in `utils.js`, it is even instantiated multiple times in the same file. Is that necessary?)

www/webapp/src/store/user.js
```diff
@@ -0,0 +1,53 @@
+import {defineStore} from 'pinia';
+
+// export interface User {
```
Rotzbua, line 3: should be removed..

Rotzbua:
As I understand it the `useUserStore()` is just a getter to the store. The instantiation is just done in the `main.js`.

peterthomassen:
I meant, we take `useUserStore` from `user.js` and then set `user = useUserStore()` in each file or function where the store is used. Why not do `export user = useUserStore()` in `user.js`, and then import `user` directly, instead of calling `useUserStore()` all over the place?

Rotzbua:
> Why not do export user = useUserStore() in user.js

This will cause an error:

```
Uncaught Error: [🍍]: getActivePinia was called with no active Pinia. Did you forget to install pinia?
	const pinia = createPinia()
	app.use(pinia)
This will fail in production.
```

Because `pinia` is initialized in `main.js` with https://github.com/desec-io/desec-stack/blob/4bf44c379a8e2627bde6010241d8c229f6ea97fb/www/webapp/src/main.js#L25

see doc: https://pinia.vuejs.org/core-concepts/outside-component-usage.html#single-page-applications

The difference behavour between vuex and pinia is, that vuex creates all stores at the start of vue and pinia only creates it when it is used.

see doc: https://pinia.vuejs.org/core-concepts/#using-the-store

peterthomassen (approved):
lgtm! thank you very much.

PR #659, opened 2023-01-26
fix(www): fix wrong html tag, doctype, javascript

Rotzbua:
* search engines should not index error pages
* replace obsolete html with css
* fix invalid html

Checked by:
https://validator.w3.org/#validate_by_input

### DIFF
--- a/www/html/429.html
+++ b/www/html/429.html
@@ -1,7 +1,13 @@
-<html>
-<head><title>429 Too Many Requests</title></head>
-<body bgcolor="white">
-<center><h1>429 Too Many Requests</h1></center>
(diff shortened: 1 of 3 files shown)


nils-wisiol (approved):
lgtm!

@peterthomassen, I cannot find where `check.html` is still used. Is that dead code?

Rotzbua:
@nils-wisiol it is used in https://github.com/desec-io/desec-stack/blob/cd086dda09bbc27b82f9d18b320bb5a63863fd3b/www/conf/sites-available/05-dedyn.conf.var#L34

PR #661, opened 2023-01-26
feat(webapp): add robots.txt for SEO

Rotzbua:
Avoid indexation and robot rage of `/api/`.

## Background

Some robots may find a link `domain.or/api/xyz` and then try to index the complete api causing a pseudo DOS attack. Good robots will check the robots.txt and stop doing this. Evil robots will do this anyway.

## Reference

* https://web.dev/robots-txt/


peterthomassen (approved):
thanks!

Issue #662, opened 2023-01-26
Webapp: overview lint warnings/errors


Rotzbua:
With the newer eslint versions new inspections were added.
The warnings/errors can cause problems with newer package versions and future migrations.
So they should be solved or if false warnings suppressed.

Current output to fix:

* [x] www\webapp\src\components\Field\Code.vue
  7:9  warning  Name "Code" is reserved in HTML  vue/no-reserved-component-names

* [x] www\webapp\src\views\About.vue
  39:29  warning  Using v-text on component may break component's content  vue/no-v-text-v-html-on-component

* [x] www\webapp\src\views\Console\Confirmation.vue
  3:14  error  Unexpected mutation of "value" prop  vue/no-mutating-props

* [x] www\webapp\src\views\CrudList.vue
  375:5  warning  Name "Code" is reserved in HTML  vue/no-reserved-component-names

* [x] www\webapp\src\views\PrivacyPolicy.vue
  13:29  warning  Using v-text on component may break component's content  vue/no-v-text-v-html-on-component

* [x] www\webapp\src\views\Terms.vue
  35:26  warning  Using v-html on component may break component's content  vue/no-v-text-v-html-on-component

Add strongly-recommended:

* [x] `vue/v-bind-style`
* [x] `vue/v-on-style`
* [x] `vue/v-slot-style`

PR #664, opened 2023-01-26
fix(webapp): warning vue/no-v-text-v-html-on-component

Rotzbua:
* fix warning Using v-html on component may break component's content vue/no-v-text-v-html-on-component
* add pure html tag to insert html
* fix warning Using v-text on component may break component's content vue/no-v-text-v-html-on-component
* set the "normal" way

## Reference

Mentioned in #662


nils-wisiol (approved):
lgtm

PR #668, opened 2023-01-30
feat(webapp): add more lint rules

Rotzbua:
Introduce and fix some eslint rules from `strongly-recommended`.
Enforces consistent slot/bind/on style in templates.

## Reference

Mentioned in #662


peterthomassen (approved):
lgtm, thanks!

FWIW, I think we didn't use the `#` shorthand anywhere, which is [declared "good" style](https://vuejs.org/style-guide/rules-strongly-recommended.html#directive-shorthands). (It's only "bad" to mix `#` with `v-slot:`.) Seems like the rule does not account for this (which is no problem, just seemed peculiar).

Rotzbua:
Indeed mixing styles is bad. It is possible to configure `v-slot:` as default style if you wish.

PR #669, opened 2023-01-30
remove(webapp): unused component

Rotzbua:
* component not used and causes lint error `Unexpected mutation of "value" prop  vue/no-mutating-props`

## Reference

Mentioned in #662


peterthomassen (approved):
lgtm, thanks!

PR #671, opened 2023-01-31
feat(api): add metrics for RRsetSerializer

peterthomassen:


api/desecapi/serializers/records.py
```diff
@@ -492,10 +494,16 @@ def _validate_length(self, attrs):
...
+            qs = models.BlockedSubnet.objects.values_list("subnet", flat=True).order_by(
+                Masklen(F("subnet")).desc()
+            )
             for record in attrs["records"]:
-                if models.BlockedSubnet.objects.filter(
-                    subnet__net_contains=record["content"]
-                ).exists():
+                subnet = qs.filter(subnet__net_contains=record["content"]).first()
+                if subnet:
+                    print(str(subnet))
```
nils-wisiol, line 503: remove debug output

api/desecapi/serializers/records.py
```diff
@@ -492,10 +494,16 @@ def _validate_length(self, attrs):
...
+                    metrics.get(
+                        "desecapi_records_serializer_validate_blocked_subnet"
+                    ).labels(str(subnet)).inc()
```
nils-wisiol, line 506: This increases the counter n times if a single API request with an RRset with n records is made. Is that desirable for the resulting statistic?

Related question, are we interested in annotating this with the domain name?

nils-wisiol (changes requested):
some discussion needed

api/desecapi/serializers/records.py
peterthomassen: done

api/desecapi/serializers/records.py
peterthomassen: > This increases the counter n times if a single API request with an RRset with n records is made. Is that desirable for the resulting statistic?

No, because the first one raises.

> Related question, are we interested in annotating this with the domain name?

Why? (Can't think of why we would want to extract that from Grafana.)

nils-wisiol: Woops. :)

nils-wisiol (approved):
lgtm

PR #672, opened 2023-02-01
feat(prometheus): increase retention to 1 year

peterthomassen:
It would be nice to observe trends for longer than one month. - This is just event metrics, no personal(izable) data.


nils-wisiol (approved):
lgtm

PR #676, opened 2023-02-08
feat(webapp): add svg favicon

Rotzbua:
## Advantage

* Independent of screen resolution
* Modern image format

## Changes

* `favicon.svg` copy of `logo.notext.svg`
* Removed white border from favicon. logo is black, the white frame disturbs with colored background

### DIFF
--- a/www/webapp/public/favicon.svg
+++ b/www/webapp/public/favicon.svg
@@ -1,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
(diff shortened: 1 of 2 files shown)


peterthomassen (approved):
lgtm! Do you think we should also add an `apple-touch-icon` variant? (see https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#sect3)

Rotzbua:
Not sure about the current Apple implementation. The referenced documentation is from 2016. I do not own Apple devices..

peterthomassen:
Can you please sign your commit? thanks.

PR #677, opened 2023-02-08
fix(webapp): fix naming

Rotzbua:
Some names are wrong. Has no impact.

### DIFF
--- a/www/webapp/src/views/About.vue
+++ b/www/webapp/src/views/About.vue
@@ -198,7 +198,7 @@
 
 <script>
 export default {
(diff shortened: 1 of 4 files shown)


peterthomassen (approved):
sure, lgtm. Although I don't know why one case would be better than other :) I guess for consistency.

Rotzbua:
The plan is to solve the essential lint rule `vue/multi-word-component-names`. That would cause file renaming and possibly have conflicts on other PRs. So rather a small reviewable step before the larger step comes 😉

Issue #649, opened 2023-01-06
Webapp: dependency status overview


Rotzbua:
With PR #648 I got in touch with some dependencies.
I think a short overview is useful:

* `vue` framework
  * project `2`, current `3`
  * `2` eol date: 31 Dec 2023
  * migration needs a lot of refactoring + incompatible plugins
* `@vue/cli-service` compiler
  * replaced by `vite`
  * plugins must be compatible
* `vuetify` ui
  * project `2`, current `3` only usable with vue3
  * `2` depends on old `sass@~1.32` otherwise a lot of deprecation msg
  * `2` has problems with `vite`
* `axios` api requests
  * project `0.27`, current `1.2`
  * `0` still maintained, `1.2` not so stable
* `vuex` store
  * project `3`, current `4` (just compatible with vue3), both replaced by `pinia` https://pinia.vuejs.org/
  * `pinia` supports both `vue` (2&3), typescript support
  * migration possible
  * PR in queue
* `vuelidate` validation
  * project `0.7.7`, current `2`
  * `@vuelidate/*` supports both `vue` (2&3)
  * migration possible
* `cypress` test
  * tests are not implemented
  * should be removed

peterthomassen:
Thank you! Your help with this is much appreciated. Would you be up for addressing these as you see fit?

PR #699, opened 2023-03-11
Upgrade to Vue 3

lukaslihotzki:
Refs #649

There are still some issues with the upgrade to Vue 3 that are not solved in this draft. Some of them can be solved independently of the upgrade:
- [x] vue-router 4 (part of Vue 3) does not support routes to other origins. We should probably use simple `a href` links instead of `router-link` for those.
- [ ] Vuelidate 2 (new version compatible with Vue 2 and Vue 3) adds nested validators that should be used instead of the now unsupported `$each`.
- [ ] `npx depcheck` returns many other unused dependencies. Can they be removed safely?

The other issues need to be fixed in this draft:
- [ ] The new copy to clipboard implementation is untested.
- [ ] The style is broken, probably because there are a lot of eslint errors that cannot be fixed automatically.

I don't exactly understand how vue-demi should be used. If there are problems, use `npx vue-demi-switch 2` or `npx vue-demi-switch 3` depending on the version you currently want to use. Clearing node_modules can also help.

I `.eslintignore` everything, because `npm run serve` refuses to run otherwise. This should be removed if needed.

The commit [chore(webapp): update to vue 3](https://github.com/desec-io/desec-stack/commit/eaee43f5bba881b8ea92dc7df97b0805ee66868e) was made by changing `.eslintrc.js` and running this in www/webapp:
```
npm remove --save vue-template-compiler vuetify-loader vue-cli-plugin-vuetify
npm install --save vue@3 vuetify@3 vue-router@4 webpack-plugin-vuetify @vue/compiler-sfc
npm install --save-dev vue-demi eslint-plugin-vuetify
npx vue-demi-switch 3
./node_modules/.bin/eslint --fix 'src/**'
```


Rotzbua:
Nice work 👍 
I would recommend to migrate `Vuelidate` in a separate PR before upgrading `vue`.

peterthomassen:
@lukaslihotzki Thanks for the work! Are you still in for moving it on? :)

peterthomassen:
@Rotzbua @lukaslihotzki Hey... long time no chat! Are you interested in finishing this migration?

peterthomassen:
We're looking for paid help in resolving this - are you interested, @Rotzbua? Anyone else? @lukaslihotzki

peterthomassen:
done in #1173

PR #702, opened 2023-03-14
fix(api): allow underscores in CNAME targets

peterthomassen:
The previous check was based on the wrong pdns code location. Consultation on pdns IRC revealed the right one.

Fixes regression from 0605f3879012ce6dc4c06f5d7711875348c9e468


nils-wisiol (approved):
lgtm

PR #705, opened 2023-03-22
feat(www): improve noscript for better SEO and text browsers

peterthomassen:
upon a user's request

### DIFF
--- a/www/webapp/public/index.html
+++ b/www/webapp/public/index.html
@@ -9,7 +9,21 @@
   </head>
   <body>
     <noscript>
-      <strong>This page does requires JavaScript to function. Please enable it to continue.</strong>
+      <img src="assets/logo.svg" alt="deSEC"/>
+      <h1>Modern DNS Hosting for Everyone</h1>
+      <p>
+        deSEC is a <strong>free DNS hosting</strong> service, <strong>designed with security in mind</strong>.<br/>
+        Running on <strong>open-source software</strong> and <strong>supported by <a href="https://securesystems.de/">SSE</a></strong>,
+        deSEC is free for everyone to use.
+      </p>
+      <h2>🛑 Warning</h2>
+      <p>
+        deSEC requires JavaScript for accessing the control panel for DNS management.<br/>
+        Please enable JavaScript for this site to proceed.
+      </p>
+      <p>
+        Hint: Accessing the <a href="https://desec.readthedocs.io/">API documentation</a> does not require JavaScript.
+      </p>
     </noscript>
     <div id="app"></div>
     <!-- built files will be auto injected -->


peterthomassen:
> Maybe we also want to include a link to the docs here. After all, those can be used without JS :)

done

> Rather than adding a `<br>`, it would be better to wrap this entire block in a `<div class="NoScript">` (if you add a class to the `<noscript>` rather a new block, you can have issues with how uBlock Origin and other script blockers behave) and use CSS `.NoScript { max-width: 66ch; }` and it can be easily centered and padded `.NoScript { margin: auto; padding: 1em; }`. This doesn't cause weird or unnecessary breaks on small viewports.

I'm not sure about the benefit of that styling; the way w3m displays this here is totally fine IMO:
image

Also, we don't know what's a reasonable max-width for people -- not everyone with JS disabled is working on the terminal. It think it's best to leave this to the user agent to sort out.

> The first two paragraphs can and probably should be combined as they're part of the same description. If line breaking was the concern, see above.

Thanks, done.

> The warning falls under the category of "admonition". This should be sectioned separately because it's not a part of the main document

I'm not sure why that wouldn't be part of the document.

> and the `<h2>` makes it seem like it a part of the main document which make it be parsed as content by some crawlers. I believe a `<table>` with a row of two `<td>` elements may be better suited though--not for semantics reasons, but it's common to use as it will get default user agent styling that make it _pop out_ to the user with the box around it which is great especially for TUI browsers ([Asciidoctor defaults to this](https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/)).

I tried that when I made this PR, and neither w3m nor links display any lined boxes in that case. To keep it simple and semantically logical, I dropped the table.

> Otherwise, something like `<section class="Admonition">` as a wrapper make help the semantics be more apparent ([`asciidoctor-html5s` alternative back-end](https://github.com/jirutka/asciidoctor-html5s/blob/master/data/templates/admonition.html.slim) goes this route).

I researched for a spec for how to do that, but couldn't find any. Can you point to any normative document that assign meaning to such classes? Otherwise, I'm afraid it's pretty arbitrary / proprietary behavior (and the above screenshot already looks fine).

> Personally, I think it's a good idea to reference "or make an exception in your script blocker" in the warning/danger admonition body to cover the reality that a lot of the folks that would see this error, are likely privacy & security conscious. This acknowledgment is better for trust because these users have deliberately disabled JavaScript and know it's disabled.

I agree with you, but think it's better to make a minimal statement, which is "enable JavaScript for this site" (note the second half of this, which *is* respectful for the user's privacy decisions on the remaining Internet).

toastal:
> w3m + width

`elinks` (and probably `links2`) support a max width in their browsers. If the user wants longer lines, supporting them is fine, else on small terminals you can have double breaks. Usually users will have an emulator open to their favorite size, or have a multiplexer split inside it to the desired width. `less` can control this as well, but I don't know that it's the site's job to impose a forced layout with breaks over the user agent's engine.

> part of the document

Apologies, I mean it's not a part of the "main" document. It's an aside/callout which is tangent, but not the main content.

> I tried that when I made this PR, and neither w3m nor links display any lined boxes in that case. To keep it simple and semantically logical, I dropped the table.

Well there you have it. Keep it as is.

> I researched for a spec for how to do that, but couldn't find any.

Right, it's all stringly-based semantics now where everyone is sort of making up their own thing (though I _do_ see `(A|a)dmonition` commonly and a tool could be scraping for it unsafely because it is common). I truly wish there was an element 😅

> enable

Sure. If you think it's better, then go for that. The actionable message which what went wrong + what you're missing is the important part.

Cheers!

PR #708, opened 2023-03-28
Add eslint plugin for vuetify 

Rotzbua:
### Changes

* Add eslint plugin for vuetify 
* Lint:fix: Migrate deprecated/outdated vuetify classes

Left fixed css classes, right current.

grafik


nils-wisiol:
I generally like the outlines, but this leaves the style of the password change form inconsistent, see below. Similar for delete account.

I've checked other places, in particular the DNS management parts, they look fine (but no outlines).

image

nils-wisiol (changes requested):
as commented above

peterthomassen:
We don't have outlines in the DNS record editor (behind login), and we can't change that because it makes heavy use of combining several input fields into a row that looks like a single one. (That would break with outlines.)

So for consistency, I think we shouldn't have outlines anywhere. The absence of outlines also doesn't seem to have been a UX problem.

nils-wisiol:
I don't think we need consistency across all forms, and I personally like the fresh look of the more prominent forms for login and registration. Just as a chance. I agree that the DNS management can't and shouldn't use outlines and merely meant that it should be consistent within the password change and delete account form.

peterthomassen:
sure, I don't feel strongly.

peterthomassen (approved):
lgtm, rebased and bumped eslint-plugin-vuetify version

Issue #710, opened 2023-04-08
Replace icon font `@mdi/font` 


Rotzbua:
## Problem

Currently the icon webfont is used. The webfont includes every available icon and is therefore very big with around 400kb.

## Solution

This can be avoided by using tree shakeable `@mdi/js`.

## Reference

* https://v2.vuetifyjs.com/en/features/icon-fonts/#material-design-icons-js-svg
* https://pictogrammers.com/docs/guides/webfont-alternatives/ (incomplete, not up to date)

PR #711, opened 2023-04-08
Migrate from `@mdi/font` to `@mdi/js`

Rotzbua:
Use tree shakeable version instead of big webfont.

Login, signup seems to need some refactoring. Migration later.

## Reference

Partial migration of #710


nils-wisiol (approved):
lgtm!

PR #712, opened 2023-04-08
feat(webapp): lazy load image

Rotzbua:
## Changes

* Change: lazy load image
* Fix image alt text

### DIFF
--- a/www/webapp/package.json
+++ b/www/webapp/package.json
@@ -11,6 +11,7 @@
   "dependencies": {
     "@fontsource/roboto": "^4.5.8",
     "@mdi/font": "^6.9.96",
+    "@mdi/js": "~7.2.96",
     "axios": "^0.27.2",
     "core-js": "^3.27.1",
     "javascript-time-ago": "^2.5.9",
--- a/www/webapp/src/App.vue
+++ b/www/webapp/src/App.vue
@@ -70,7 +70,7 @@
                       v-on="on"
               >
                 more
-                <v-icon right>mdi-menu-down</v-icon>
+                <v-icon right>{{mdiMenuDown}}</v-icon>
               </v-btn>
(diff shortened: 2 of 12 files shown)


nils-wisiol (changes requested):
I've applied these changes but it does not lazy-load. I've confirmed that the markup is shown in the browser; but everything is loaded on page load.

nils-wisiol:
Also, please use the `feat(webapp):` prefix for this commit

Rotzbua:
@nils-wisiol You can throttle the network speed in the dev console to see the effect.

A good overview is: https://web.dev/browser-level-image-lazy-loading/

nils-wisiol:
The reason I couldn't see the lazy-loading was that I used Firefox for testing. In Firefox, the only lazy-loading I see is for worldmap and worldmap marker, which is not part of this PR. In Chrome, the four logos and worldmap, worldmap marker are lazily loaded. Not sure why, they are not marked for lazy loading.

Anyways, with this PR it's better than without. Lgtm!

@Rotzbua if you want to debug the Firefox issue and/or shed some light on why the worldmap is lazily loaded, we can hold off on merging. Otherwise let's merge this soon. Waiting for your preference.

nils-wisiol (approved):

Rotzbua:
@nils-wisiol found the issue: order of the attr

Seems to be an ancient Firefox bug 😠 :

   * If you define the "loading" attribute at "lazy" AFTER the "src" attribute, it has no effect (all images are loaded, even those not in the viewport).
   * If you define the "loading" attribute at "lazy" BEFORE the "src" attribute, it it applied as expected (only images in the viewport are loaded).

Source:
https://bugzilla.mozilla.org/show_bug.cgi?id=1647077

PR #713, opened 2023-04-08
Fix html validator info messages

Rotzbua:
## Change 

Fix html validator info messages:

> Info: Trailing slash on void elements has no effect and interacts badly with unquoted attribute values.

https://validator.w3.org/nu/?doc=https%3A%2F%2Fdesec.io%2F


nils-wisiol (approved):
lgtm

PR #729, opened 2023-06-06
20230606 bumps

peterthomassen:


peterthomassen:
This now does not include Python 3.11 because not all required Python packages are yet compatible. Still, this should not keep us from keeping our app up to date.

nils-wisiol (approved):
lgtm

PR #738, opened 2023-06-12
fix(webapp): migrate deprecated `substr()`

Rotzbua:
## Problem

`substr()` is deprecated.

## Solution

Replace by `substring` or `slice`.

Used `slice` because it is used in the project.

## Reference

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

### DIFF
--- a/www/webapp/src/components/Field/RecordItem.vue
+++ b/www/webapp/src/components/Field/RecordItem.vue
@@ -228,7 +228,7 @@ export default {
 
       event.preventDefault();
       const pos = this.getPosition();
-      this.update(this.value.substr(0, pos - 1) + this.value.substr(pos), pos - 1);
+      this.update(this.value.slice(0, pos - 1) + this.value.slice(pos), pos - 1);
     },
     deleteHandler(index, event) {
       if (!this.positionBeforeDelimiter(index)) {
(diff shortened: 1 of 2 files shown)


www/webapp/src/components/Field/RecordItem.vue
```diff
@@ -228,7 +228,7 @@ export default {
 
       event.preventDefault();
       const pos = this.getPosition();
-      this.update(this.value.substr(0, pos - 1) + this.value.substr(pos), pos - 1);
+      this.update(this.value.slice(0, pos) + this.value.slice(pos), pos - 1);
```
peterthomassen, line 231: This is not equivalent to the old code:
```js
> 'test123'.substr(0, 4 - 1) == 'test123'.slice(0, 4)
false
```

www/webapp/src/components/Field/RecordItem.vue
```diff
@@ -237,7 +237,7 @@ export default {
 
       event.preventDefault();
       const pos = this.getPosition();
-      this.update(this.value.substr(0, pos) + this.value.substr(pos + 1), pos);
+      this.update(this.value.slice(0, pos + 1) + this.value.slice(pos + 1), pos);
```
peterthomassen, line 240: ditto

peterthomassen (changes requested):

peterthomassen (approved):

PR #691, opened 2023-02-19
refactor(webapp): fix lint rule `vue/multi-word-component-names`

Rotzbua:
## Changes

Fixes all the problems with vue essential lint rule `vue/multi-word-component-names`.

## Advantage

> This prevents conflicts with existing and future HTML elements, since all HTML elements are a single word.

Source: https://eslint.vuejs.org/rules/multi-word-component-names.html

Prevents unforeseen side effects.

## Reference

Part 2 of merged #677.

PR #740, opened 2023-06-14
fix(webapp): RRset value field doesn't update

### LINKED ISSUE #739: RRset Value Field doesn't Update with Type
When creating a new RRset, the value input field isn't updated with the selected type of RRset anymore. This means that it is harder for the user to input a valid value. (But not impossible.) Similarly, when editing RRsets, the value shown isn't properly annotated.

This looks like it happened when #713 was deployed, @Rotzbua would you be able to take a look? My gut-feeling is that some renaming has gone wrong.

In the screenshot, MX is selected, so there should be separate fields for priority and hostname, but there is only one.

image

Rotzbua:
Fix bug introduced by #691 .

Closes #739 .

### DIFF
--- a/www/webapp/src/components/Field/RecordList.vue
+++ b/www/webapp/src/components/Field/RecordList.vue
@@ -143,8 +143,9 @@ export default {
   },
   methods: {
     getRecordComponentName(type) {
-      const genericComponentName = 'RecordItem';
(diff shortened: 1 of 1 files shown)


peterthomassen (approved):
thanks!

PR #744, opened 2023-06-26
feat(www): add NLnet Foundation credits

peterthomassen:


nils-wisiol (approved):
For the first commit, I did not run the webapp to verify the layout.

For the second commit, eventually we should freeze Python dependencies as part of the build process.

PR #745, opened 2023-06-27
feat(webapp): replace dependency `vue-timeago`

Rotzbua:
As @lukaslihotzki mentioned in #699 the dependency `javascript-time-ago` is not used.

The time ago component `vue-timeago` is a wrapper for `date-fns`. The component is just used once and limited functionality so it could be implement very easy. This simplifies the dependency graph and for the migration to vue3 there is no need for replacement.


peterthomassen (approved):
thank you very much - lgtm!

PR #746, opened 2023-07-01
feat(webapp): use component name instead of string

Rotzbua:
## Problem

Renaming the components in pr  #691 caused runtime issues because constant strings are used.

## Solution

Use component name directly.


peterthomassen (approved):
lgtm, thanks!

PR #748, opened 2023-07-02
Remove reset password

### LINKED ISSUE #747: Password reset should not be shown if logged in
## Problem

If you are logged in you have the password :)
grafik

## Solution

Password reset should not be shown if logged in.

Rotzbua:
Password reset should not be shown if logged in.

## Reference

Fixes #747


peterthomassen (approved):
lgtm, thanks! :)

PR #749, opened 2023-07-04
fix(webapp): disable complete form not every input field

Rotzbua:
## Change

Disable form instead of every input field.

## Reason

Vuetify did not support the disable-forms-feature from the beginning. Today it supports to disable the form instead of every single form input.

## Fixes

Signup: domain selection is not disabled.

## Reference

https://github.com/vuetifyjs/vuetify/releases/tag/v2.3.0


peterthomassen (approved):
lgtm, thank you very much!

PR #758, opened 2023-07-11
feat(webapp): add eslint rule for mustache style

Rotzbua:
Two styles are used in project:

* `{{abc}}`
* `{{ abc }}`

With spaces `{{ abc }}` is eslint default and forced by this new rule.

## Advantage

Style can be automatically applied by `npm lint --fix`.

## Reference

https://eslint.vuejs.org/rules/mustache-interpolation-spacing.html


peterthomassen (approved):
lgtm

PR #759, opened 2023-07-12
refactor(webapp): replace `vue-clipboard2`

Rotzbua:
From #699 as preparation for vue3. Native implementation is simpler than an additional dependency.


peterthomassen (approved):
lgtm, thanks!

PR #760, opened 2023-07-13
feat(webapp): preparation for vue3

Rotzbua:
Some APIs are deprecated with vue3. A few of them can be migrated despite of vue2.
Added `eslint` rules to prevent future changes adding them back in.

### DIFF
--- a/www/webapp/src/components/Field/RecordItem.vue
+++ b/www/webapp/src/components/Field/RecordItem.vue
@@ -21,13 +21,7 @@
         @click:append="$emit('remove', $event)"
         @input="inputHandler()"
         @paste.prevent="pasteHandler($event)"
-        @keydown.8="backspaceHandler(index, $event)"
-        @keydown.32="spaceHandler(index, $event)"
-        @keydown.35="endHandler($event)"
-        @keydown.36="homeHandler($event)"
-        @keydown.37="leftHandler(index, $event)"
-        @keydown.39="rightHandler(index, $event)"
-        @keydown.46="deleteHandler(index, $event)"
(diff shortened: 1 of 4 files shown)


www/webapp/src/components/Field/RecordItem.vue
```diff
@@ -208,6 +202,31 @@ export default {
       return index < this.fields.length - 1
         && this.$refs.input[index].$refs.input.selectionStart === this.fields[index].value.length;
     },
+    keydownHandler(index, event) {
+      switch (event.keyCode) {
```
peterthomassen, line 206: The point of the change in Vue is that `[.keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) (because it is implementation-dependent). Instead, it's now recommended to look for `.key` values: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

peterthomassen (changes requested):

www/webapp/src/components/Field/RecordItem.vue
Rotzbua: Thank for the hint. My main intention was to get a (better) working vue3 test build.
Should I change it to `.key` in this PR or extra PR?

peterthomassen: Let's do it in this one :) -- If you have some time, can you send me an email please? Thanks.

peterthomassen (approved):
lgtm!

PR #761, opened 2023-07-15
fix(webapp): wrong import path for qrcode

Rotzbua:
Should be direct import like the other import in 2fa component.


peterthomassen (approved):
lgtm, thanks!

PR #763, opened 2023-07-15
fix(webapp): Fix component file name

Rotzbua:
* Extends #762
* Add eslint rule to prevent further mistakes
* Rename components

## Note

`vite` has a problem with `A.vue` which is compiled as `<a>` . Not sure why...
Renaming to `RecordA.vue` fixes this problem.

## Questions

- [ ] keep folder `Record` or replace by shorter `r`?

### DIFF
--- a/www/webapp/.eslintrc.js
+++ b/www/webapp/.eslintrc.js
@@ -6,7 +6,11 @@ module.exports = {
   env: {
     browser: true,
     node: true, // Can be removed after migration to vite.
-    es2022: true,
+    es2024: true,
+  },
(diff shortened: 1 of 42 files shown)


Rotzbua:
To be honest, I found it more useful to spend time on migration than on finding the cause.

peterthomassen:
Sure. - Let's get rid of the `Record` folder then, and we can go ahead.

peterthomassen (changes requested):
see comment above

peterthomassen (approved):
lgtm

PR #770, opened 2023-07-20
feat(webapp): use `eslint` as standalone

Rotzbua:
Remove dependency on vue-cli to enable migration to vite.

### DIFF
--- a/www/webapp/.eslintrc.js
+++ b/www/webapp/.eslintrc.js
@@ -1,8 +1,12 @@
+/* eslint-env node */
+
 /** @type {import('eslint').Linter.Config} */
 module.exports = {
   root: true,
   env: {
-    node: true
+    browser: true,
+    node: true, // Can be removed after migration to vite.
+    es2022: true,
   },
   extends: [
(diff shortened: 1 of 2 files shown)


www/webapp/.eslintrc.js
```diff
@@ -24,17 +28,11 @@ module.exports = {
...
   },
   ignorePatterns: ['**/src/modules/**/*'],
-  parserOptions: {
-    parser: '@babel/eslint-parser'
-  },
   overrides: [
     {
       files: [
-        '**/__tests__/*.{j,t}s?(x)'
+        'src/**/*.{vue,js,jsx,cjs,mjs,ts,tsx,cts,mts}'
```
peterthomassen, line 34: Does this actually override anything? (This override does not have a `rules` key.)

Rotzbua: Good catch. Seems to be a leftover. If I remember correctly the override "filtered" the input by the all files of "eslint src/". Override seems useless with the current version.

Issue #771, opened 2023-07-20
webapp(cli): migration from `vue-cli` to `vite`


Rotzbua:
Following steps are necessary:

- [x] Remove/replace dependencies relying on `vue-cli` #770
- [x] Fix file naming #763
- [ ] `vuelidate`
  - [ ] Migrate `vuelidate` to v2 **OR**
  - [x] Fix `vuelidate` (v0) build process
- [x] Migration to `vite`

peterthomassen:
vuelidate migration also tracked in #649.

PR #772, opened 2023-07-20
webapp(cli): migrate to `vite`

Rotzbua:
Finally a working vite system :)

## Reference

solves #771


www/webapp/vite.config.js
```diff
@@ -14,6 +15,7 @@ export default defineConfig({
         Components({
             resolvers: [VuetifyResolver()],
         }),
+        legacy(), // Build for old browser.
```
peterthomassen, line 18: How does this interact with `/www/webapp/.browserslistrc` (both the fact that both are there, and the fact that the policies aren't the same)?

peterthomassen (changes requested):
Yay, many kudos to you!! :tada: Thank you very much for the effort.

A few comments inline. I also noticed that the build failed with:
```
> vite build --no-clean

file:///usr/src/app/node_modules/vite/dist/node/cli.js:442
          throw new CACError(`Unknown option \`${name.length > 1 ? `--${name}` : `-${name}`}\``);
                ^

CACError: Unknown option `--clean`
    at Command.checkUnknownOptions (file:///usr/src/app/node_modules/vite/dist/node/cli.js:442:17)
    at CAC.runMatchedCommand (file:///usr/src/app/node_modules/vite/dist/node/cli.js:640:13)
    at CAC.parse (file:///usr/src/app/node_modules/vite/dist/node/cli.js:579:12)
    at file:///usr/src/app/node_modules/vite/dist/node/cli.js:906:5
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

Node.js v18.17.0
The command '/bin/sh -c npm run build -- --no-clean' returned a non-zero code: 1
Service 'www' failed to build : Build failed
```

www/webapp/src/router/index.js
```diff
@@ -15,24 +15,24 @@ const routes = [
     // route level code-splitting
     // this generates a separate chunk (about.[hash].js) for this route
     // which is lazy-loaded when the route is visited.
-    component: () => import(/* webpackChunkName: "signup" */ '@/views/SignUp.vue')
```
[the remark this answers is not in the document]
Rotzbua: `vue-cli` (webpack) created a big js. The comment `webpackChunkName` it was possible to split this big js into smaller chunks.
`vite` (rollup) transforms every source code into a chunk. So it creates automatically the smallest possible chunk. If required you can summarize multiple small chunks into a bigger to avoid a lot of requests to the server or delays by requests.

Example build output (partial):

grafik

www/webapp/vite.config.js
Rotzbua: `vite` uses `esbuild` which does not support `.browserslistrc`.
The legacy build uses broswerslist and creates js for older browsers.

References:
https://vitejs.dev/guide/build.html#browser-compatibility
https://dev.to/meduzen/when-vite-ignores-your-browserslist-configuration-3hoe

peterthomassen: So we should remove `/www/webapp/.browserslistrc`?

Also, the [default `targets` argument for `legacy`](https://www.npmjs.com/package/@vitejs/plugin-legacy#user-content-targets) is `last 2 versions and not dead, > 0.3%, Firefox ESR`, which is different from what we had in the other file. I'm not saying the change is bad, I'm just making the change explicit.

www/webapp/src/router/index.js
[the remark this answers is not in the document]
peterthomassen: Aha. I think that's fine for now. Earlier, you put quite some effort into reducing latency -- I trust that in case the many requests pose a problem, you're probably going to do something about it :p

www/webapp/vite.config.js
Rotzbua: The `target` setting is for the `vite-legacy` package. After generation the chunks are processed again with `babel` which uses `.browserlistrc`.

[Source](
https://github.com/vitejs/vite/tree/06e4f57724f947b584dad68fdd446989bdc76aa0/packages/plugin-legacy#vitejsplugin-legacy-)

> Generate a corresponding legacy chunk for every chunk in the final bundle, transformed with [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env) and emitted as [SystemJS modules](https://github.com/systemjs/systemjs) (code splitting is still supported!).

The default config of the plugin includes more browser versions than the current config.

So I think the config could be kept, if you wish to have control otherwise it can be removed and trust the default config of the plugin.

www/Dockerfile
```diff
@@ -14,14 +14,14 @@ ARG DESECSTACK_API_SEPA_CREDITOR_ID
...
-    VUE_APP_EMAIL=support@desec.${DESECSTACK_DOMAIN}
+ENV VITE_APP_DESECSTACK_NS=${DESECSTACK_NS} \
+    VITE_APP_DESECSTACK_API_SEPA_CREDITOR_ID=${DESECSTACK_API_SEPA_CREDITOR_ID} \
+    VITE_APP_DESECSTACK_API_SEPA_CREDITOR_NAME=${DESECSTACK_API_SEPA_CREDITOR_NAME} \
+    VITE_APP_LOCAL_PUBLIC_SUFFIXES=dedyn.${DESECSTACK_DOMAIN} \
+    VITE_APP_EMAIL=support@desec.${DESECSTACK_DOMAIN}
 
 COPY webapp/ /usr/src/app/
-RUN npm run build -- --no-clean
+RUN npm run build -- --emptyOutDir false
```
peterthomassen, line 24: Why this change?

peterthomassen (changes requested):
I think the last commit should be squashed into 97f1fd0f649ca48f45e096b58329119d688ba132, otherwise seems good to go.

www/webapp/vite.config.js
```diff
@@ -19,6 +19,14 @@ export default defineConfig({
...
     server: {
         port: 8080,
+        proxy: {
+            '/api': {
+                target: "https://desec.io",
+                changeOrigin: true,
+                secure: true,
+                // rewrite: path => path.replace('/api', ''),
+            }
+        }
```
peterthomassen, line 29: I usually run against the local dev API. I assume the above means that this would no longer work. Can we add a switch for this, e.g. via an env variable?

Rotzbua: Good point. I think this should be done with a separate PR.

peterthomassen (dismissed):

peterthomassen:
Before deploying, I tried to fire up the dev environment, and with this PR applied, the `www` container gives me "502 Bad Gateway". Do you know why that is?

peterthomassen:
The webapp container is still not listening on port 8080. I tried adding `--host` after `npm run dev`, but nothing changes. Do you have an idea what could cause this?

Rotzbua:
> I tried adding `--host` after `npm run dev`, but nothing changes. Do you have an idea what could cause this?

You found the reason and solution 👍  . I think you forgot the `--` before `--host`

```bash
npm run dev -- --host
```

docker-compose.dev.yml
```diff
@@ -70,15 +70,13 @@ services:
     volumes:
     - ./www/webapp/:/usr/src/app/
     working_dir: /usr/src/app/
-    command: bash -c "npm install && npm run serve --fix"
+    command: bash -c "npm install && npm run dev -- --host"
     environment:
-    - CODESANDBOX_SSE=1  # so that ws: protocol ends up as auto: in node_modules/@vue/cli-service/lib/commands/serve.js
-    - CYPRESS_CACHE_FOLDER=/tmp/.cache  # https://github.com/cypress-io/cypress/issues/1281#issuecomment-404823001
```
peterthomassen, line 76: Good catch!

peterthomassen (approved):
lgtm, thank you very much!

peterthomassen:
Before:
image

After:
image

The JS size has decreased a little, while the CSS size has increase by about 8-9%. Also, the index page is no longer compressed (it now has a lot of whitespace).

Loading overall is faster now (but I didn't control for statistical artifacts).

Rotzbua:
I found an issue with CSP #788

PR #774, opened 2023-07-22
fix(webapp): `eslint` regression by #770

Rotzbua:
## Problem

Not all files are processed by `eslint`.
Parser was not set correctly.

## Reference

#770


peterthomassen (approved):
lgtm

PR #794, opened 2023-08-15
fix(webapp): add missing hashes to CSP

### LINKED ISSUE #788: nginx: update `CSP` header for webapp
## Problem

After migration to `vite` there is some inline javascript.

> Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-+5XkZFazzJo8n0iOP4ti/cLCMUudTf//Mzkb7xNPXIc='), or a nonce ('nonce-...') is required to enable inline execution.

Website works even with the error but inline code seems required for legacy browser support.

## Solution

Temporary fix: update CSP to more unsafe config.

Update header to include hashes.

Rotzbua:
Add inline script hashes for `@vitejs/plugin-legacy`:

Add the corresponding hashes to your script-src list:

    sha256-MS6/3FCg4WjP9gwgaBGwLpRCY6fZBgwmhVCdrPrNf3E=
    sha256-tQjf8gvb2ROOMapIxFvFAYBeUJ0v1HCbOcSmDNXGtDo=
    sha256-4y/gEB2/KIwZFTfNqwXJq4olzvmQ0S214m9jwKgNXoc=
    sha256-+5XkZFazzJo8n0iOP4ti/cLCMUudTf//Mzkb7xNPXIc=

Source:
https://github.com/vitejs/vite/tree/main/packages/plugin-legacy#content-security-policy

## Reference

fixes #788

### DIFF
--- a/test/e2e2/spec/test_www.py
+++ b/test/e2e2/spec/test_www.py
@@ -87,10 +87,13 @@ def test_unknown_hosts(api_anon, protocol, hostname):
 
 def test_security_headers(api_anon):
     api_anon.headers = {}
+    # CSP hashes are for legacy browser support.
+    # Source: https://github.com/vitejs/vite/tree/v4/packages/plugin-legacy#content-security-policy
     expected_headers = {
         'Strict-Transport-Security': 'max-age=31536000; includeSubDomains; preload',
         'Content-Security-Policy': "default-src 'self'; frame-src 'none'; connect-src 'self'; font-src 'self' data:; "
-                                   "img-src 'self' data:; media-src data:; script-src 'self' 'unsafe-eval'; "
+                                   "img-src 'self' data:; media-src data:; "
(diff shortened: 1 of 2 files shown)


peterthomassen (commented):
Thanks. Now, unfortunately, the hashes have changed yesterday (see https://github.com/vitejs/vite/commit/4033a320d6809c9a0c2552f0ef2bf686c63aa35a).

How are we going to go about this? -- If they get outdated, it's about as good as not having them at all.

Rotzbua:
> Thanks. Now, unfortunately, the hashes have changed yesterday (see [vitejs/vite@4033a32](https://github.com/vitejs/vite/commit/4033a320d6809c9a0c2552f0ef2bf686c63aa35a)).

The commit is part of vite v5 which is not released yet.
Since this would be a mayor update, it is ok that the hash changes.

peterthomassen:
Mh. Nevertheless, the code comment linking to the source is not correct (it lists the new hashes, not the ones that are in this PR).

peterthomassen (approved):
thanks, lgtm!

Issue #795, opened 2023-08-15
openvpn: update configuration


Rotzbua:
Following warnings are thrown:

- [ ] Note: Treating option '--ncp-ciphers' as  '--data-ciphers' (renamed in OpenVPN 2.5).
- [ ] Consider setting groups/curves preference with tls-groups instead of forcing a specific curve with ecdh-curve.
- [ ] --cipher is not set. Previous OpenVPN version defaulted to BF-CBC as fallback when cipher negotiation failed in this case. If you need this fallback please add '--data-ciphers-fallback BF-CBC' to your configuration and/or add BF-CBC to --data-ciphers.

PR #796, opened 2023-08-15
fix(openvpn): update config for v2.5

Rotzbua:
## Change

Fixes warning:
* Note: Treating option '--ncp-ciphers' as  '--data-ciphers' (renamed in OpenVPN 2.5).
* --cipher is not set. Previous OpenVPN version defaulted to BF-CBC as fallback when cipher negotiation failed in this case. If you need this fallback please add '--data-ciphers-fallback BF-CBC' to your configuration and/or add BF-CBC to --data-ciphers.

## Reference

* #795
* Source: https://community.openvpn.net/openvpn/wiki/CipherNegotiation

### DIFF
--- a/openvpn-server/conf/server.conf
+++ b/openvpn-server/conf/server.conf
@@ -10,7 +10,9 @@ key /etc/openvpn/secrets/server.key  # This file should be kept secret
 
 tls-version-min 1.3 or-highest
 tls-cipher TLS-ECDHE-RSA-WITH-AES-256-GCM-SHA384
-ncp-ciphers AES-256-GCM
+data-ciphers AES-256-GCM
+# Fallback for OpenVPN before version 2.5, use "-" to disable support.
+data-ciphers-fallback AES-256-GCM
 auth SHA256
 dh none
 ecdh-curve secp384r1


openvpn-server/conf/server.conf
```diff
@@ -10,7 +10,8 @@ key /etc/openvpn/secrets/server.key  # This file should be kept secret
 
 tls-version-min 1.3 or-highest
 tls-cipher TLS-ECDHE-RSA-WITH-AES-256-GCM-SHA384
-ncp-ciphers AES-256-GCM
+data-ciphers AES-256-GCM
+data-ciphers-fallback -
```
peterthomassen, line 14: Our clients are currently OpenVPN 2.4. My understanding is that this fallback is needed for these clients (["Openvpn Clients which do not support --data-ciphers (Openvpn 2.4, 2.3)"](https://community.openvpn.net/openvpn/wiki/CipherNegotiation#Effectivedirectivesandterms)). When setting it to `-`, I would think that no connection could be established.

It seems to me that the correct setting would be to set the fallback to one of the algorithms in `data-ciphers`, of which we have only one, so `data-ciphers-fallback AES-256-GCM` would be correct. Do you think that makes sense?

Rotzbua: The documentation is a bit confusing. The cipher example does not mention `data-ciphers-fallback` but shows a valid connection 😕 : https://community.openvpn.net/openvpn/wiki/CipherNegotiation#Serverversion2.5
Anyway I think the fallback will not harm the security.

peterthomassen (approved):
lgtm; squashed and rebased.

PR #799, opened 2023-08-29
feat(dependabot): group dependency updates

### LINKED ISSUE #798: Dependabot groups
[Dependency grouping](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) is now [generally available](https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/).

I think it would be great to use this, in order to reduce the number of PRs. Suggested grouping:
- By production vs. by development dependencies
- For production dependencies, don't group major version updates (keep single PR)

@Rotzbua Would you want to take a look at this?

Rotzbua:
Closes #798

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,10 @@ updates:
       interval: "monthly"
       timezone: "Europe/Berlin"
       time: "05:00"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
   - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


peterthomassen (approved):
Thanks! I added an `update-types` selector for production dependencies, so that major version updates aren't grouped. Otherwise, lgtm!

Issue #752, opened 2023-07-06
Webapp: refactor form inputs


Rotzbua:
## Problem

Some form inputs are C&P and code is redundant.

Affected inputs:

- [ ] E-Mail
- [ ] Password
- [ ] Captcha

## Solution

Refactor to generic component.

Implementation already done, but must be evaluated. PR in some days...

peterthomassen:
Cool! - Can you shoot me an email when convenient?

PR #801, opened 2023-08-31
feat(webapp): create generic email & password field

Rotzbua:
## Change

* generic email field
* generic password field

## Reference

Mentioned in #752

### DIFF
--- a/www/webapp/src/components/Field/GenericEmail.vue
+++ b/www/webapp/src/components/Field/GenericEmail.vue
@@ -1,0 +1,101 @@
+<template>
+  <v-text-field
+      v-bind="computedProps"
+      v-on="$listeners"
+  />
+</template>
+
+<script>
+import {email_pattern} from '@/validation';
+import {mdiEmail, mdiEmailLock} from '@mdi/js';
+import {VTextField} from 'vuetify/lib/components';
+
+export default {
+  name: 'GenericEmail',
+  extends: VTextField,
+  props: {
+    value: {
+      type: String,
+      required: false,
+    },
+    errorMessages: {
+      type: [String, Array],
+      default: () => [],
+    },
+    label: {
+      type: String,
+      required: false,
+    },
+    placeholder: {
+      type: String,
+      required: false,
+    },
+    required: {
+      type: Boolean,
+      default: true,
+    },
+    standalone: {
+      type: Boolean,
+      default: false,
+    },
+    readonly: {
+      type: Boolean,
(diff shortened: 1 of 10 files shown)


www/webapp/src/views/ChangeEmail.vue
```diff
@@ -35,14 +35,10 @@
...
-                                    v-model="email"
-                                    label="Current Email"
-                                    prepend-icon="mdi-blank"
-                                    outlined
-                                    required
-                                    :disabled="true"
-                                    validate-on-blur
+                            <generic-email
+                                v-model="email"
+                                label="Current Email Address"
```
peterthomassen, line 40: Isn't this overwritten by `computed`?

www/webapp/src/components/Field/GenericEmail.vue
```diff
@@ -0,0 +1,101 @@
...
+      if(this.standalone) {
+        label = '';
+      }
+      if(this.label) { // Override if custom is set.
+        label = this.label;
+      }
+      const icon = this.readonly ? mdiEmailLock : mdiEmail;
+      const ruleDefs = {
+        required: v => !!v || 'Email is required.',
+        valid: v => (v !== undefined && !!email_pattern.test(v)) || 'We need an valid email address for account recovery and technical support.',
```
peterthomassen, line 72: This text either should be generic or configurable through a prop.

www/webapp/src/views/ResetPassword.vue
```diff
@@ -131,9 +124,12 @@
   export default {
     name: 'ResetPassword',
     components: {
+      GenericEmail,
       ErrorAlert,
     },
     data: () => ({
+      user: useUserStore(),
```
peterthomassen, line 131: What is this needed for?

peterthomassen (commented):
Besides some inline comments -- why go through the hassle of generalizing this?

I see there currently is some code duplication, but the functionality is not identical. For example, the change email form has a prepended letter icon, and for consistent spacing that requires other fields *on that page* to be prepended with a blank icon.

Making such functionality available by putting a bunch of knobs into a generalized implementation comes with a lot of complexity, and also code that's not needed so far (e.g. various pass-through props), which would be there instead of the current code duplication. Note that the new implementation adds about 65% more LOC than it removes. I'm not yet convinced that's a good idea.

Rotzbua:
Thanks for the feedback 👍 I will have a look.

PR #809, opened 2023-09-22
docs: Add daemon timeout to ddclient config

Gustl22:
Daemon is needed in the most cases, as by default the user isn't expected to enable it by themselves or via a separate cron. This can may prevent from unexpected outages of the host.

### DIFF
--- a/docs/dyndns/configure.rst
+++ b/docs/dyndns/configure.rst
@@ -123,6 +123,8 @@ file similar to this one, with the three placeholders replaced by your domain
 name and your token secret::
 
   protocol=dyndns2
+  # Run in daemon mode: auto-update DNS every 10min. (Alternatively, use cron.)
+  #daemon=600
   # "use=cmd" and the curl command is one way of doing this; other ways exist
   use=cmd, cmd='curl https://checkipv4.dedyn.io/'
   ssl=yes


peterthomassen:
First, thanks for your contribution!

> Daemon is needed in the most cases

Do we have any data on that?

So far, it seems like nobody was missing this, which actually suggests that daemon is *not* needed in most case.

It thus seems to me that the preferable way to go would be to add the `daemon=600` (or 300?) option as a comment (leaving the details as they are now, but giving a hint to people who need daemon mode).

Gustl22:
> Do we have any data on that?

Setting the `daemon` the one or the other way is proposed in most online tutorials, or by ddclient themselves:
https://ddclient.net/#configuration
https://ddclient.net/usage.html#simple-configuration
https://help.ubuntu.com/community/DynamicDNS#ddclient
https://help.dyn.com/ddclient/

> So far, it seems like nobody was missing this, which actually suggests that daemon is not needed in most case.

I more think that the majority did have this issue and after a bit of googling they found the solution by setting this parameter and not cared about updating the docs. If someone uses the dynamic dns service they need to somehow update the IP after a while, right? I don't see there's a more elegant solution to that, or is there? But all of that is only a hypothesis of course, I can't really prove it.

> It thus seems to me that the preferable way to go would be to add the daemon=600 (or 300?) option as a comment (leaving the details as they are now, but giving a hint to people who need daemon mode).

You mean as a note or as a comment in the code block? I strongly prefer the last one, as it is directly inherited in the config and is far more obvious, if something doesn't work after a while. And for me it doesn't really matter, if that are 5m (300s) or 10m (600s). Ddclient contradict themselves in that sense. 

The frustrating part here is, that it works the first time, as the service is started or the PC is restarted. But in the other case e.g. after a few weeks when the IP has changed, but the dyndns is not updated, that's unexpected, but which is the purpose of the whole thing, right (?) That differs from the other parameters, which work or may not work, but always on the first try and then can immediately adjusted.

peterthomassen:
> > Do we have any data on that?
> 
> Setting the `daemon` the one or the other way is proposed in most online tutorials, or by ddclient themselves:
> https://ddclient.net/#configuration

This example precisely doesn't list the `daemon` option in the config.

> > So far, it seems like nobody was missing this, which actually suggests that daemon is not needed in most case.
> 
> I more think that the majority did have this issue and after a bit of googling they found the solution by setting this parameter and not cared about updating the docs.

Given the amount of questions in our [forum](https://talk.desec.io/) about other issues, I'm pretty sure this would have surfaced.

> If someone uses the dynamic dns service they need to somehow update the IP after a while, right? I don't see there's a more elegant solution to that, or is there?

I guess it's a matter of taste whether this is more elegant than a cron job or not. But it's certainly worth making people aware that the option exists.

> > It thus seems to me that the preferable way to go would be to add the daemon=600 (or 300?) option as a comment (leaving the details as they are now, but giving a hint to people who need daemon mode).
> 
> You mean as a note or as a comment in the code block? I strongly prefer the last one, as it is directly inherited in the config and is far more obvious, if something doesn't work after a while.

I'm not sure I understand this sentence.

I've squashed your commits and turned the `daemon` option into a comment, so that people who need it can enable it easily.

Thanks for raising this!

peterthomassen (approved):

Gustl22:
>>  https://ddclient.net/#configuration

> This example precisely doesn't list the daemon option in the config.

Yes, but it's passed while calling, which is basically the same, I think.

> Given the amount of questions in our [forum](https://talk.desec.io/) about other issues, I'm pretty sure this would have surfaced.

Indeed, haven't looked in the forum. Maybe I got a wrong impression. I kind of wonder why noone had this problem before.

>> You mean as a note or as a comment in the code block? 

> I'm not sure I understand this sentence.

There were also other parameters mentioned in the notes below (`Note 1: `, etc.), which could have been also an option to add a `Note 3: daemon...`
But I'm more than fine with the current solution :)

Thank you for reviewing, investing your time, and providing such a good feature-rich service!

PR #830, opened 2023-11-05
fix(docs): single backticks not allowed for keywords

Rotzbua:


peterthomassen (approved):
good catch, thanks! -- The intention was italic, not monospace, so I replaced your ` `` ` with `*`.

PR #831, opened 2023-11-05
fix(docs): update outdated links

Rotzbua:
Target not available. Update to new headings.


peterthomassen (approved):
good catch, thanks! I changed the text to reference the overall section on record types, rather than each subsection.

PR #836, opened 2023-11-16
New DS setup dialog

peterthomassen:
Ready for review.

(Still need to adjust an error detail when user is not authenticated, but everything else is ready for review.)

image


peterthomassen:
Ready!

nils-wisiol (approved):
lgtm!

PR #849, opened 2023-12-05
20231205 donate

peterthomassen:
image

Note: first sentence changed to
> If you prefer to donate in a crypto currency, you can use the following wallets:


nils-wisiol (approved):
lgtm

PR #840, opened 2023-11-23
Improvements on Token Scoping

peterthomassen:
- [x] Change policy `id` field to UUID
- [x] Add columns `subname`, `type` to policies
- [x] Adjust policies view for new fields
- [x] Include subname/type in Token default policies
- [x] Extend RRset serializers to honor policies
- [x] Update docs
- [ ] ~~Turn `perm_manage_tokens` into a virtual field (`true` iff token has no policies) -- rethink: no policy means "allow/create delete domains", but not necessarily manage tokens~~

Related: #347

PR #850, opened 2023-12-12
fix(tests): repair libfaketime build

peterthomassen:


peterthomassen:
will be included in #840

PR #851, opened 2023-12-12
fix(api): don't use serializer for watchdog zone updates

peterthomassen:
RRsetSerializer now assumes there's a request with a token, and checks its permission. The watchdog updates aren't based on a request, so we now use the model directly (instead of the serializer).

Tested on my machine. (The model interface also has unit tests.)


nils-wisiol (approved):
lgtm

PR #867, opened 2024-01-09
feat(webapp): migrate from @mdi/font to @mdi/js

### LINKED ISSUE #710: Replace icon font `@mdi/font`

Rotzbua:
Final migration. Mdi webfont not used anymore.

## Change

Use tree shakeable version instead of big webfont.

## Reference

Closes #710

### DIFF
--- a/www/webapp/src/views/ResetPassword.vue
+++ b/www/webapp/src/views/ResetPassword.vue
@@ -42,7 +42,7 @@
                             <v-text-field
                                     v-model="email"
                                     label="Email"
-                                    prepend-icon="mdi-email"
+                                    :prepend-icon="mdiEmail"
                                     outlined
                                     required
                                     :rules="email_rules"
@@ -59,7 +59,7 @@
                                     <v-text-field
                                             v-model="captchaSolution"
                                             label="Type CAPTCHA text here"
-                                            prepend-icon="mdi-account-check"
+                                            :prepend-icon="mdiAccountCheck"
                                             outlined
                                             required
                                             :rules="captcha_rules"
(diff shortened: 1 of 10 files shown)


www/webapp/src/views/ResetPassword.vue
```diff
@@ -122,6 +122,7 @@
   import {email_pattern} from '@/validation';
   import {digestError} from '@/utils';
   import ErrorAlert from '@/components/ErrorAlert.vue';
+  import {mdiAccountCheck, mdiDns, mdiEarHearing, mdiEmail, mdiEye, mdiRefresh} from "@mdi/js";
```
peterthomassen, line 125: Unused `mdiDns`

www/webapp/src/assets/global.css
```diff
@@ -0,0 +1,5 @@
+/* Dummy for a blank icon. Used to align input fields. */
+.mdi-blank::before {
+    content: "\F68C";
```
peterthomassen, line 3: Why this one specifically?

If I'm understanding this correctly, this is in a private-use area of the codepoint space, so it's rendering (in particular, width), is undefined. Why should one expect this to work?

peterthomassen (changes requested):

peterthomassen (approved):
lgtm! Thank you very much <3

peterthomassen:
This has saved about 20% of bytes transmitted when loading the webapp. Very cool!

PR #870, opened 2024-01-12
Address dynDNS mass registration problem

peterthomassen:
Due to a [tutorial on Youtube](https://www.youtube.com/watch?v=i3tmB2H7d-s) and related blog posts, dynDNS registrations are going through the roof, causing frequent resigning of dedyn.io and excessive load.

To address the immediate problem, the domain limit for new accounts has been set to 0. This still causes a support request about every 2 minutes, and also affects users that want to bring their own domain.

Before we find a permanent solution, we're shutting down dedyn.io registrations through the sign-up form and via POST requests to the `domains/` endpoint. We will attempt to continue Nextcloud VM's use case.

The following screenshot shows this PR's sign-up form. I hope that with respect to the number of support tickets, being explicit about the registration suspension will be more effective than a silent removal.

image

@enoch85


peterthomassen:
Screenshot for POST domains/ via GUI with domain name under LPS below (049e437dba01b9d4e1f6901621ea1b7396a7157b). Other domain names work normally.

image

peterthomassen:
When `DESECSTACK_API_REGISTER_LPS=0` is set, signing up with an LPS domain looks like this (fc610143a010a41b3719f24e2e8cb91f9dbd7db9):

image

peterthomassen:
Tested that Nextcloud VM workflow continues to function.

peterthomassen:
Rebased, and picked a few dependabot version bumps on that occasion

nils-wisiol (approved):

PR #871, opened 2024-01-12
fix(webapp): class rename subheading → subtitle-1

Rotzbua:
fix missing migrations to vuetify v2
https://v2.vuetifyjs.com/en/getting-started/upgrade-guide/#typography

fix illegal paragraph in header.

text size is a bit smaller.


peterthomassen (approved):
lgtm

PR #882, opened 2024-02-04
fix(webapp): transform asset url for `v-img`

Rotzbua:
`require` not required anymore 😄 

`require` does not work with Vue3.


peterthomassen (approved):
lgtm

Issue #472, opened 2020-11-10
A11y: Add aria-labels to webapp buttons


peterthomassen:

robin24:
In terms of the heading, a level 1 (h1) heading with the page title, e.g. "Domains" at the beginning of the section would probably do the trick very nicely, and would likely be all that's needed here to improve screen reader navigation.

acorncom:
Howdy, are you still looking for help on accessibility issues? I've been doing work like this for a client using https://accessibilityinsights.io/docs/web/overview/ and might be able to do some work toward this

peterthomassen:
Yes, David, that would be very much appreciated! <3

xyhhx:
I can lend a hand with this. I'll try to set aside time for it this weekend

peterthomassen:
Thank you!

xyhhx:
Geez, fast reply lol

xyhhx:
> make progress bar at the top of page less irritating

What does this mean?

nils-wisiol:
This is the progress bar in question: 
![image](https://github.com/desec-io/desec-stack/assets/1026940/8b4de325-d5bf-47c2-9061-e1a0b8807180)

Not sure why @peterthomassen says its irritating - I hardly notice it when I use desec.io. I took the screenshot after I set the network speed to "GPRS" in the debug tools and even then it took me a couple of attempts to get a good shot. Maybe we can just get rid of it entirely.

xyhhx:
Seems fine enough to me, but more relevantly, it doesn't seem to affect accessibility either

peterthomassen:
I recall this was based on feedback from a visually impaired person.

I'm OK removing the progress bar, but maybe we should have some other kind of indicator when a request is pending. Perhaps a spinner somewhere?

PR #890, opened 2024-02-11
feat(a11y): add aria-labels to various buttons

xyhhx:
Affects #472


www/webapp/src/views/CrudList.vue
```diff
@@ -61,6 +61,7 @@
                         fab
                         depressed
                         :disabled="user.working"
+                        aria-label="Add new record"
```
peterthomassen, line 64:
```suggestion
                        aria-label="Add new entry"
```
`CrudList.vue` is generic an e.g. also used in the token management GUI.

www/webapp/src/App.vue
```diff
@@ -105,13 +106,15 @@
             depressed
             :href="alert.href"
             v-if="alert.href"
+            aria-label="More"
```
peterthomassen, line 109:
```suggestion
            :aria-label="alert.button || 'More'"
```

www/webapp/src/views/CrudList.vue
```diff
@@ -170,6 +171,7 @@
                                 :outlined="!createDialogSuccess"
                                 :disabled="createDialogWorking"
                                 @click.native="close"
+                                aria-label={{ createDialogSuccess ? 'Close' : 'Cancel' }}
```
peterthomassen, line 174: Is that valid syntax? I guess it should be:
```suggestion
                                :aria-label="createDialogSuccess ? 'Close' : 'Cancel'"
```

www/webapp/src/views/DomainSetup.vue
```diff
@@ -74,7 +74,7 @@
             <v-col cols="auto">DS Format</v-col>
             <v-spacer></v-spacer>
             <v-col class="text-right">
-              <v-btn @click="copyToClipboard(ds.join('\n'))" text small>
+              <v-btn @click="copyToClipboard(ds.join('\n'))" text small aria-label="Copy DS">
```
peterthomassen, line 77:
```suggestion
              <v-btn @click="copyToClipboard(ds.join('\n'))" text small aria-label="Copy DS information">
```

There might be several.

www/webapp/src/views/DomainSetup.vue
```diff
@@ -94,7 +94,7 @@
             <v-col cols="auto">DNSKEY Format</v-col>
             <v-spacer></v-spacer>
             <v-col class="text-right">
-              <v-btn @click="copyToClipboard(dnskey.join('\n'))" text small>
+              <v-btn @click="copyToClipboard(dnskey.join('\n'))" text small aria-label="Copy DNSKey">
```
peterthomassen, line 97:
```suggestion
              <v-btn @click="copyToClipboard(dnskey.join('\n'))" text small aria-label="Copy DNSKEY information">
```

peterthomassen (changes requested):
This is good stuff, thanks!

Overall, I was wondering whether it's possible to reduce the duplication, by somehow telling it that `aria-label` should have the same value as the element's inner text node. Is that possible? And if so, does it make sense to do it?

xyhhx:
> Overall, I was wondering whether it's possible to reduce the duplication, by somehow telling it that aria-label should have the same value as the element's inner text node. Is that possible? And if so, does it make sense to do it?

Good idea. I'm not sure if it's possible, I can look around; but I do think that at least for some places it does make sense!

xyhhx:
> This is good stuff, thanks!
> 
> Overall, I was wondering whether it's possible to reduce the duplication, by somehow telling it that `aria-label` should have the same value as the element's inner text node. Is that possible? And if so, does it make sense to do it?

Looking more into this, I'm not even sure every button _needs_ `aria-label`.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) `aria-label` is for when the accessible name isn't available or is inaccurate:

> Sometimes the default [accessible name](https://w3c.github.io/accname/#dfn-accessible-name) of an element is missing, or does not accurately describe its contents, and there is no content visible in the DOM that can be associated with the object to give it meaning.

[According to the W3C](https://w3c.github.io/accname/#dfn-accessible-name), the accessible name can be automatically derived from the visible text of a button/link/etc:

> The value of the accessible name may be derived from a visible (e.g., the visible text on a button) or invisible (e.g., the text alternative that describes an icon) property of the user interface element

So maybe I should review this PR and actually remove any occurrences where the `aria-label` I've provided is the exact match of the elements' values

peterthomassen:
> So maybe I should review this PR and actually remove any occurrences where the `aria-label` I've provided is the exact match of the elements' values

Yes, I think so!

PR #919, opened 2024-05-03
feat(api): temporarily raise rate limits

peterthomassen:


nils-wisiol (approved):
I'm a bit concerned because the tests fail after the bot changes, but haven't looked why. The commit msg title shouldn't cover plannings ("temporarily"), but that's just a nit that doesn't necessarily need fixing. Otherwise lgtm

peterthomassen:
Thanks!

> I'm a bit concerned because the tests fail after the bot changes, but haven't looked why.

The tests fail because of some tests doing live DNS queries (related to abuse handling, where we map IP addresses to IP subnets using a DNS lookup service). It's on my radar, and when I got time I'll turn the DNS queries into mock ones.

PR #928, opened 2024-05-22
fix(webapp): dependabot ignore `eslint`

Rotzbua:
eslint v9 has breaking changes and plugins are not yet fully compatible.

So let dependabot ignore `eslint`.


peterthomassen (approved):
lgtm

PR #937, opened 2024-06-18
20240618 misc

peterthomassen:


api/desecapi/tests/test_throttling_user.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    Based on DRF's test_throttling.py.
+    """
+
+    def setUp(self):
+        super().setUp()
+        self.factory = APIRequestFactory()
+
+    def _test_requests_are_throttled(self, counts, user=None):
+        def do_test():
```
nils-wisiol, line 36: this seems to be only called once so it doesn't need to be encapsulated. "de"capsulating it could improve readability of the code

api/desecapi/tests/test_throttling_user.py
```diff
@@ -0,0 +1,80 @@
...
+        request = self.factory.get("/")
+        if user is not None:
+            force_authenticate(request, user=user)
+        with override_settings(
+            REST_FRAMEWORK={"DEFAULT_THROTTLE_RATES": {"user": "10/d"}}
+        ):
+            do_test()
+
+    def test_requests_are_throttled_unauthenticated(self):
+        self._test_requests_are_throttled([(0, 10, 86400)])
```
nils-wisiol, line 67: how can the 86400 requests be successful? I only see a limit of 10/d

nils-wisiol (approved):
nits and questions only

api/desecapi/tests/test_throttling_user.py
peterthomassen: This is not the number of requests, but the expected "Retry-After" header value. :) The numbers say that it's trying 10 times, after which such a header with value 86400 is expected.

nils-wisiol: Is the ability to run several tests with one call to `_test_requests_are_throttled` actually used anywhere? I see two calls, both only have a single entry in the `counts` array. I find that 'anonymous data structure' confusing, if it is not required it could be removed.

nils-wisiol (approved):
lgtm

api/desecapi/tests/test_throttling_user.py
peterthomassen: That's an omission; I've now modified `test_requests_are_throttled_unauthenticated` to make multiple requests, in order to test that a new quota is made available after the wait time is over..

PR #943, opened 2024-07-09
Update update-api.rst

jameskimmel:
not sure if it would be better the remove or move the preserve examples.
What do you think? Feel free to delete or move them.

### DIFF
--- a/docs/dyndns/update-api.rst
+++ b/docs/dyndns/update-api.rst
@@ -173,9 +173,11 @@ stand-in for an IPv6 address. Replace those (including the ``<`` and ``>``)
 with your respective values.
 
 
-Basic authentication with automatic IP detection (IPv4 **or** IPv6)::
+Basic authentication with automatic IP detection (IPv4 **or** IPv6), option 1::
 
   curl --user <your domain>:<your token secret> https://update.dedyn.io/
+
+or option 2::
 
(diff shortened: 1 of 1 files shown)


docs/dyndns/update-api.rst
```diff
@@ -185,17 +187,31 @@ Basic authentication with forced use of IPv4 (will remove IPv6 address from the
...
-Basic authentication with simultaneous update of IPv4 and IPv6::
+or option #2::
+
+  curl --ipv6 https://update.dedyn.io/?hostname=<your domain> \
+    --header "Authorization: Token <your token secret>"
+
+Basic authentication with forced use of IPv6 (while preserving the IPv4 address from the DNS)::
+
+  curl --ipv6 "https://update.dedyn.io/?hostname=<your domain>&myipv4=preserve" \
+    --header "Authorization: Token <your token secret>"
```
peterthomassen, line 207: Using `update6.dedyn.io` is preferred over `--ipv6` (because the latter is a curl flag, and the former works with all clients).

peterthomassen (changes requested):
Thanks! One small detail to change.

On that occasion, can you please also remove the `#` from `#1` and `#2`? Thanks!

peterthomassen (approved):

PR #944, opened 2024-07-10
Update configure.rst

jameskimmel:
made some clarifications

### DIFF
--- a/docs/dyndns/configure.rst
+++ b/docs/dyndns/configure.rst
@@ -115,14 +115,16 @@ Enter ``update.dedyn.io`` as the dynamic DNS service.
 Enter the domain you want to update as the username.
 As your password, use an API token.
 Tell ddclient how to detect your IP address. One option to detect your IP would be
-to use ``checkipv4.dedyn.io`` to check your IPv4 or ``checkipv6.dedyn.io`` to check your IPv6.
+to use ``https://checkipv4.dedyn.io`` to check your IPv4 or ``https://checkipv6.dedyn.io`` to check your IPv6.
 Insert the update interval and the domain name to update.
 
 **Note:** As of the time of this writing, ddclient does not use an encrypted
 HTTPS connection by default. To enable it, open ``/etc/ddclient.conf`` and add
 ``ssl=yes`` above the ``server=`` statement. We **strongly recommend** doing
-so; otherwise, your credentials will be exposed during transmission. Also webservers
-like ``checkipv4.dedyn.io`` for IP checks depend on it.
+so; otherwise, your credentials will be exposed during transmission.
+
+To test your setup, run ``sudo ddclient -force`` and see if everything works as
+expected.
 
 Manual configuration (other systems)
 ************************************
--- a/docs/dyndns/configure.rst
+++ b/docs/dyndns/configure.rst
@@ -99,6 +99,7 @@ Option 2: Use ddclient
 
 Automatic configuration (Debian-/Ubuntu-based systems)
 ******************************************************
+
 If you're on Debian, Ubuntu or any other Linux distribution that provides you
 with the ddclient package, you can use it to update your IP address with our
 servers. Note that depending on the ddclient version you are using, IPv6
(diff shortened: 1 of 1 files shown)


docs/dyndns/configure.rst
```diff
@@ -108,16 +109,21 @@ To install ddclient, run ``sudo apt-get install ddclient``. If a configuration
...
 
-In the configuration process, select "other" dynamic DNS service provider, and
-enter ``update.dedyn.io`` as the dynamic DNS server. Next, tell ddclient to use
-the "dyndns2" protocol to perform updates. Afterwards, enter the username and
-the token secret that you received during registration. Last, tell ddclient how to
-detect your IP address, your domain name and the update interval.
+In the configuration process, select "other" dynamic DNS service provider. 
+Next, tell ddclient to use the "dyndns2" protocol to perform updates.
+Enter ``update.dedyn.io`` as the dynamic DNS service. Afterwards, enter the domain 
+you want to update as the username.
```
peterthomassen, line 115: The order is arbitrary.

docs/dyndns/configure.rst
```diff
@@ -108,16 +109,21 @@ To install ddclient, run ``sudo apt-get install ddclient``. If a configuration
...
+As your password, use a token that you generated under ``https://desec.io/tokens``.
```
peterthomassen, line 116: For folks signing up with a .dedyn.io domain (currently suspended), a token is displayed after signing up. Telling them to create a token would confuse them.

docs/dyndns/configure.rst
```diff
@@ -108,16 +109,21 @@ To install ddclient, run ``sudo apt-get install ddclient``. If a configuration
...
+Tell ddclient how to detect your IP address. You can use any webservice on the list
```
peterthomassen, line 117: Not sure which list.

docs/dyndns/configure.rst
```diff
@@ -108,16 +109,21 @@ To install ddclient, run ``sudo apt-get install ddclient``. If a configuration
...
+or select ``others`` and then insert ``https://checkipv4.dedyn.io/`` to check your IPv4
+or ``https://checkipv6.dedyn.io/`` to check your IPv6.
+Insert the update interval and then the domain name to update.
 
 **Note:** As of the time of this writing, ddclient does not use an encrypted
 HTTPS connection by default. To enable it, open ``/etc/ddclient.conf`` and add
 ``ssl=yes`` above the ``server=`` statement. We **strongly recommend** doing
-so; otherwise, your credentials will be exposed during transmission.
+so; otherwise, your credentials will be exposed during transmission. Also webservices
+like ``https://checkipv4.dedyn.io/`` for IP checks depend on it!
```
peterthomassen, line 126: I'm not sure that checking with IP with a HTTPS URL breaks when you don't see `ssl=yes`. Isn't this setting for the update request?

peterthomassen (changes requested):
Thanks for the PR.

Not sure if these changes are improvements though; I added a few comments. I'm leaning towards closing this.

docs/dyndns/configure.rst
jameskimmel: I used the order of the ddclient wizard.

docs/dyndns/configure.rst
jameskimmel: hmm... so maybe something more generic?
Something like "use your access token as password"

docs/dyndns/configure.rst
jameskimmel: ddclient wizard offers you a list of webservices after you click on webservices

docs/dyndns/configure.rst
peterthomassen: Perhaps, but that's not significantly different from the current wording.

docs/dyndns/configure.rst
peterthomassen: The wizard is provided by the Linux distribution (not by ddclient), so it's not generic ddclient advice. Those who don't use your Linux distribution will be confused.

docs/dyndns/configure.rst
peterthomassen: That's not true; I just tested it. Doing an update via HTTPS using IP discovery via HTTPS works, and the correct IP address is discovered and set, even when `ssl=yes` is not configured.

docs/dyndns/configure.rst
jameskimmel: Ahhh, I am sorry, I did not know that. I thought it was the default for all ddclients. 

Removing "next" and "afterwards" sounds good to me.

docs/dyndns/configure.rst
jameskimmel: for the update.dedyn.io, you don't need it. 
But if you set 
`use=web, web=checkipv4.dedyn.io/ \`
without https, then you need 
`ssl=yes`

If you set 
`use=web, web=https://checkipv4.dedyn.io/ \`
you don't need it.

I also just tested this on an Ubuntu test machine

jameskimmel: and of course I don't know how other IP checkers handle this. Maybe someone besides you also forces you to use https for an IP check.

jameskimmel: Strange. 

I tested this on a Ubuntu machine. Then I wanted to implement this on my production debian host. 
On Debian even setting ssl=yes seems to not be enough.
I have to set https://checkipv4.dedyn.io, otherwise it will fail.

jameskimmel: Makes sense, removed the line. 
Added the test line form manual config. 
Feel free to delete it if you don't like it.

PR #945, opened 2024-07-11
error in update url

jameskimmel:
Wrong Update URL that will only update IPv6 instead of both.


peterthomassen (approved):
It is actually not an error, as both address families are specific in the query parameters, and will thus get updated. But there indeed is little motivation to use update6 here.

PR #946, opened 2024-07-11
split up examples

jameskimmel:
split up the different methods in the examples to make it more readable.


docs/dyndns/update-api.rst
```diff
@@ -172,12 +172,9 @@ affiliated with the respective account (see :ref:`manage-tokens` for details.)
 stand-in for an IPv6 address. Replace those (including the ``<`` and ``>``)
 with your respective values.
 
+**Examples with headers** (recommended)
```
jameskimmel, line 175: Not sure if I understood it correctly, but I think that the header version is more secure, since the token does not end up in your logs?

PR #953, opened 2024-07-22
feat(tests): use built-in typing

Rotzbua:
New since python 3.9:
https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections


peterthomassen (approved):
lgtm, thanks!

PR #954, opened 2024-07-23
small help and ssl update

jameskimmel:


docs/dyndns/configure.rst
```diff
@@ -113,9 +113,12 @@ enter ``update.dedyn.io`` as the dynamic DNS server. Next, tell ddclient to use
 the "dyndns2" protocol to perform updates. Afterwards, enter the username and
 the token secret that you received during registration. Last, tell ddclient how to
 detect your IP address, your domain name and the update interval.
+To test your setup, run ``sudo ddclient -force`` and see if everything works as
+expected.
 
 **Note:** As of the time of this writing, ddclient does not use an encrypted
-HTTPS connection by default. To enable it, open ``/etc/ddclient.conf`` and add
+HTTPS connection by default when the scheme (either "http://" or "https://") is
```
jameskimmel, line 120: https://github.com/ddclient/ddclient/pull/705

peterthomassen:
Thank you for your contribution!

"by default" implies "when the scheme is missing", so this does not seem to add anything.

jameskimmel:
It adds that if I set https in the URL, I don't have to enable ssl in the conf file.

PR #996, opened 2024-12-02
fix(api): fix trigger order on token CASCADE policy deletion, fixes #995

### LINKED ISSUE #995: Bug: Can not delete tokens with auto_policy
I just played around with the new token `auto_policy` attribute. It seems that tokens with `auto_policy` enabled can not be deleted. The API just returns status code 500 and leaves the token in place.
After setting the attribute to `false`, I could delete the token as expected.

peterthomassen:


api/desecapi/models/tokens.py
```diff
@@ -65,7 +65,7 @@ class Meta:
             # Ensure that a default policy is defined when auto_policy=true
             pgtrigger.Trigger(
                 name="token_auto_policy",
-                operation=pgtrigger.Update | pgtrigger.Insert,
+                operation=pgtrigger.Insert | pgtrigger.UpdateOf("auto_policy"),
```
peterthomassen, line 68: This one ensures that `auto_policy` checks don't run when it's not being changed (e.g., when only changing `last_used` during authentication)

api/desecapi/models/tokens.py
```diff
@@ -234,7 +234,8 @@ class Meta:
             pgtrigger.Trigger(
                 name="default_policy_when_auto_policy",
                 operation=pgtrigger.Delete,
-                when=pgtrigger.Before,
+                when=pgtrigger.After,
+                timing=pgtrigger.Deferred,
```
peterthomassen, line 238: This one is the actual fix

nils-wisiol (approved):

api/desecapi/tests/test_token_domain_policy.py
```diff
@@ -752,3 +752,15 @@ def test_create_domain(self):
         self.assertTrue(self.token.get_policy(rrset).perm_write)
         rrset = models.RRset(domain=models.Domain.objects.get(name=name_other))
         self.assertFalse(self.token.get_policy(rrset).perm_write)
+
+    def test_delete_domain_with_autopolicy(self):
```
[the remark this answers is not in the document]
peterthomassen: done

PR #1004, opened 2024-12-25
feat(www): optional hybrid PQC key exchange for TLS connections

nils-wisiol:
Enables key agreement for TLS connections of www using hybrid PQC schemes (ECDHE + ML-KEM). Does not remove support for pure classical schemes (ECDHE only). Does not remove support for non-elliptic-curve DHE-RSA key agreement.

Hybrid PQC schemes using ML-KEM are supported from Chrome 131 onward; Firefox from 132 onward.

### DIFF
--- a/www/Dockerfile
+++ b/www/Dockerfile
@@ -1,3 +1,20 @@
+FROM alpine AS oqs-provider
+
+RUN apk add --no-cache cmake git g++ make ninja openssl openssl-dev
+
+ARG LIBOQS_VERSION="0.12.0"
+ARG OQSPROVIDER_VERSION="0.8.0"
+
+RUN if [ -z ${LIBOQS_VERSION} ]; then exit 1; fi;
+RUN if [ -z ${OQSPROVIDER_VERSION} ]; then exit 2; fi;
+
+RUN git clone --depth 1 --branch ${LIBOQS_VERSION} https://github.com/open-quantum-safe/liboqs.git
+RUN git clone --depth 1 --branch ${OQSPROVIDER_VERSION} https://github.com/open-quantum-safe/oqs-provider.git
+
+ARG liboqs_DIR=/.local
+RUN cd liboqs && cmake -GNinja -DCMAKE_INSTALL_PREFIX=$liboqs_DIR -S . -B _build && cd _build && ninja && ninja install
+RUN cd oqs-provider && cmake -S . -B _build && cmake --build _build && cmake --install _build
+
 FROM node:lts AS webapp
 RUN \
   apt-get update \
@@ -28,6 +45,19 @@ FROM nginx:mainline-alpine
 
 # Add dependencies for our scripts
 RUN apk add --no-cache bash openssl inotify-tools
+
+# configure openssl to use OQS provider
+COPY --from=oqs-provider /oqs-provider/_build/lib/oqsprovider.so /usr/local/lib/
+RUN cat <<EOF  >> /etc/ssl/openssl.cnf
+[provider_sect]
+default = default_sect
+oqsprovider = oqsprovider_sect
+[default_sect]
+activate = 1
+[oqsprovider_sect]
+activate = 1
+module = /usr/local/lib/oqsprovider.so
+EOF
 
 # nginx configuration and entrypoint
 COPY conf /etc/nginx
--- a/www/conf/conf.d/ssl.conf
+++ b/www/conf/conf.d/ssl.conf
@@ -1,6 +1,11 @@
-# According to https://ssl-config.mozilla.org/#server=nginx&version=1.16.1&config=intermediate&openssl=1.1.1c&guideline=5.4 and added PQC hybrids support
+# generated 2025-02-04, Mozilla Guideline v5.4, nginx 1.27.3, OpenSSL 3.3.2, intermediate config
+# https://ssl-config.mozilla.org/#server=nginx&version=1.27.3&config=intermediate&openssl=3.3.2&guideline=5.4 and added PQC hybrids support
 ssl_protocols TLSv1.2 TLSv1.3;
 ssl_ciphers ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384;
 ssl_prefer_server_ciphers off;
 ssl_dhparam /etc/nginx/dhparam.pem;
-ssl_ecdh_curve X25519MLKEM768:x25519_kyber768:p384_kyber768:x25519:secp384r1:x448:secp256r1:secp521r1;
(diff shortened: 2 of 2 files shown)


nils-wisiol:
Note that the Dockerfile hardcodes liboqs and oqs-provider versions which will fall through the cracks of our usual maintenance/upgrade checks. Not sure how to improve on that.

www/Dockerfile
```diff
@@ -1,3 +1,14 @@
+FROM alpine AS oqs-provider
+
+RUN apk add --no-cache cmake git g++ make ninja openssl openssl-dev
+
+RUN git clone --depth 1 --branch 0.12.0 https://github.com/open-quantum-safe/liboqs.git
+RUN git clone --depth 1 --branch 0.8.0 https://github.com/open-quantum-safe/oqs-provider.git
```
Martyrshot, line 6: To avoid hard coding these, I would use a [build arg](https://docs.docker.com/build/building/variables/). Then you can do something like:

```suggestion
ARG LIBOQS_VERSION=""
ARG OQSPROVIDER_VERSION=""

RUN if [ -z ${LIBOQS_VERSION} ]; exit 1; fi;
RUN if [ -z ${OQSPROVIDER_VERSION} ]; exit 2; fi;

RUN git clone --depth 1 --branch ${LIBOQS_VERSION} https://github.com/open-quantum-safe/liboqs.git
RUN git clone --depth 1 --branch ${OQSPROVIDER_VERSION} https://github.com/open-quantum-safe/oqs-provider.git
```

Then as part of the build step, or if you are using docker compose you can define it in the yaml file, you can define the version you want. The above should cause a build error if the versions aren't defined.

www/conf/conf.d/ssl.conf.var
```diff
@@ -1,8 +1,9 @@
-# According to https://ssl-config.mozilla.org/#server=nginx&version=1.16.1&config=intermediate&openssl=1.1.1c&guideline=5.4
+# According to https://ssl-config.mozilla.org/#server=nginx&version=1.16.1&config=intermediate&openssl=1.1.1c&guideline=5.4 and added PQC hybrids support
```
Martyrshot, line 1: I know this is just a comment pointing to a reference, but it should probably be updated to OpenSSL 3.2 at some point.

www/Dockerfile
```diff
@@ -29,6 +40,24 @@ FROM nginx:mainline-alpine
 # Add dependencies for our scripts
 RUN apk add --no-cache bash openssl inotify-tools
 
+# mountable content for web app (remove default stuff in there)
+RUN rm /usr/share/nginx/html/*
+COPY html/503.html /usr/share/nginx/html
+COPY --from=webapp /usr/src/app/dist /usr/share/nginx/html
```
peterthomassen, line 46: At least the last thing should run last, because it changes most frequently; no need to redo the OpenSSL config when updating webapp contents

www/Dockerfile
```diff
@@ -29,6 +40,24 @@ FROM nginx:mainline-alpine
...
+
+# configure openssl to use OQS provider
+COPY --from=oqs-provider /oqs-provider/_build/lib/oqsprovider.so /
+RUN cat <<EOF  >> /etc/ssl/openssl.cnf
```
peterthomassen, line 50: This is an append operation. Do we know what's in the file before this?

www/conf/conf.d/ssl.conf.var
peterthomassen: Good point!

www/Dockerfile
nils-wisiol: This is better, but still won't be caught by dependabot.

www/Dockerfile
nils-wisiol: Redoing openssl config takes <1s, while updateing the webapp takes minutes; hence this order is best for runtime (but not for number of computed layers).

www/Dockerfile
nils-wisiol: openssl's config as shipped by Debian.

nils-wisiol:
This is ready for an iteration of reviews. The fixup commit is to be squashed into the first commit before merging.

www/conf/conf.d/ssl.conf.var
peterthomassen: Comparing to the recommendations applicable to our current setting (https://ssl-config.mozilla.org/#server=nginx&version=1.27.3&config=intermediate&openssl=3.3.2&ocsp=false&guideline=5.4), it looks like our `ssl_ecdh_curve` is lacking `prime256v1`. Is that intentional?

(Also let's update the link in the comment.)

www/Dockerfile
peterthomassen: How do we keep this up to date? (Is there a way to get notified upon new releases, perhaps through GitHub?)

www/Dockerfile
```diff
@@ -1,9 +1,13 @@
 FROM alpine AS oqs-provider
 
-RUN apk add --no-cache cmake git g++ make ninja openssl openssl-dev
```
peterthomassen: This doesn't look right (`docker run --rm alpine cmake` says `cmake` isn't found)

www/Dockerfile
```diff
@@ -1,9 +1,13 @@
...
+ARG LIBOQS_VERSION="0.12.0"
+ARG OQSPROVIDER_VERSION="0.8.0"
 
-RUN git clone --depth 1 --branch 0.12.0 https://github.com/open-quantum-safe/liboqs.git
-RUN git clone --depth 1 --branch 0.8.0 https://github.com/open-quantum-safe/oqs-provider.git
+RUN test -z ${LIBOQS_VERSION} && exit 1
```
peterthomassen, line 6: CI fails with
```
#105 ERROR: process "/bin/sh -c test -z ${LIBOQS_VERSION} && exit 1" did not complete successfully: exit code: 1
```

peterthomassen (changes requested):

www/Dockerfile
nils-wisiol: I could not find a way to automate this. I've set a reminder to do it manually every couple of weeks.

www/conf/conf.d/ssl.conf.var
nils-wisiol: Left out `prime256v1` by mistake. The comment is already updated.

www/Dockerfile
nils-wisiol: `docker run --rm alpine apk add cmake` shows successful installation for me

peterthomassen: yes, but you're removing this line!

www/Dockerfile
nils-wisiol: @peterthomassen is right - the commands in fact do not compile the webapp. I've undone this change via a fixup commit.

www/Dockerfile
nils-wisiol: You're right! I got confused by Github's green check mark.

nils-wisiol:
Rebased and squashed. Ready for new review.

www/Dockerfile
```diff
@@ -29,6 +46,19 @@ FROM nginx:mainline-alpine
 # Add dependencies for our scripts
 RUN apk add --no-cache bash openssl inotify-tools
 
+# configure openssl to use OQS provider
+COPY --from=oqs-provider /oqs-provider/_build/lib/oqsprovider.so /
```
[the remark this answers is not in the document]
nils-wisiol: Fixed. The Github tests didn't ran, but I confirmed locally that the www image builds and is functional with this change.

PR #1014, opened 2025-01-25
resurrect dedyn.io registrations

peterthomassen:


peterthomassen:
After this is deployed, the plan is to send out messages like the following (with bullet points included as appropriate):

```
Dear deSEC user,

We have noticed that you have modified the NS records of your dedyn.io domain(s) in a way that is technically not meaningful. The modifications are not compatible with the automatic DNSSEC services that we provide, in line with our mission to improve Internet security.

You might have changed these records with the intention to delegate your domain(s) to a different DNS operator. However, domains under dedyn.io cannot be moved to a different operator. This is because each domain is tied to deSEC's cryptographic keys, which means that DNS responses from another operator are invalid, leading to DNS resolution failures.

To remediate the situation, we will implement the following changes:

- Starting immediately, it is no longer possible to modify NS records for
  domains under dedyn.io.

- The main NS records of the following domains have been replaced completely,
  and they don't contain any deSEC nameservers anymore. As a result, they are
  dysfunctional:

    * <masked>.dedyn.io:
        igor.ns.cloudflare.com.
        kinsley.ns.cloudflare.com.

    * <masked>.dedyn.io:
        henrik.ns.cloudflare.com.
        kia.ns.cloudflare.com.

    * <masked>.dedyn.io:
        bethany.ns.cloudflare.com.
        sullivan.ns.cloudflare.com.

  The above domain(s) will be deleted during the week of February 10, 2025.
  If your account does not contain any other domains at this time, we will
  delete your account as well.

- The NS records of the following domains have been modified such that they
  contain some deSEC nameserver(s), but also some other nameserver(s):

    * <masked>.dedyn.io:
        ns1.desec.io.
        ns1.desec.org.  (unofficial)

  During the week of February 10, 2025, we will replace these NS records with
  our standard NS records (ns1.desec.io and ns2.desec.org).

Thank you for your understanding! If you have any questions, please let us know.

You are welcome to use deSEC in the future. However, please note that deSEC does not support third-party DNS operators for domains under dedyn.io; we will make no exceptions. To use another operator, please register a domain name elsewhere.
```

api/desecapi/tests/base.py
```diff
@@ -1338,8 +1344,9 @@ def _test_rr_sets(cls, subname=None, type_=None, records=None, ttl=None):
     def setUp(self):
         super().setUp()
         # TODO this test does not cover "dyn" / auto delegation domains
-        self.my_empty_domain = self.create_domain(suffix="", owner=self.owner)
-        self.my_rr_set_domain = self.create_domain(suffix="", owner=self.owner)
+        domain_kwargs = {"suffix": self.AUTO_DELEGATION_DOMAINS if self.DYN else None}
+        self.my_empty_domain = self.create_domain(owner=self.owner, **domain_kwargs)
+        self.my_rr_set_domain = self.create_domain(owner=self.owner, **domain_kwargs)
```
nils-wisiol, line 1349: 1. Remove TODO in line above
2. Use `suffix=self.AUTO_DELEGATION_DOMAINS if self.DYN else None` instead of creating a kwargs object with a single entry

api/desecapi/tests/test_rrsets.py
```diff
@@ -1636,3 +1638,131 @@ def test_create_my_rr_sets_ip_block(self):
...
+    def test_delete_ns_rrset_nonapex(self):
+        data = dict(self.ns_data, subname="sub", records=[])
+        for method in (self.client.patch_rr_set, self.client.put_rr_set):
+            self.create_rr_set(
+                self.my_domain, settings.DEFAULT_NS, subname="sub", type="NS", ttl=3600
+            )
+            with self.assertRequests(
+                self.requests_desec_rr_sets_update(name=self.my_domain.name)
+            ):
+                response = method(self.my_domain.name, "sub", "NS", data)
```
nils-wisiol, line 1728: subname arg seems redundant - once as argument to `method` and once in `data`.

api/desecapi/tests/test_rrsets.py
```diff
@@ -1625,6 +1625,8 @@ def assertRequests(*, allowed):
 class AuthenticatedRRSetLPSTestCase(AuthenticatedRRSetBaseTestCase):
     DYN = True
```
nils-wisiol, line 1657: do we have tests for `DYN = False`?

api/desecapi/serializers/records.py
```diff
@@ -508,14 +508,28 @@ def _validate_blocked_content(self, attrs, type_):
...
+        if (
+            type_ == "NS"
+            and self.domain.is_locally_registrable
+            and (
+                attrs.get("records", True)
+                or not attrs.get("subname", self.instance.subname)
+            )
+        ):
+            raise serializers.ValidationError(
+                {"type": ["Cannot modify NS records for this domain."]}
```
nils-wisiol, line 537: Be more specific? I.e., deletion is allowed under circumstances?

api/desecapi/views/records.py
```diff
@@ -99,6 +99,10 @@ def update(self, request, *args, **kwargs):
         return response
 
     def perform_destroy(self, instance):
+        # Disallow modification of apex NS RRset for locally registrable domains
+        if instance.type == "NS" and self.domain.is_locally_registrable:
+            if instance.subname == "":
+                raise ValidationError("Cannot modify NS records for this domain.")
```
nils-wisiol, line 105: This validation is in the view layer, whereas other validations in this PR are in the serializer layer. For better maintenance, related validations should be located in the same layer.

api/desecapi/tests/test_rrsets.py
```diff
@@ -1636,3 +1638,131 @@ def test_create_my_rr_sets_ip_block(self):
         )
         self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
         self.assertIn("IP address 3.2.2.5 not allowed.", str(response.data))
+
+    def test_create_ns_rrset(self):
```
nils-wisiol, line 1673: I've used delegation of dedyn subdomains myself and found it pretty useful for testing purposes, e.g. for the pqc testbench. Do we have to disallow it?

nils-wisiol (commented):
Some concerns around policy. Nits on code style.

api/desecapi/serializers/records.py
peterthomassen: I'd rather not, as an NS RRset can't be created in the future, so deletion applies to legacy scenarios only. We don't need to inform people who're trying to change the apex NS with info about deleting legacy NS RRsets.

api/desecapi/tests/test_rrsets.py
peterthomassen: yes, that's the default, and the tests are in class `AuthenticatedRRSetTestCase`

api/desecapi/tests/test_rrsets.py
peterthomassen: The load problems last year stemmed from people trying to delegate their domains to Cloudflare, at the apex. It is my suspicion that if we only disallow *that*, we'll again soon have tens of thousands of registrations that then delegate a subdomain to Cloudflare. We're not prepared to handle these numbers, and I'm worried about that / would like to put stability first.

There are currently only very few lower-level delegations under LPS domains, and those continue to work. If it's ever needed, we can (1) do it in the backend (the model allows it, just not the REST endpoint), (2) add a flag that would allow certain accounts, or (3) use a custom domain.

What do you think?

api/desecapi/tests/test_rrsets.py
peterthomassen: No, PUT requires sending the full object

api/desecapi/views/records.py
peterthomassen: Deletion does not go through the serializer in DRF, except on bulk endpoints. That's why it's different (not our decision).

peterthomassen:
I've addressed all comments, except one (pls continue conversation).

Would also be interested in your opinion on https://github.com/desec-io/desec-stack/pull/1014#issuecomment-2617221470 (including "no opinion").

api/desecapi/tests/test_rrsets.py
nils-wisiol: Resuming operation for a few (or even most) users is better than resuming for none - so let's go for it the way it is first.

For the record, another idea could be to somehow (2b) privilege accounts that have at least one delegated domain - but extend that privilege to all domains associated with that account.

nils-wisiol:
For the email you propose - I'm afraid it will take a while to craft the emails and it might generate a bulk of support workload. I was going to propose to silently remove or fix these but Sec. 4 of the Terms mentions there "may be" a removal warning.

If you share my concern, I think fixing without warning is an acceptable way forward. I don't think we need to support dysfunctional (or slow) resolution of certain names. Removing without warning is also a possibility but perhaps a bit of a stretch of the terms (which could be clarified in this regard).

If you do not share the concern (or think the benefits outweigh), my only comment on the text is that I believe most users that are affected aren't well-trained in DNS things. To ease their understanding, I'd suggest to simplify the language and shorten it.

nils-wisiol (approved):

peterthomassen:
> For the email you propose - I'm afraid it will take a while to craft the emails

I already have a script for that, which works off a database query CSV output.

> and it might generate a bulk of support workload.

I'm prepared to take that :)

> If you share my concern, I think fixing without warning is an acceptable way forward.

Ack

> Removing without warning is also a possibility but perhaps a bit of a stretch of the terms (which could be clarified in this regard).

I'd rather not do apply that retroactively.

> If you do not share the concern (or think the benefits outweigh), my only comment on the text is that I believe most users that are affected aren't well-trained in DNS things. To ease their understanding, I'd suggest to simplify the language and shorten it.

I'll try, but if I don't have good ideas I'll leave (nearly) as is. I'll add a link to the forum so people can gather there and don' have to repeat their questions.

PR #1016, opened 2025-01-29
 chore(webapp): update vite & sass

Rotzbua:


peterthomassen (approved):
Thanks!

I wonder why dependabot hasn't caught this. The bot recently left a [note](https://github.com/desec-io/desec-stack/pull/1010#issuecomment-2613751311) that it doesn't support our npm version; however, we're [using `node:lts`](https://github.com/desec-io/desec-stack/blob/191c6ac984d26cef95d28d96fa5a150d9e7690be/www/Dockerfile#L1), so I was suprised by that note. Do you think this is related, or do you have any other thoughts?

peterthomassen:
Mh. It seems like dependabot's [support for npm version 6 was discontinued](https://github.blog/changelog/2024-12-19-dependabot-updates-ceases-supporting-npm-version-6-closing-down-private-preview/) on January 20.

However, according to https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch, it seems like node 18 comes with npm 10, so that can't be it. :shrug: 

On another note: Are you interested in working on #699? - If you reach out via email, we might also make a paid arrangement.

PR #1028, opened 2025-02-10
Webapp and Error Handling Improvements

peterthomassen:


www/webapp/src/components/ActivateAccountActionHandler.vue
```diff
@@ -3,16 +3,33 @@
     <div class="text-center" v-if="captcha_required && !success">
         <generic-captcha
             @update="(id, solution) => {setCaptchaPayload(id, solution)}"
-            tabindex="3"
+            tabindex="1"
             ref="captchaField"
         />
+        <v-layout class="justify-center">
+          <v-checkbox
+                v-model="terms"
```
nils-wisiol, line 11: It's interesting that we don't enforce terms agreement at the server side (both here and at the 'regular' signup form).

www/webapp/src/views/DonatePage.vue
```diff
@@ -134,6 +134,10 @@
                       <th>Ethereum (ETH)</th>
                       <td><span class="fixed-width">0xb9594A8d1BF512eAC59069d9d5cbA457E0d35998</span></td>
                     </tr>
+                    <tr>
+                      <th>Litecoin (LTC)</th>
+                      <td><strong>(No MWEB!)</strong> <span class="fixed-width">LeFYoRRcrMPiruRGQQknQBdjGdmsEpHzno</span></td>
```
nils-wisiol, line 139: haven't checked correctness of wallet id

docs/auth/account.rst
```diff
@@ -168,12 +168,10 @@ If email address and password match our records, the server will reply with
 
 As indicated in the response, login tokens expire 7 days after creation or
 when not used for 1 hour, whichever comes first (see :ref:`token object`).
+Expired login tokens are purged on each new login.
```
nils-wisiol, line 171: The login tokens that we keep is the closest thing to an audit log that we can offer. Should we really be deleting it unconditionally? Alternatives:

- offer an endpoint to remove them
- offer a user setting that controls if they are automatically deleted
- do not display them by default when querying the tokens endpoint

api/desecapi/debug.py
```diff
@@ -0,0 +1,12 @@
+from django.views.debug import ExceptionReporter
+
+
+class PayloadExceptionReporter(ExceptionReporter):
+    def get_traceback_data(self):
+        data = super().get_traceback_data()
+        if self.request is not None:
+            try:
+                data["request_meta"]["_body"] = self.request.body
+            except:
```
nils-wisiol, line 10: catching all exceptions is bad practice because it can hide problems and make debugging hard - what is the reason you want to do it here anyways?

www/webapp/src/components/ActivateAccountActionHandler.vue
peterthomassen: If I recall correctly, that's because the API is programmatic and can't enforce the terms text to be displayed next to the terms payload field boolean value.

api/desecapi/debug.py
peterthomassen: `request.body` is a property that either reads from a cached `._body` attribute or from a stream. Indeed, who knows what could go wrong.

On the other hand, we're already in error handling code for a *different* error. We just want to cleanly log it and then move away from it with a sane mind. I'm not sure if we should go into second-order error-handling issues.

Further, any issue in that `try` code is not our issue, but Django's. I'm not sure we could enumerate all exception types that potentially could occur, even more so should their code change.

Errors in `request.body` will be properly handled whenever it is used "normally" (outside of error handling) during request handling; they would then be a direct cause of exception handling and cause full logging (with traceback etc.).

docs/auth/account.rst
peterthomassen: I also doubt people would use this as an audit log, as there is a lack of understanding about these tokens in the first place. We've received a couple (10?) support request over time where people expressed concerns their accounts had been hacked and those have been created etc. An audit log would create transparency, not confusion. :)

api/desecapi/debug.py
nils-wisiol: Agree that its not realistic to draft and maintain a list of exceptions. What's more, upstream code for `ExceptionReporter` uses similar except clauses to determine other fields of the report.

So, I don't like this solution but also don't have a better way of doing it. We do need the body in the reports.

docs/auth/account.rst
nils-wisiol: Agree that it conflates concepts. Agree with user demand for deletion.

The audit comment wasn't aimed at users but at admins of deployments (specifically, you :) ).

nils-wisiol (approved):

PR #1042, opened 2025-03-05
Hopefully fix description when tokens are invalidated.

jpmens:
I have read each of these two sentences more than half a dozen times and still believe they are both wrong, hence the nit. :-)

### DIFF
--- a/docs/auth/tokens.rst
+++ b/docs/auth/tokens.rst
@@ -101,7 +101,7 @@ Field details:
     :Access mode: read, write
     :Type: string (time duration: ``[DD] [HH:[MM:]]ss[.uuuuuu]``) or ``null``
 
-    Maximum token age.  If ``created + max_age`` is less than the current time,
+    Maximum token age.  If ``created + max_age`` is greater than the current time,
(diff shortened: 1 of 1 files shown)


docs/auth/tokens.rst
```diff
@@ -101,7 +101,7 @@ Field details:
     :Access mode: read, write
     :Type: string (time duration: ``[DD] [HH:[MM:]]ss[.uuuuuu]``) or ``null``
 
-    Maximum token age.  If ``created + max_age`` is less than the current time,
+    Maximum token age.  If ``created + max_age`` is greater than the current time,
     the token is invalidated.  Invalidated tokens are not automatically deleted
```
peterthomassen, line 105: The expiration time of a token is its creation date plus the maximum age. The token is invalid if the expiration date is in the past, that is, if it is *less* than the current time.

peterthomassen:
Thanks for this PR, but I think you're wrong :-) See https://github.com/desec-io/desec-stack/pull/1042/files#r1980914152.

PR #1044, opened 2025-03-06
dynDNS domain usage limitations

peterthomassen:
This appears only when selecting the dynDNS option:

image


peterthomassen:
I think the text need some compression (so that they are more to the point, with less lines). I wanted to give a preview to discuss the general idea.

www/webapp/src/views/SignUp.vue
```diff
@@ -58,11 +58,11 @@
               <v-alert class="ml-8 mt-2" v-if="domainType == 'dynDNS'" type="info" outlined>
                 <p class="text-h6">Limitations of Domains Registered under {{ LOCAL_PUBLIC_SUFFIXES[0] }}</p>
```
fiwswe, line 59: Correct capitalization:
Domains -> domains
Registered -> registered

peterthomassen: We use [title case](https://en.wikipedia.org/wiki/Title_case) (see main page)

fiwswe: Fair enough, as long as it's‚ intentional 😉

peterthomassen:
I've optimized texts based on the types of support requests we're getting, and added a checkbox requiring acknowledgement of these limitations. Here's how the most recent commit looks like.

image

www/webapp/src/views/SignUp.vue
```diff
@@ -155,43 +196,62 @@
...
+        limitationsAccepted_rules: [v => (self.domainType !== 'dynDNS' || !!v)],
+      }
+    },
     async mounted() {
       if ('email' in this.$route.params && this.$route.params.email !== undefined) {
         this.email = this.$route.params.email;
       }
       await this.initialFocus();
+
+      const el = this.$refs['limitations'];
```
nils-wisiol, line 245: This doesn't look right. Have you tried using `v-if` on this DOM element instead?

peterthomassen: I've tried using `v-if`, but it makes the interface jump. UX is much smoother if the box is displayed with an expand animation. Unfortunately, CSS can't animate the `height` property from `0` to `auto`; if only works if the target height absolute value is provided. The workaround used here is commonly referenced e.g. on Stack Overflow for this problem.

How do you suggest to proceed? (Get rid of the animation?)

nils-wisiol (approved):
lgtm

PR #1047, opened 2025-03-16
Webapp: warn if subname ends with domain name

peterthomassen:
Previously, the hint was shown persistently in standard color (grey). It's now hidden by default, and appears in orange when subname ends with domain name. The message does not block form submission (it is not a validation error).

image


nils-wisiol (approved):
lgtm! <3

PR #1084, opened 2025-05-20
Bump Python version

peterthomassen:


peterthomassen:
I had planned to include a switch at the top to "hide log-in tokens", but I couldn't figure out how to set up a default filter that works even when nothing is typed in the "search" input field.

So I'll leave it as is; at least, it's now possible to identify all log-in tokens, and delete the old ones.

api/desecapi/tests/test_tokens.py
```diff
@@ -102,6 +102,28 @@ def test_update_my_token(self):
                 for k, v in data.items():
                     self.assertEqual(response.data[k], v)
 
+            datas = [
+                {"last_used": "2018-09-06T09:08:43.762697Z"},
+                {"last_used": None},
+            ]
+            for data in datas:
+                response = method(url, data=data)
+                self.assertStatus(response, status.HTTP_200_OK)
```
nils-wisiol, line 112: why do we expect 200? Confirm that writes to non-writeable fields get dropped silently

api/desecapi/serializers/tokens.py
```diff
@@ -52,6 +53,7 @@ class Meta:
             "id",
             "created",
             "last_used",
+            "mfa",
```
nils-wisiol, line 56: add a test for non-writeability. Since this field is high-stakes and different type (i.e., perhaps different code path) from the tested ones, I think an additional test is warranted.

nils-wisiol (approved):

PR #1106, opened 2025-06-29
fix(docs): token policies are created by POST

rom4nik:
87a98e8b8a834d9224a4ce3900c63d0bb7652cf2 made it a bit less obvious that this must be a POST request, in case it's sent with a tool other than curl. :sweat_smile:


peterthomassen (approved):
thanks, good point! lgtm.

PR #1112, opened 2025-07-26
Allow updating multiple subdomains in single dyndns request

### LINKED ISSUE #13: Update IPv6 subnet on several subdomains with one request
Currently, users can update there IPv4 and IPv6 addresses for their domain.dedyn.io. IPv6 users, however, often control a dynamic subnet and may want to run services on various IP addresses within this subnet.

It would be nice to allow users to create subdomains with static "local IP bits", only updating the subnet through our update mechanism. The user would have to specify the subnet size (e.g. `/80`), and upon IP update, only the first 80 bits of the AAAA records would be updated.

### LINKED ISSUE #187: dyn: support multiple hostname update

molikuner:
# What does this PR do?
This PR introduces the possibility to update multiple subdomains in a single dyndns request.

## Multi-Domain Updates
The update endpoint now allows passing a comma separated list of domains in places of the `hostname` and `host_id` request parameter, http basic auth username and the `username` request parameter. This allows updating `www.example.com` and `host.example.com` at the same time using a single update request, e.g. if for whatever reason a CNAME record can't be used.

## Specific-Domain Overwrite
In addition, the update endpoint now supports parameters overwriting the IP-address for a specific domain. E.g. it would be possible to update `www.example.com`, `host.example.com` and `other.example.com` with the same IPv4 address, but overwrite the IPv6 address for `other.example.com` in a single request. This is achieved by supporting new request parameters in the form of `{domain}.{parameter-name}`, e.g. the parameter `other.example.com.ipv6` could be used to specify the IPv4 address of `other.example.com`, while the other updated domains fallback to the value given in `ipv6`.

## Example
A full request could look like this (assuming you own `example.com`):

```http
GET https://update.dedyn.io/
  ?hostname=www.example.com,host.example.com,other.example.com
  &ip=1.2.3.4
  &ipv6=2001:1::/64
  &other.example.com.ipv6=2001:2::1
Authorization: Token <update-token>
```

This would do the following updates:

| Domain | IPv4 | IPv6 |
| -------- | ---- | ---- |
| www.example.com | 1.2.3.4 | 2001:1::/64 |
| host.example.com | 1.2.3.4 | 2001:1::/64 |
| other.example.com | 1.2.3.4 | **2001:2::1** |

# How is this implemented?
This change was implemented in two parts:

1. Supporting Multi-Domain Updates
2. Supporting Specific-Domain Overwrites

## Supporting Multi-Domain Updates
This change adjusts the `DynDNS12UpdateView` to support parsing multiple subdomains from the request by splitting the domain name by comma. Afterwards it checks whether all given subdomains belong to a single domain. If that's not the case, a 400 bad request is returned to the caller. Internally the view had to now do more than one update in a single transaction.

## Supporting Specific-Domain Overwrites
This change again adjusts the `DynDNS12UpdateView` to search the request parameters for the new `{domain}.{parameter-name}` parameters and respect them in case they were provided. In case none of the new parameters was provided, a fallback to the "global" `{parameter-name}` parameters is used. Thus this change is backwards compatible and could technically be released also at a later point in time.

# References

 * Partially resolves #13, as it's possible to update multiple subdomains, but not across domains.
 ** Though I would argue that this will solve 99% of the requests. E.g. people transitioning from dynv6.com, which was mentioned multiple times in the issue, would be able to fully replicate their update since all updated subdomains belong to a single domain in their API.
 * Partially resolves #187, as it's possible to update multiple subdomains, but not across domains.

# Further Considerations

This PR contains 2 commits to make the review easier. If required it could also be split into different PRs.


peterthomassen:
First, thanks for this great and significant contribution!

Unfortuantely, while `PDNSChangeTracker` support updating multiple RRsets (including at multiple `subname` values) at once (cf. REST API bulk requests), it does not support updating RRsets of multiple distinct domains at once. This is because PowerDNS does a commit per domain; it's not possible to keep the transaction open until all domains' updates are processed.

Thus, this PR either
- would have to be restricted to one domain (but continue to allow multiple `subname` values), or
- would have to be add multi-domain transaction support to the platform. We have a long-term plan of migrating from PowerDNS to Knot, which is expected to support this, but that will require a major refactor.

Note that in the first case, the interface / URL syntax could remain untouched, but a 4xx response would have to be returned if FQDNs transcend the scope of one domain.

In the second case, the migration would specifically have to be for the `nslord` component. Before we do that, we'd like to practice the transition by migration `nsmaster` (which is also more urgent for replication performance reasons) - in other words, that's really a long-term project.

molikuner:
> This is because PowerDNS does a commit per domain; it's not possible to keep the transaction open until all domains' updates are processed.

I see, I guess I could have noticed this while writing the tests, when multiple update requests are made to PowerDNS from a single change, e.g updating multiple domains (e.g. `test_update_multiple_v4`) vs. updating domain + subdomain (e.g. `test_update_and_create`). Now my question would be: Is it a problem?

I could imagine that for some reason the second update fails and thus there would be difference between `dbapi` and `nslord` as the first update would be rolled back on `dbapi`, but kept on `nslord`. I could imagine some other ways where differences between `dbapi` and `nslord` could occur, even without these changes, e.g. when an update is accepted by `nslord`, but because of network issues `api` considers the update failed and triggers the rollback. Is there any way such differences would be noticed at the moment and "heal" automatically?

> Note that in the first case, the interface / URL syntax could remain untouched, but a 4xx response would have to be returned if FQDNs transcend the scope of one domain.

Sounds reasonable, in case we need to introduce this restriction for now.

> In the second case, the migration would specifically have to be for the `nslord` component. ... in other words, that's really a long-term project

Yeah, that's definitely not something I would be able to do. Sounds like an interesting project though!

Side note: Unfortunately I have problems getting the whole project to run locally, thus I wasn't really able to test the change fully (apart from the unit tests).

renne:
@molikuner How would this work with a german full dual-stack internet socket?
On IPv6-prefix updates the router (e.g. a Fritz!Box) hands over two parameters in a DynDNSv2 update-request:
1. The public global IPv6-prefix routed to the hosts in the LAN
2. The public global IPv6-address of itself (not inside the /64-prefix but e.g. in the /56-prefix)

How do you aggregate the combination of the IPv6-prefix + device identifier (either EUI64 or manually set) for each AAAA-record?

Your proposal seems to be limited to one IPv6-address for all hosts.

molikuner:
Hi @renne,

If this PR is accepted, you should be able to configure the Fritz!Box using this:

```
Update URL: https://update.dedyn.io/?hostname=<domain>&password=<pass>&ip=<ipaddr>&DOMAIN_OF_FRITZ_BOX.ipv6=<ip6addr>&ipv6=<ip6lanprefix>
Domainname: DOMAIN_OF_FRITZ_BOX,OTHER_DOMAIN,YET_ANOTHER_DOMAIN
Password: TOKEN
```

This should update `OTHER_DOMAIN` and `YET_ANOTHER_DOMAIN` with the IPv6 prefix that it received (preserving the previous interface identifier of the IPv6 address), and the `DOMAIN_OF_FRITZ_BOX` with the specific IPv6 address that it received.

molikuner:
> sorry for the extra work!
No worries at all! I've just updated the PR to restrict updates to a single domain, still supporting updates of multiple subdomains as suggested. I've also adjusted the PR description accordingly.

api/desecapi/views/dyndns.py
```diff
@@ -151,30 +219,68 @@ def get_serializer_context(self):
...
-                "ttl": 60,
-                "records": ip_params,
-            }
-            for type_, ip_params in record_params.items()
-            if "preserve" not in ip_params
-        ]
-
-        serializer = self.get_serializer(instances, data=data, many=True, partial=True)
+        data = []
+        for qname, subname in zip(self.qnames, self.subnames):
```
peterthomassen, line 284: It seems like you don't need `qname` here

docs/dyndns/configure.rst
```diff
@@ -174,10 +174,15 @@ expected.
...
-You can use do that either via the web interface or the API.
-
-If you try to update several subdomains directly (by issuing multiple update
-requests), your update requests may be refused (see :ref:`rate-limits`).
+To have multiple domain names point to the same dynamic IP address, it is best
+to designate one of them as the main domain for updates and create CNAME records
+for the others. This way, the other domains will automatically resolve to the same
+IP address as the main domain.
+
+Alternatively, for cases where a CNAME is not suitable (e.g. at the zone apex),
```
peterthomassen, line 182: If multiple names are to be updated (within the same domain), then only max. one of them is at the apex, so this can always be the main domain to which the others point via CNAME. So this is not a good example for where CNAME is not suitable. What would be a good use case?

docs/dyndns/configure.rst
```diff
@@ -174,10 +174,15 @@ expected.
 
 Updating Multiple Domains
 `````````````````````````
-To update multiple domain or subdomains, it is best to designate one of them
-as the main domain, and create CNAME records for the others, so that they act
-as DNS aliases for the main domain.
-You can use do that either via the web interface or the API.
-
-If you try to update several subdomains directly (by issuing multiple update
-requests), your update requests may be refused (see :ref:`rate-limits`).
```
peterthomassen, line 183: I'd like to retain the reference to rate limiting if you try updating using individual requests.

docs/dyndns/configure.rst
```diff
@@ -174,10 +174,15 @@ expected.
...
+for the others. This way, the other domains will automatically resolve to the same
+IP address as the main domain.
+
+Alternatively, for cases where a CNAME is not suitable (e.g. at the zone apex),
+it is possible to update multiple hostnames (such as ``example.com`` and ``sub.example.com``)
+in a single API call, as long as they all belong to the same domain. It is not
+possible to update hostnames belonging to different domains (e.g. ``example.com``
+and ``example.org``) in the same request.
+
+For technical details, please see the :ref:`update-api` documentation.
```
peterthomassen, line 188: Shouldn't we have an example here, e.g., for ddclient?

peterthomassen (changes requested):
Thanks, this looks quite promising!

I've only reviewed the first commit, and will review the second once the first is stable. (I also haven't look at tests yet; will do in the next iteration.)

api/desecapi/views/dyndns.py
```diff
@@ -151,30 +219,68 @@ def get_serializer_context(self):
         }
 
     def get_queryset(self):
-        return self.domain.rrset_set.filter(
-            subname=self.subname, type__in=["A", "AAAA"]
+        return (
+            self.domain.rrset_set.filter(
+                subname__in=self.subnames, type__in=["A", "AAAA"]
+            )
+            .order_by("subname")
```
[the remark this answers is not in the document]
peterthomassen: I did not know that, interesting. - I guess it's fine to have the database do the sorting.

peterthomassen:
> I'll try to fix the issues in the coming days.

Thanks. Please comment here explicitly when done, so I know when to take a look.

> Would be nice to already get some responses on some followup questions that I have added in the corresponding threads.

See above! Let me know if anything else is unclear.

molikuner:
@peterthomassen I think I've addressed all your comments and updated the code accordingly. Please have another look 😄

api/desecapi/views/dyndns.py
```diff
@@ -132,16 +187,29 @@ def qname(self):
...
-                self.qname, owner=self.request.user
-            ).order_by("-name_length")[0]
+            domains = list(
+                {
+                    Domain.objects.filter_qname(
+                        qname, owner=self.request.user
+                    ).order_by("-name_length")[0]
+                    for qname in self.qnames
+                }
+            )
```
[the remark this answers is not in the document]
peterthomassen: I'm thinking it through; not sure if it can be done without the union. But, out of interest - why is the window needed? Couldn't you do
```python
qsets = [
    Domain.objects.filter_qname(qname, **kwargs).order_by("-name_length")[:1]
    for qname in qnames
]
```
... and then take the union of that?

api/desecapi/views/dyndns.py
```diff
@@ -130,18 +183,28 @@ def qname(self):
...
-            return Domain.objects.filter_qname(
-                self.qname, owner=self.request.user
-            ).order_by("-name_length")[0]
-        except (IndexError, ValueError, Domain.DoesNotExist):
+    def domain(self) -> Domain:
+        domains = Domain.objects.filter_qnames(
+            self.qnames, owner=self.request.user
+        ).all()
+
+        if len(domains) != len(self.qnames):
```
peterthomassen, line 205: `domains` is that result of an SQL union. According to the docs, ["The `UNION` operator selects only distinct values by default. To allow duplicate values, use the `all=True` argument"](https://docs.djangoproject.com/en/dev/ref/models/querysets/#django.db.models.query.QuerySet.union).

Originally I thought that if the subnames all belong to the same domain, this condition would usually fail. Then I noticed that the queryset is annotated with `dotted_qname`, which will differ for each subname, and probably that's why it works. - Do you agree?

If so, that would break if the queryset would no longer return that field (which is really only there because it's left over from constructing the query result). But, in the next check, you count how many different primary keys there are. Perhaps that would be a safer bet here (checking against `== 1`)?

I'll be happy to make the change myself (I have a few minor stylistics things lined up locally already); just aligning with you before I go ahead. Thanks!

api/desecapi/views/dyndns.py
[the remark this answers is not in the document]
molikuner: That's a good idea. I don't have time right now to change it. Please feel free to do the change together with your other stylistic changes. Otherwise I'll do it in the next few days.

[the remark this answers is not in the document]
molikuner: @peterthomassen I've updated the code with this suggestion. Please feel free to push your stylistic changes and/or review again 👍

api/desecapi/models/records.py
```diff
@@ -56,24 +56,17 @@
...
-        ).records.all()
-    except ObjectDoesNotExist:
-        records = []
     return [
         str(
             ip_address(int(ip_address(record.content)) & int(subnet.hostmask))  # suffix
             + int(subnet.network_address)  # prefix
         )
         for record in records
+        if type(ip_address(record.content)) is type(subnet.network_address)
```
nils-wisiol, line 69: Do not use `type` to compare types. Use `isinstance` instead. (https://docs.quantifiedcode.com/python-anti-patterns/readability/do_not_compare_types_use_isinstance.html)

`if isinstance(ip_address(record.content), type(subnet.network_address))`

api/desecapi/tests/test_dyndns12update.py
```diff
@@ -296,6 +296,191 @@ def test_subnet(self):
         self.assertEqual(response.data, "good")
         self.assertIP(ipv6="2a01::3303:72dc:f412:7233", subname="foo")
 
+    def test_update_multiple_v4(self):
+        # /nic/update?hostname=a.io,sub.a.io&myip=1.2.3.4
+        new_ip = "1.2.3.4"
+        domain1 = self.my_domain.name
+        domain2 = "sub." + self.my_domain.name
+
+        with self.assertRequests(
```
nils-wisiol, line 319: why is there no request regarding `domain2`?

nils-wisiol (approved):
Great work, thanks! Looking forward to have this deployed. Found a couple of nits.

api/desecapi/models/records.py
peterthomassen: done

api/desecapi/tests/test_dyndns12update.py
```diff
@@ -296,6 +296,191 @@ def test_subnet(self):
...
+                {"hostname": f"{domain1},{domain2}", "myip": new_ip},
+            )
+
+        self.assertStatus(response, status.HTTP_200_OK)
+        self.assertEqual(response.data, "good")
+
+        self.assertIP(ipv4=new_ip)
+        self.assertIP(subname="sub", ipv4=new_ip)
+
+    def test_update_same_domain_twice(self):
```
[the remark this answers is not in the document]
peterthomassen: It's only one request because the hostnames get deduplicated (via a set)

api/desecapi/tests/test_dyndns12update.py
peterthomassen: Because `domain2` is a subdomain of `domain1`, so it's not a cross-domain update and done atomically in one backend transaction.

The other case (cross-domain) is not supported by our backend and tested in `test_update_multiple_with_subdomains_of_different_domains` (returns `400`).

api/desecapi/tests/test_dyndns12update.py
```diff
@@ -296,6 +310,191 @@ def test_subnet(self):
...
+        domain1 = "sub1." + self.my_domain.name
+        domain2 = "sub2." + self.my_domain.name
+
+        with self.assertRequests():
+            response = self.client.get(
+                self.reverse("v1:dyndns12update"),
+                {"hostname": f"{domain1},{domain2}", "myip": "1.2.3.4/64"},
+            )
+
+        self.assertStatus(response, status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
peterthomassen: done

molikuner:
The follow-up PR can be found at #1141.

schnz:
> Hi @renne,
> 
> If this PR is accepted, you should be able to configure the Fritz!Box using this:
> 
> ```
> Update URL: https://update.dedyn.io/?hostname=<domain>&ip=<ipaddr>&fritz.my-domain.de.ipv6=<ip6addr>&ipv6=<ip6lanprefix>
> Domainname: fritz.my-domain.de,host.my-domain.de,mail.my-domain.de
> Username: my-domain.de
> Password: TOKEN
> ```
> 
> This should update `host.my-domain.de` and `mail.my-domain.de` with the IPv6 prefix that it received (preserving the previous interface identifier of the IPv6 address), and the `fritz.my-domain.de` with the specific IPv6 address that it received.
> 
> EDIT: replaced domain names for clarity

@molikuner  Sorry to dig this up. I just stumbled upon this use-case and I am glad you put the effort into this PR :).

Is the quotation still accurate? Because I can't seem to get it working. Suppose I have the following DNS state:

| Type | Subname                           | Content                                 |
| ---- | --------------------------------- | --------------------------------------- |
| A    | location.dyndns.mydomain.tld      | 22.33.44.55                             |
| AAAA | location.dyndns.mydomain.tld      | fdac:d6ab:301d:1db2:b6fc:7dff:fe12:5dcd |
| A    | host.location.dyndns.mydomain.tld | 22.33.44.55                             |
| AAAA | host.location.dyndns.mydomain.tld | fd0b:ef5b:54c7:0:c505:413d:8ba4:e4ee    |

Now, suppose my router (`location.dyndns.mydomain.tld`) gets a new IPv4 and v6 address and also a new IPv6 network for delegation purposes:

```sh
# line breaks are added for the sake of readability

curl -4 --user dyndns.mydomain.tld:mytoken \
  "https://update.dedyn.io/?hostname=location.dyndns.mydomain.tld,host.location.dyndns.mydomain.tld
  &ipv4=55.44.33.22
  &location.dyndns.mydomain.tld.ipv6=fd08:2d7e:b40a:6d00:fd61:75a:fe5b:d565
  &host.location.dyndns.mydomain.tld.ipv6=fdac:2812:b465::/48"
```

Then I would _expect_ the following state (changes in **bold**):
| Type | Subname                           | Content                                 |
| ---- | --------------------------------- | --------------------------------------- |
| A    | location.dyndns.mydomain.tld      | **55.44.33.22**                             |
| AAAA | location.dyndns.mydomain.tld      | **fd08:2d7e:b40a:6d00:fd61:75a:fe5b:d565** |
| A    | host.location.dyndns.mydomain.tld | **55.44.33.22**                             |
| AAAA | host.location.dyndns.mydomain.tld | **fdac:2812:b465**:0:c505:413d:8ba4:e4ee    |

However, this is the state I actually get (i.e. AAAA records are removed entirely)
| Type | Subname                           | Content                                 |
| ---- | --------------------------------- | --------------------------------------- |
| A    | location.dyndns.mydomain.tld      | **55.44.33.22**                             |
| A    | host.location.dyndns.mydomain.tld | **55.44.33.22**                             |

When using `curl -6` (as opposed to -4) then both AAAA records are updated to the IPv6 address of the requesting client. And when providing the `&ipv6=<ip6lanprefix>` parameter explicitly, then both AAAA records are updated according to the provided IPv6 prefix. In essence, it seems like the last two query parameters in the curl command above are ignored entirely.

molikuner:
Hey @schnz,

At the moment only the support for "Multi-Domain Updates" is implemented. The support for "Specific-Domain Overwrites" in the same request was moved into #1141 and isn't merged (yet).

Depending on your needs, you might be able to already replicate your desired setup, e.g. if it's fine for you to not have any IPv6 address for `location.dyndns.mydomain.tld`. Please checkout the comments in #1141 to see how to achieve this.

PR #1117, opened 2025-08-06
TOTP URI fix, logo updates

peterthomassen:


api/desecapi/models/mfa.py
```diff
@@ -56,7 +56,7 @@ def base32_secret(self):
     @property
     def uri(self):
         return self._totp.provisioning_uri(
-            name=self.name,
+            name=self.user.email,
```
nils-wisiol, line 59: Won't that break backwards compatibility?

nils-wisiol (commented):
Question pending

api/desecapi/models/mfa.py
peterthomassen: With what? The URI ist only conveyed once (and displayed in the QR code) when setting up the TOTP.

PR #1141, opened 2025-09-17
 Allow overwriting IP address for specific (sub)domain in dyndns request

molikuner:
# What does this PR do?
This PR introduces the possibility to set different IP addresses and/or networks for each updated (sub)domain in the dyndns update. It does so by allowing to overwrite request parameters for specific (sub)domains using a new `myipv4:<domain>` and `myipv6:<domain>` syntax.

Given the following request:
```
https://update.dedyn.io/
  ?hostname=host1.example.com
  &myip=1.2.3.4
  &myipv6:host2.example.com=2001:db8::1
```
The following IP addresses would be set after:

| domain | IPv4 address | IPv6 address |
| ------- | ------------- | ------------- |
| host1.example.com | 1.2.3.4 | none / discovered by client IP address |
| host2.example.com | preserved | 2001:db8::1 |

# Why is it needed?
This PR addresses most remaining concerns from #13, as it allows specifying specific addresses for specific (sub)domains. For example in Germany some ISPs provide an IPv6 network and an IPv6 address for the router that is outside of the given network (see for example [this comment in my previous PR](https://github.com/desec-io/desec-stack/pull/1112#issuecomment-3124568796)).

### DIFF
--- a/api/desecapi/views/dyndns.py
+++ b/api/desecapi/views/dyndns.py
@@ -76,53 +76,18 @@ class DynDNS12UpdateView(generics.GenericAPIView):
         as a fallback.
 
         Returns:
-            UpdateAction: A dataclass instance (`SetIPs`, `UpdateWithSubnet`, or `PreserveIPs`)
-            representing the action to be taken.
+            UpdateAction: A dataclass instance representing the action to be taken.
         """
         # Check URL parameters
         for param_key in param_keys:
             try:
-                params = set(
-                    filter(
-                        lambda param: separator in param or param in ("", "preserve"),
-                        map(str.strip, self.request.query_params[param_key].split(",")),
-                    )
-                )
+                param_value = self.request.query_params[param_key]
             except KeyError:
                 continue
-            if not params:
-                continue
 
-            try:
-                (param,) = params  # unpacks if params has exactly one element
-            except ValueError:  # more than one element
-                if params & {"", "preserve"}:
-                    raise ValidationError(
-                        detail=f'IP parameter "{param_key}" cannot have addresses and "preserve" or an empty value at the same time.',
-                        code="inconsistent-parameter",
-                    )
-                if any("/" in param for param in params):
-                    raise ValidationError(
-                        detail=f'IP parameter "{param_key}" cannot use subnet notation with multiple addresses.',
-                        code="multiple-subnet",
-                    )
-            else:  # one element
-                match param:
-                    case "":
-                        return SetIPs(ips=[])
-                    case "preserve":
-                        return PreserveIPs()
-                    case str(x) if "/" in x:
-                        try:
-                            subnet = ip_network(param, strict=False)
-                            return UpdateWithSubnet(subnet=subnet)
-                        except ValueError as e:
-                            raise ValidationError(
-                                detail=f'IP parameter "{param_key}" is an invalid subnet: {e}',
-                                code="invalid-subnet",
-                            )
-
-            return SetIPs(ips=list(params))
+            action = self._get_action_from_param(param_key, param_value, separator)
+            if action is not None:
+                return action
 
         # Check remote IP address
         client_ip = self.request.META.get("REMOTE_ADDR")
@@ -131,6 +96,72 @@ class DynDNS12UpdateView(generics.GenericAPIView):
 
         # give up
         return SetIPs(ips=[])
+
+    @staticmethod
+    def _get_action_from_param(
+        param_key: str, param_value: str, separator: str
+    ) -> UpdateAction | None:
+        """
+        Parses a single query parameter value to determine the DynDNS update action.
+
+        This function is responsible for interpreting the `param_value` (which can be a single IP,
+        a comma-separated list of IPs, 'preserve', or a CIDR subnet) and converting it into
+        a structured UpdateAction dataclass. It also performs validation on the parameter's format.
+
+        Args:
+            param_key: The name of the query parameter (e.g., 'myip', 'myipv4', 'myipv6', or a qname for extra actions).
+                       Used for error messages.
+            param_value: The string value of the query parameter (e.g., '1.2.3.4', '1.2.3.4,5.6.7.8',
+                         '192.168.1.0/24', 'preserve', or '').
+            separator: The character used to distinguish IP versions (e.g., '.' for IPv4, ':' for IPv6).
+
+        Returns:
+            An instance of SetIPs, UpdateWithSubnet, PreserveIPs, or None if no valid action can be
+            derived from the parameter (e.g., an IPv4 address was given, but IPv6 is required by the separator).
+            Returns SetIPs(ips=[]) if param_value is an empty string.
+
+        Raises:
+            ValidationError: If param_value is invalid (e.g., 'preserve' with addresses)
+        """
+        params = set(
+            filter(
+                lambda param: separator in param or param in ("", "preserve"),
+                map(str.strip, param_value.split(",")),
+            )
+        )
+        if not params:
+            return None
+
+        try:
+            (param,) = params  # unpacks if params has exactly one element
+        except ValueError:  # more than one element
+            if params & {"", "preserve"}:
+                raise ValidationError(
+                    detail=f'IP parameter "{param_key}" cannot have addresses and "preserve" or an empty value at the same time.',
+                    code="inconsistent-parameter",
+                )
+            if any("/" in param for param in params):
+                raise ValidationError(
+                    detail=f'IP parameter "{param_key}" cannot use subnet notation with multiple addresses.',
+                    code="multiple-subnet",
+                )
+        else:  # one element
+            match param:
+                case "":
+                    return SetIPs(ips=[])
+                case "preserve":
+                    return PreserveIPs()
+                case str(x) if "/" in x:
+                    try:
+                        subnet = ip_network(param, strict=False)
+                        return UpdateWithSubnet(subnet=subnet)
+                    except ValueError as e:
+                        raise ValidationError(
+                            detail=f'IP parameter "{param_key}" is an invalid subnet: {e}',
+                            code="invalid-subnet",
+                        )
+
+        return SetIPs(ips=list(params))
 
     @staticmethod
     def _sanitize_qnames(qnames_str) -> set[str]:
@@ -189,10 +220,37 @@ class DynDNS12UpdateView(generics.GenericAPIView):
             )
 
     @cached_property
+    def extra_qname_params(self) -> dict[str, dict[str, str]]:
+        """
+        Parses query parameters of the form 'myipv4:qname' or 'myipv6:qname'
+        to extract additional qnames and their associated update arguments.
+
+        Returns:
+            A dictionary where keys are qnames (e.g., 'sub.example.com') and values
+            are dictionaries mapping RR type ('A' or 'AAAA') to the raw query parameter
+            value (e.g., {'A': '1.2.3.4,5.6.7.8'} or {'AAAA': 'preserve'}).
+        """
+        qnames = defaultdict(dict)
+        param_prefix_types = {"myipv4": "A", "myipv6": "AAAA"}
+
+        for param, value in self.request.query_params.items():
+            try:
+                param_prefix, param_suffix = param.split(":", 1)
+            except ValueError:
+                continue
+            type_ = param_prefix_types[param_prefix]
+
+            for qname in self._sanitize_qnames(param_suffix):
+                qnames[qname][type_] = value
+
+        return qnames
+
+    @cached_property
     def domain(self) -> Domain:
+        qnames = self.qnames | self.extra_qname_params.keys()
         qname_qs = (
             Domain.objects.filter_qname(qname, owner=self.request.user)
-            for qname in self.qnames
+            for qname in qnames
         )
         domains = (
             Domain.objects.none()
@@ -200,7 +258,7 @@ class DynDNS12UpdateView(generics.GenericAPIView):
             .all()
         )
 
-        if len(domains) != len(self.qnames):
+        if len(domains) != len(qnames):  # Some qname doesn't map to a domain
             metrics.get("desecapi_dynDNS12_domain_not_found").inc()
             raise NotFound("nohost")
 
@@ -218,6 +276,26 @@ class DynDNS12UpdateView(generics.GenericAPIView):
     def subnames(self) -> list[str]:
         return [qname.rpartition(f".{self.domain.name}")[0] for qname in self.qnames]
 
+    @cached_property
+    def extra_actions(self) -> dict[tuple[str, str], UpdateAction]:
+        """
+        Converts the raw string arguments from `extra_qname_params` into structured `UpdateAction` objects.
+
+        Returns:
+            A dictionary where keys are `(type, subname)` tuples (e.g., ('A', 'sub'))
+            and values are `UpdateAction` instances (SetIPs, UpdateWithSubnet, PreserveIPs).
+        """
+        return {
+            (
+                type_,
+                qname.rpartition(f".{self.domain.name}")[0],
+            ): self._get_action_from_param(
+                qname, param_value, "." if type_ == "A" else ":"
+            )
+            for qname, param_values_by_type in self.extra_qname_params.items()
+            for type_, param_value in param_values_by_type.items()
+        }
+
     def get_serializer_context(self):
         return {
             **super().get_serializer_context(),
@@ -226,8 +304,12 @@ class DynDNS12UpdateView(generics.GenericAPIView):
         }
 
     def get_queryset(self):
+        subnames = [
+            *self.subnames,
+            *[subname for (_, subname) in self.extra_actions.keys()],
+        ]
         return self.domain.rrset_set.filter(
-            subname__in=self.subnames, type__in=["A", "AAAA"]
+            subname__in=subnames, type__in=["A", "AAAA"]
         ).prefetch_related("records")
 
     @staticmethod
@@ -261,6 +343,12 @@ class DynDNS12UpdateView(generics.GenericAPIView):
             "A": self._find_action(["myip", "myipv4", "ip"], separator="."),
             "AAAA": self._find_action(["myipv6", "ipv6", "myip", "ip"], separator=":"),
         }
+        subname_actions = {
+            (type_, subname): action
+            for subname in self.subnames
+            for type_, action in actions.items()
+        }
+        subname_actions.update(self.extra_actions)
 
         data = [
             {
@@ -269,8 +357,7 @@ class DynDNS12UpdateView(generics.GenericAPIView):
                 "ttl": 60,
                 "records": records,
             }
-            for subname in self.subnames
-            for type_, action in actions.items()
+            for (type_, subname), action in subname_actions.items()
             if (records := self._get_records(subname_records[subname], action))
             is not None
         ]
--- a/api/desecapi/tests/test_dyndns12update.py
+++ b/api/desecapi/tests/test_dyndns12update.py
@@ -330,6 +330,101 @@ class DynDNS12UpdateTest(DynDomainOwnerTestCase):
         self.assertIP(ipv4=new_ip)
         self.assertIP(subname="sub", ipv4=new_ip)
 
+    def test_update_multiple_with_overwrite(self):
+        # /nic/update?hostname=sub1.a.io,sub2.a.io,sub3.a.io&myip=1.2.3.4&ipv6=::1&myipv6:sub2.a.io=::2
+        new_ip4 = "1.2.3.4"
+        new_ip6 = "::1"
+        new_ip6_overwrite = "::2"
+        domain1 = "sub1." + self.my_domain.name
+        domain2 = "sub2." + self.my_domain.name
+        domain3 = "sub3." + self.my_domain.name
+
+        with self.assertRequests(
+            self.request_pdns_zone_update(self.my_domain.name),
+            self.request_pdns_zone_axfr(self.my_domain.name),
+        ):
+            response = self.client.get(
+                self.reverse("v1:dyndns12update"),
+                {
+                    "hostname": f"{domain1},{domain2},{domain3}",
+                    "myip": new_ip4,
+                    "ipv6": new_ip6,
+                    f"myipv6:{domain2}": new_ip6_overwrite,
+                },
+            )
+
+        self.assertStatus(response, status.HTTP_200_OK)
+        self.assertEqual(response.data, "good")
+
+        self.assertIP(subname="sub1", ipv4=new_ip4, ipv6=new_ip6)
+        self.assertIP(subname="sub2", ipv4=new_ip4, ipv6=new_ip6_overwrite)
+        self.assertIP(subname="sub3", ipv4=new_ip4, ipv6=new_ip6)
+
+    def test_update_multiple_with_extra(self):
+        # /nic/update?hostname=sub1.a.io,sub3.a.io&myip=1.2.3.4&ipv6=::1&myipv6:sub2.a.io=::2
+        old_ip4 = "10.0.0.2"
+        new_ip4 = "1.2.3.4"
+        new_ip6 = "::1"
+        new_ip6_extra = "::2"
+        domain1 = "sub1." + self.my_domain.name
+        domain2 = "sub2." + self.my_domain.name
+        domain3 = "sub3." + self.my_domain.name
+        self.create_rr_set(self.my_domain, [old_ip4], subname="sub2", type="A", ttl=60)
+
+        with self.assertRequests(
+            self.request_pdns_zone_update(self.my_domain.name),
+            self.request_pdns_zone_axfr(self.my_domain.name),
+        ):
+            response = self.client.get(
+                self.reverse("v1:dyndns12update"),
+                {
+                    "hostname": f"{domain1},{domain3}",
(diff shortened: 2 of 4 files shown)


peterthomassen:
Do we need a discriminator suffix at all? (The address family is evident from the value .)

fiwswe:
@peterthomassen Thinking about it, I agree that the suffix is unnecessary. That would also help to keep down the length of the URL.

Even the scenario where one would want to set both IPv4 and IPv6 addresses for a hostname independently could be handled like this:
`<domain>=<ipv4>,<ipv6>` or `<domain>=<ipv6>,<ipv4>`
Where `<ipv4>` and `<ipv6>` could be actual IPs, with or without a subnet specifier, or keywords like `preserve` or `none`.

Extrapolating this syntax further, I don't see any practical value in allowing more than one IP value for each address family. But I could be missing some use case?

fiwswe:
@molikuner Thanks for your work!

Any API where the same parameter can be used multiple times (possibly relying on the order of parameters) is probably not a good idea. That eliminates your section 1.

The syntax `<domain>=<ip1>,<ip2>` should be fairly easy to parse. The list of IPs should have some constraints:
  * Minimum of one list item (IP).
  * Maximum of 2 list items (IPs).
  * Only 0-1 of each address family.
  * If keywords (`preserve`, `none`) are used, either the specified IP defines which address family the keyword applies to or in the case of `preserve,preserve` (this combination could be flagged as an error because there is no use-case, I can see) or `none,none` it does not matter. So `<ipv4>,preserve` equals `preserve,<ipv4>` and `<ipv6>,preserve` equals `preserve,<ipv6>`. Same for `none` respectively. Only `preserve,none` would actually be ambiguous. But is there a use-case for that? This is a static setting, no need for dynamic updates. In this case the order could define which is which, if it is not flagged as an error. So a constraint could be specified that at least one list item must be an IP. That solves all of these issues.
  * Empty list elements are not allowed (no `,<ip>` or `<ip>,`). The keyword `none` (or maybe the shorter `no`) designates that the respective IP should be deleted, avoiding the problem you described with empty list elements.

As for having different rules for `<domain>=` and `ip=`, either ignore he slight inconsistency or adjust `ip=` to work the same.

Having deSEC manage a potential future TLD `ip.` and creating an ambiguity is more a theoretical concern, I'd say. I'd file this under _very unlikely_.

My Python skills are not sufficient to help with coding, sorry. But the design of the API should not be influenced by what happens to be convenient in code I think. Keeping URLs short and having a syntax that is able to express all use cases is more important IMHO.

fiwswe

molikuner:
Thanks @fiwswe for your response. There are a few points that I don't necessarily agree with, or at least want to explain some of my thoughts:

> Any API where the same parameter can be used multiple times (possibly relying on the order of parameters) is probably not a good idea.

This is already how the existing query parameters work. If you specify them multiple times, only the last one will be considered. I don't want to say that I disagree with you, but we have to live with the existing API (and must not break it, as existing clients might rely on this behavior).

Please also note, that I wasn't suggesting to rely on the order. Just that the domain could be specified twice one time for the IPv4 address and one time for the IPv6 address. But I see how there is little benefit over the other solution with comma separated values.

> Minimum of one list item (IP).

IMHO a request like this should be allowed: `?hostname=one.a.io,two.a.io&ip=1.2.3.4&two.a.io=`. I would interpret it as "set IPv4 `1.2.3.4` for `one.a.io`, infer the IPv6 for `one.a.io` and remove both IPv4 and IPv6 for `two.a.io`. Or do you see a specific reason this shouldn't be allowed? Sure, maybe it doesn't make sense to have this in a static script, run daily to renew a dynamic IP address, but maybe someone has a use-case where the IPv4 address and IPv6 address should be removed every now and then.

> Maximum of 2 list items (IPs).
> Only 0-1 of each address family.

We can already do requests like this: `?ip=1.2.3.4,5.6.7.8,9.10.11.12`. It sets three A records for the updated domain. Why should we prevent this in the new domain based parameters?

> Only `preserve,none` would actually be ambiguous.

That's actually a good point. With my current proposal this would be possible: `?a.io.myipv4=&a.io.myipv6=preserve`. The result would be that the IPv4 address would be preserved while the IPv4 address is deleted. It wouldn't be really possible to represent that with a `?a.io=,preserve` since it would be impossible to know what applies where.

> Empty list elements are not allowed (no `,<ip>` or `<ip>,`). The keyword `none` (or maybe the shorter `no`) designates that the respective IP should be deleted, avoiding the problem you described with empty list elements.

Why create yet another keyword? Your proposal of `none` behaves exactly the same way as an empty string. I would personally not like to have even more differences between this domain parameter and the existing parameters. My point wasn't that the empty value is a problem, the point is just that we have to consider how the new parameter compares to the old ones and whether they should be adjusted as well.

> either ignore he slight inconsistency or adjust ip= to work the same.

From my experience, ignoring inconsistencies increases support burden, as people don't understand these slight differences. Keeping things simple is usually the better approach. So even if it might seem negligible I would prefer consistency. Adjusting the way the existing parameters are handled is also weird since we have the parameters that specify the IP version in their name already.

---

I also want to highlight another point: "scriptability". This API is intended to be used by some script and having both IPv4 and IPv6 address in the same parameter creates certain edge-cases that need to be handled:

Given someone tries to create a simple script that updates `updater.example.com` with the IP address of the execution host and has two inputs, the expected IPv4 and IPv6 address of `server.example.com`:

```bash
curl \
  "https://update.dedyn.io/?hostname=updater.example.com,server.example.com&server.example.com=$IPV4,$IPV6" \
  --header ...
```

Now this would fail, if the expected IPv4 address as well as the expected IPv6 address are none, i.e. there shouldn't be any A or AAAA record. It would fail as the value of the `server.example.com` would be just `,`. The edge-case that neither a IPv4 nor a IPv6 address are given would need to be handled explicitly be removing the comma between the empty values.
With separate parameters this problem wouldn't happen, as it's okay to have both parameters as an empty value.

fiwswe:
> This is already how the existing query parameters work. If you specify them multiple times, only the last one will be considered.

Are you referring to URL parameters in general or the deSEC IP Update API specifically? For the general case it would depend on the application interpreting the URL parameters I think. But you are probably correct for the deSEC case.

> Please also note, that I wasn't suggesting to rely on the order.

No you weren't. I had some alternative specifications in  the back of my mind, akin to `hostname=<host1>&ip=1.2.3.4&hostname=<host2>&ip=5.6.7.8` But forget those, as they are confusing and waste URL length.

> We can already do requests like this: ?ip=1.2.3.4,5.6.7.8,9.10.11.12. It sets three A records for the updated domain. Why should we prevent this in the new domain based parameters?

We can? I didn't know that. In that case I withdraw that constraint. OTOH in that case determining to which address family the `preserve` or empty / `none` list items apply gets harder.

> From my experience, ignoring inconsistencies increases support burden, as people don't understand these slight differences. Keeping things simple is usually the better approach. So even if it might seem negligible I would prefer consistency.

ok.

> Adjusting the way the existing parameters are handled is also weird since we have the parameters that specify the IP version in their name already.

But they don't need to change and the `ip` parameter is already documented to hold either IPv4 or IPv6 address(es), which to me is mildly confusing. But I think that has its origins in other DDNS APIs (probably from a time when IPv6 was not in general use)?

So with this adjusted, instead of `…&myip=<ipv4>&ipv6=<ipv6>` this syntax would be equivalent and shorter: `…&ip=<ipv4>,<ipv6>`

Regarding scriptability, I agree that is a valid concern. Maybe that was why I discarded empty list elements and suggested the more explicit `none` / `no`. However with arbitrary length lists that doesn't work anymore. Unless the keywords are qualified with the address family they apply to, e.g. `<domain>=6:preserve,1.2.3.4,5.6.7.8`, `<domain>=2001:db8::/56,4:no` or something similar.

It would need to be specified what happens in cases like: `<domain>=4:preserve,1.2.3.4,5.6.7.8`? Would the keyword override the explicit IPs for the same address family? Or the other way around?

nils-wisiol:
Thanks for this PR and the interesting discussion!

deSEC's API already supports [bulk operations](https://desec.readthedocs.io/en/latest/dns/rrsets.html#bulk-operations). Is there a kind of change that is supported in this PR but not through bulk operations?

fiwswe:
@nils-wisiol No, I don't think anything we are trying to change in this PR is a feature not supported by the general API. (Let's turn that statement around: I think the general API supports everything we need for this PR. No issues there.)

Apparently the IP Update API also allows specifying multiple IPs for an address family. I was not aware of that fact before and I must have missed it in the documentation as well. But it is in fact [documented](https://github.com/desec-io/desec-stack/blame/7f753a9bdc43299cff15a8e9f019b01d2ee82ac5/docs/dyndns/update-api.rst#L122). So that was my mistake. Sorry for the confusion.

molikuner:
Hey @nils-wisiol,

The [bulk operations API](https://desec.readthedocs.io/en/latest/dns/rrsets.html#bulk-operations) indeed supports the updates that we're talking here about already (and even more as it seems). I would love to use it, but unfortunately some of my devices aren't too flexible on the request being send. Thus also supporting bulk operations via the "[dyndns API](https://desec.readthedocs.io/en/latest/dyndns/update-api.html#update-request)" would be appreciated as I can influence the requests parameters a bit on these devices.

In my previous PR (#1112) the support to update more than one subdomains was already introduced. But so far it's not possible to set different IP addresses for specific domains that are updated. E.g. it's not possible to set one subdomain to `1.2.3.4` and another subdomain to `5.6.7.8` in the same request (though it's already possible to update the network of all subdomains by sending e.g. `10.0.0.0/24` as the IP address which would preserve the host part of the existing IP address).

This current PR would improve on the previous PR by allowing to specify different IP addresses for the updated subdomains in the "dyndns API".

fiwswe:
Oops sorry, I just noticed (after sending of course) that my alternatives do not handle removing the `A` RRset for `ipv6.example.com`. They keep it unchanged (if it exists). Correcting that would indeed require repetition of the prefix.

renne:
Hi,

I lost track what is implemented in #1112 and what is open for this PR.
My application is a Fritz!Box with port-forwarding at a full dual-stack internet socket of Deutsche Telekom:

| Subdomain | RR-Type | Fritz!OS variable |
| --- | --- | --- |
| example.com | A | \<ipaddr> |
| *.example.com | A | \<ipaddr> |
| fritz.example.com | AAAA | \<ip6addr> |
| *.example.com | AAAA | \<ip6lanprefix> |
| a.example.com | AAAA | \<ip6lanprefix> |
| b.example.com | AAAA | \<ip6lanprefix> |
| c.example.com | AAAA | \<ip6lanprefix> |

Do I understand correctly it is possible since the merge of #1112 to use multiple subdomains in a comma-separated list but not to overwrite IPv6 addresses with the LAN prefix?

molikuner:
I just tested with an old Fritz!Box that I had around and noticed some weird behavior in the dyndns client of it. For some reason it doesn't support using the `<ipv6lanprefix>` and `<ipaddr>` placeholder in the same request.

The update URL was set to (as suggested above): 

```
https://update.dedyn.io/api/update?hostname=<domain>,a.<domain>&myipv4=<ipaddr>&myipv6=<ip6lanprefix>
```

This is the request that I observed:

```
/api/update?hostname=example.com,a.example.com&myipv4=1.2.3.4&myipv6=::
```

As you can see, the `myipv6` parameter was set to `::`, which is definitely wrong...

As a workaround, I noticed one can do two requests by configuring the client to use two update URLs by concatenating them using a space:

```
https://update.dedyn.io/api/update?hostname=<domain>,a.<domain>&myipv4=<ipaddr>&myipv6=preserve https://update.dedyn.io/api/update?hostname=<domain>,a.<domain>&myipv4=preserve&myipv6=<ip6lanprefix>
```

Then the first request will set the IPv4 addresses correctly and the second request will update the IPv6 addresses. I could observe the following requests in this case:

```
/api/update?hostname=example.com,a.example.com&myipv4=1.2.3.4&myipv6=preserve

/api/update?hostname=example.com,a.example.com&myipv4=preserve&myipv6=2001:9e8:3:4::/64
```

Note 1: You can not do more than two requests [due to rate limiting](https://desec.readthedocs.io/en/latest/rate-limits.html) in this way.

Note 2: This was tested on FRITZ!OS version `06.88` which is quite old at this point, but that particular Fritz!Box doesn't receive any updates anymore, so it's the newest version that I can test. The behavior might have changed in a newer versions.

molikuner:
Since I had a rather elaborate setup process faking multiple things to test this, I wasn't 100% sure before, but I also had this problem that the Fritz!Box dyndns client would just not do anything.

You could try changing the value for the "Domainname" in config. Every time I changed the value, it was actually doing something for me. The logs are still empty, but at least in the "Online-Monitor" page I was able to see "IPv4-Status: angemeldet, IPv6-Status: angemeldet" after a few seconds.

I'm using an actual `,` in the "Update-URL". Using the URL you shared worked for me with the caveat that `myipv6=::` was send as explained earlier.

fiwswe:
How about separating the potential FRITZ!OS issues from the actual deSEC update mechanism?

Build the test URL manually and send it via e.g. curl. E.g. use these two URLs alternately (replacing `host1` and `host2` with actual hostnames in the deSEC account of course):
* `https://update.dedyn.io?hostname=host1,host2&myipv4=192.0.2.1&myipv6=2001:db8:1::/56`
* `https://update.dedyn.io?hostname=host1,host2&myipv4=192.51.100.2&myipv6=3fff:0:2::/56`

If FRITZ!OS does not work as advertised then a bug should be reported to FRITZ! GmbH.

I have tried something like the above (using curl) but received HTTP Status 500 whenever I tried to update an IPv6 prefix. IPv6 addresses work fine. See [Problems updating IPv6 prefix](https://talk.desec.io/t/problems-updating-ipv6-prefix/1527) in the deSEC Forum.

I have not built a testing setup for testing the FRITZ!OS DynDNS client yet. It would require a webserver or proxy that logs the request to see what it is actually sending. I might do that though.

fiwswe:
I have built a testing setup for the FRITZ!OS DynDNS client as follows:
* A webserver (hosted outside of my LAN) that always returns HTTP status 200 and `good` for valid queries and which logs the URL and the request headers to a file.

Then I modified the URL in the DynDNS settings in FRITZ!OS on my FRITZ!Box to point to my webserver instead of the DDNS provider I normally use. Nothing else was changed and the URL in the settings looked like this:
`https://_mywebserver_/?hostname=<domain>&ipv4=<ipaddr>&ipv6=<ip6addr>&token=<username>&ipv6prefix=<ip6lanprefix>`

This resulted in correct values for the parameters `hostname`, `ipv4`, `ipv6`, `token` and `username` being logged.

So the FRITZ!OS 8.20 DynDNS client works as advertised for me. All of the placeholders were replaced with the expected values.

@molikuner I don't know what the cause of your issue was. But if you used the `<ip6lanprefix>` placeholder with deSEC, you may have run into the Issue I have asked about in [Problems updating IPv6 prefix](https://talk.desec.io/t/problems-updating-ipv6-prefix/1527) and mentioned previously?

molikuner:
Hey @peterthomassen,

Thanks for taking the time to discuss this and coming back with a proposal. In general I do think having a prefix instead of a suffix could work too. My week was a bit busy and I couldn't reply earlier. 

---

> 4. To reduce confusion and increase symmetry, it may be a good idea to not consider ipv6 for this, as there is no ipv4 parameter.

> We cannot add an ipv4 parameter for symmetry, as that would break the documented behavior (which is that after trying myip, myipv4, ip, the IPv4 address will be taken from the connection).

While it's not possible to introduce `ipv4` as a "standalone" parameter due to the documented behaviour, the new prefixes could technically support it. I see how it could create even more confusion though, so not supporting `ipv4` and `ipv6` makes sense to me. 👍

---

> I noticed that your code proposal currently determines relevant subdomains from self.subnames. Thus, a subdomain needs to be specified both in a parameter that populates this attribute, and also in the new qualified parameter. It seems to me that the former is unnecessary, so the following should work:
> 
> `/?hostname=example.dedyn.io&myip=1.2.3.4&myipv6:router.example.dedyn.io=preserve`
> (without specifying hostname=example.dedyn.io,router.example.dedyn.io)
> 
> Am I missing something?

I guess it could work too, but to me it would not be obvious, what is going to happen, just looking at the request. Taking the following examples:

| request params                                                               | username  | a.m.io IPv4 | a.m.io IPv6        | b.m.io IPv4 | b.m.io IPv6 |
|:------------------------------------------------------|-----------|-------------|-----------------|-------------|------------|
| `hostname=a.m.io&myip=1.2.3.4&myipv6:b.m.io=1::2` | N/A           | `1.2.3.4`      | Client IP / unset | `1.2.3.4` ***1** | `1::2`          |
|`myip=1.2.3.4&myipv6:b.m.io=1::2`                                 | `a.m.io`   | `1.2.3.4` ***2**  | Client IP / unset ***2** | `1.2.3.4` ***1** | `1::2`     |
   
- *1 Or should this rather be the Client IP, unset or preserved? The high-level question is: If there is a prefix-parameter for one address family, but not for the other, what happens? With my current solution is rather simple to say, there is no specific overwrite, so it would follow the "normal" behaviour.
- *2 Or should this rather be preserved? The high-level question is: If a hostname was found in the (new) query parameters, should the username (or the other fallback, the single domain of the user) still apply? I think updating both domains makes sense, but could make it harder to see which domains are affected by a single request.

I can find arguments for all possible behaviours and while some might make more sense than others, I wouldn't know what to implement without further feedback.

My initial idea was to only update the domains that were explicitly specified in the already known places and just adding the new feature of "overwrites". This would leave the existing logic of deciding which domains to update as is and potentially cause less confusion. I see how it creates duplications in the parameters, but I think it's clearer what's going to happen.

---

> Before implementation, would you be able to try out in your lab setup whether requests as specified in https://github.com/desec-io/desec-stack/pull/1141#issuecomment-3576230099 would go through properly? (Just to make sure Fritz!Box doesn't get confused by : etc.)

Sure, I'll try it out. It might take a few days to find the time to do so, but I'll come back with some results 👍.

Hoeze:
Would it not make more sense to involve AVM/FRITZ! here so they fix their OS?

peterthomassen:
> I guess it could work too, but to me it would not be obvious, what is going to happen, just looking at the request. Taking the following examples:

I think these examples should do whatever they would do without the new parameter, plus only the action specified by the new parameter explicitly. So,

- `hostname=a.m.io&myip=1.2.3.4&myipv6:b.m.io=1::2` should do whatever `hostname=a.m.io&myip=1.2.3.4` does so far, and in addition set `b.m.io` to `1::2`
- `myip=1.2.3.4&myipv6:b.m.io=1::2` should do whatever `myip=1.2.3.4`, and in addition set `myipv6:b.m.io` to `1::2`

It would lead to the follow results:

| request params                                                               | username  | a.m.io IPv4 | a.m.io IPv6        | b.m.io IPv4 | b.m.io IPv6 |
|:------------------------------------------------------|-----------|-------------|-----------------|-------------|------------|
| `hostname=a.m.io&myip=1.2.3.4&myipv6:b.m.io=1::2` | N/A           | `1.2.3.4`      | Client IP / unset | unchanged | `1::2`          |
|`myip=1.2.3.4&myipv6:b.m.io=1::2`                                 | `a.m.io`   | `1.2.3.4`  | Client IP / unset | unchanged | `1::2`     |

`myipv4:b.m.io=` and `myipv6:b.m.io=` (empty values) can be allowed to remove a value, but `preserve` I don't think needs to be supported (because it's the same as absence of the parameter.

I think this is also easily described by adding a section to the documentation that deals with these additional parameters.

Do you see any issues with this solution?

molikuner:
> I think these examples should do whatever they would do without the new parameter, plus only the action specified by the new parameter explicitly. 
>
> ...
>
> I think this is also easily described by adding a section to the documentation that deals with these additional parameters.
>
> Do you see any issues with this solution?

That sounds reasonable. It's not how I had initially thought about it, but works as well 👍.

> > Before implementation, would you be able to try out in your lab setup whether requests as specified in https://github.com/desec-io/desec-stack/pull/1141#issuecomment-3576230099 would go through properly? (Just to make sure Fritz!Box doesn't get confused by : etc.)
>
> Sure, I'll try it out. It might take a few days to find the time to do so, but I'll come back with some results 👍.

Sorry for taking so much time. I was finally able to test out the proposed syntax. Here are my results (on the ancient Fritz!OS version `06.88`):

| Option | Value |
|-------------|-----------|
| Update URL | `http://192.168.203.35:3010/api/update?hostname=<domain>,device.<domain>&myipv4=<ipaddr>&myipv6=<ip6lanprefix>&myipv6:<domain>=<ip6addr>` |
| Domain name | `example.com` |
| User name | `example` |
| Password | `com` |


This is the request that I observed:

```
GET /api/update?hostname=example.com,device.example.com&myipv4=46.0.0.1&myipv6=2001::/64&myipv6:example.com=2001::c984 HTTP/1.1
Host: 192.168.203.35
Authorization: Basic ZXhhbXBsZTpjb20=
User-Agent: Fritz!Box DDNS/1.0.1
Connection: close
```

To test out some more special characters, I've also tried the following update url: `http://192.168.203.35:3010/api/update?hostname=<domain>,device.<domain>&myipv4=<ipaddr>&myipv6=<ip6lanprefix>&myipv6:<domain>=<ip6addr>&myipv4:_.<domain>=<ipaddr>&myipv6:_.<domain>=<ip6lanprefix>`

Using this URL I've observed the following request:

```
GET /api/update?hostname=example.com,device.example.com&myipv4=46.0.0.1&myipv6=2001::/64&myipv6:example.com=2001::c984&myipv4:_.example.com=46.0.0.1&myipv6:_.example.com=2001::/64 HTTP/1.1
Host: 192.168.203.35
Authorization: Basic ZXhhbXBsZTpjb20=
User-Agent: Fritz!Box DDNS/1.0.1
Connection: close
```

Note: I've redacted/changed the IP addresses, but they were received correctly by my server.

Overall I couldn't see anything wrong with the way Fritz!OS is behaving here, so I'll look into adjusting the code to match this new style soon 👍 .

api/desecapi/views/dyndns.py
```diff
@@ -132,6 +97,71 @@ def _find_action(self, param_keys, separator) -> UpdateAction:
         # give up
         return SetIPs(ips=[])
 
+    @staticmethod
+    def _get_action_from_param(param_key: str, param_value: str, separator: str) -> UpdateAction | None:
```
molikuner: This is my attempt at re-using the existing logic to parse the parameter (note that I just copied it from the method above). Unfortunately this leads to two questionable "features", that I would like to have some feedback on, whether it's fine to keep it or should be explicitly forbidden:

1. The new parameter supports `perserve`
2. The new parameter gets ignored if the wrong IP address version was given, e.g. if `myipv6:a.example=1.2.3.4` is given

IMHO it's fine to support the first "feature", as it doesn't harm, but could even enable some interesting API calls, e.g.

```
https://update.dedyn.io/?hostname=example.com,sub.example.com&myipv4=<ipaddr>&myipv6=<ip6addr>&myipv6:sub.example.com=preserve
```

As for the second one, I'm a bit unsure. At first I thought it's fine, as it mirrors the functionality of the existing parameters, e.g. `myipv6=1.2.3.4` will be ignored as well. But on the contrary I'm unsure whether this could cause too much confusion and since it's a new parameter we can be as strict as we want.

api/desecapi/views/dyndns.py
```diff
@@ -188,19 +218,52 @@ def qnames(self) -> set[str]:
...
+
+        for param, value in self.request.query_params.items():
+            if param.startswith("myipv6:"):
+                type_ = "AAAA"
+            elif param.startswith("myipv4:"):
+                type_ = "A"
+            else:
+                continue
+
+            for qname in self._sanitize_qnames(param.split(":", 1)[1]):
```
molikuner: Note for the reviewer: Since I'm reusing the existing `_sanitize_qnames` function, this means the new parameters will also support multiple domains, e.g. `myipv6:sub1.example.com,sub2.example.com=2001::/64`.

peterthomassen: That's fine.

The `existing` logic looks like you're also somehow combining values from multiple parameters whose list of qnames has overlap, such as: `&myipv4:a.example,b.example=1.2.3.4&myipv4:b.example=4.3.2.1`. If I read your code correctly, we then have `qname['b.example']['A'] = '1.2.3.4,4.3.2.1'`. Is that correct?

(I find that rather unexpected, and would think that the last value takes precedence.)

peterthomassen (commented):
Thank you for this! The overall approach looks fine, but I have some minor things I'd touch up a little. I can do this (pending your input on my in-line question.)

peterthomassen:
The current state is that we've reviewed the code and are waiting for follow-up.

api/desecapi/views/dyndns.py
molikuner: Yes, you understood the code correctly. I think your concern makes sense. I should have aligned the behavior with the existing parameters. I'll adjust the code to only consider the last parameter per domain. 👍 

PS: Sorry for the late reply. Somehow the email must have gone missing and I didn't see your reply earlier. Just due to the other replies in recent days I got to be aware, that you're waiting for a response from me.

Markus-Go:
Thank you so much for your efford!
I hope the other bug gets fixed soon, probably many people like me looking forward that this PR gets merged, being able to use the Fritzbox as dyndns client again after so many months!

api/desecapi/views/dyndns.py
peterthomassen: I think that is fine!

peterthomassen (approved):
Thanks everyone for this work! Will deploy in a few minutes.

PR #1160, opened 2025-10-18
fix(webapp): Fix typo in DomainSetup.vue

perzarys:
Just a little typo I noticed, would be great to see it corrected.


nils-wisiol:
Thanks for catching this. This is good to merge after the commit message has been adjusted to begin with "fix(webapp):".

peterthomassen (approved):
thanks!

peterthomassen:
Thanks for tracking this. We just haven't done a deployment yet, but this was the right place. Cheers!

Issue #1192, opened 2026-02-26
webapp: Allow "open in new tab" (have domains on /domains be <a> links)


AbelLykens:
Hi! 

The `<tr>`s on desec.io/domains currently get a click event, which sends users to /domains/example.com.

Can the domains on the list be <a> links, so open-in-new-tab works?

Currently, because the rows use JavaScript click handlers instead of native <a> elements, users lose the ability to interact with domain links the way they would with any standard hyperlink. This means right-clicking to "Open in New Tab," middle-clicking, or Ctrl/Cmd-clicking all fail to work as expected. For users managing many domains, this is a significant workflow friction, as they often want to open several domains in separate tabs to configure them in parallel without losing their place in the list.

Beyond usability, using semantic <a> links also improves accessibility and aligns with web standards. Screen readers and other assistive technologies can properly identify and announce links, giving users a clear indication that the element is navigable. It also means the browser can show the destination URL on hover in the status bar, giving users confidence about where they're going before they click. Overall, replacing the click handlers with proper anchor elements is a small change that would bring the UI in line with expected browser behavior and make the experience smoother for everyone.

Thanks!

PR #1193, opened 2026-02-26
Try <a>'ing links

AbelLykens:
Hope this works, for #1192. 

  CrudList.vue — The column template now checks for a column.link function. If present, the column content is wrapped in a <router-link> (renders as <a>). The @click.native.stop prevents the click from bubbling up to the row's @click:row handler, so regular clicks navigate via the link, and ctrl-click / middle-click / right-click all work natively. Columns without link render exactly as before.               
                                                                                                                                                                                                                                                                                                                                                                                                                                                    CrudListDomain.vue — Added link: (item) => ({name: 'domain', params: {domain: item.name}}) to the name column. This generates the /domains/example.com URL as a proper <a> href.             
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 The existing handleRowClick is kept, so clicking anywhere else on the row (e.g. the "Published" column) still navigates as before. But now the domain name is a real link with full browser link semantics: right-click menu, open in new tab, ctrl/cmd-click, middle-click, and URL preview in the status bar.

### DIFF
--- a/www/webapp/src/views/CrudList.vue
+++ b/www/webapp/src/views/CrudList.vue
@@ -197,8 +197,15 @@
               v-for="(column, id) in columns"
               #[column.name]="itemFieldProps"
             >
+              <router-link
+                v-if="cellUsesPlainText(column, rawItem(itemFieldProps.item)) && column.link"
+                :key="`${id}-link`"
+                class="crud-readonly-value crud-link"
+                :to="column.link(rawItem(itemFieldProps.item))"
+                @click.stop
+              >{{ readonlyDisplayValue(column, rawItem(itemFieldProps.item)) }}</router-link>
               <div
-                v-if="cellUsesPlainText(column, rawItem(itemFieldProps.item))"
+                v-else-if="cellUsesPlainText(column, rawItem(itemFieldProps.item))"
                 :key="`${id}-readonly`"
                 class="crud-readonly-value"
                 v-text="readonlyDisplayValue(column, rawItem(itemFieldProps.item))"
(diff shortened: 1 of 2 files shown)


peterthomassen:
Thank you! I've reworked this a bit as the Vue 3 upgrade which happened in the meantime required tackling this a bit differently (but kept your approach).

peterthomassen:
Hmm, that didn't work as I had hoped.  Opening a domain link in a new tab lands on the login page because the auth token isn't shared with the new tab.

I thought the "Remember me during this browser session" checkbox at login controls this, but turns out that's only about reloading in the same tab.

So, how to get this working? Options:
  1. localStorage — shared by all tabs immediately, no handshake. Also survives browser restarts, which the current design avoids; server-side limits (7 d max age, 1 h unused) bound the exposure. The checkbox would become "this tab" vs "all tabs". I'm not convinced keeping DNS credentials alive is a good idea.
  2. BroadcastChannel / SharedWorker handshake — a new tab asks open tabs for the token; nothing is persisted, so the datensparsam default is kept. Brief unauthenticated moment (not sure how that looks), also seems engineering-heavy and lastly I'm not sure other tabs could hijack / exfiltrate that.
  3. HttpOnly session cookie from the API — covers every tab and removes the token from JavaScript entirely. Requires CSRF, also we'd loose the unusual trait of not needing cookies.
  4. Keep the links (for status bar display etc.), but force opening through `window.open()` where session storage is inherited. This seems best to me, but I don't know if there are use cases that would interfere with. There may be an issue with pop-up blockers, although I don't think at the first time. (Also, it would probably work with <div> instead of <a> as well.)
  5. I think `target="_blank"` also shares session context if the link target is in the same origin. We could try to tweak things such that ctrl+click would somehow cause the `target="_blank"` behavior, whereas a normal click would suppress it and proceed in the same tab. No idea about interference with popup blockers, non-visual interface etc.

@AbelLykens You seemed to find it important that there are actual <a> links. Is there a specific reason for this implementation choice, or would you be fine with any implementation as long as the functionality is provided?

Also, @nils-wisiol.

PR #1211, opened 2026-07-20
Vuetify 4

Ph0tonic:
Hey,
I tried to vuetify 4 but I also needed to upgrade:
- vue-router 5
- pinia 4

I removed vue-demi which did not seemed to be used.

I also update the qrcode-vue inlined repo to a vue3 version, allowing to simplify it's vue integration.


peterthomassen:
@Ph0tonic This PR was accidentally closed as I was merging the other which was the basis of this. As the basis is now gone, this one was closed.

Vuetify 3 is now in main. We're still interested in a Vuetify 4 update! So, if you find the time, we'd appreciate if you want to take another stab at it.

(But I had noticed that the migration to 3 added about 1k lines of code, whereas this one was a whole lot more. I didn't look in detail, but I was wondering why that was?)

Ph0tonic:
Hey @peterthomassen,

I created a new one with a rebase #1214 .
Regarding the whole lot more code, I upgraded the inlined dependency qrcode component to the latest version which is directly vue3 compatible. The idea being to avoid having custom vue v2 to v3 wrapper.

Let me know what you think of all this.

PR #1212, opened 2026-07-24
fix(nsmaster): allow larger request body

peterthomassen:
As our number of domains grew, this is now needed to provision the catalog zone, which is done when API container (re)starts. We currently have a hotfix in production.

### DIFF
--- a/api/api/settings.py
+++ b/api/api/settings.py
@@ -191,8 +191,7 @@ TASK_CONFIG = {  # The first entry is the default queue
 }
 
 # pdns accepts request payloads of this size.
-# This will hopefully soon be configurable: https://github.com/PowerDNS/pdns/pull/7550
-PDNS_MAX_BODY_SIZE = 16 * 1024 * 1024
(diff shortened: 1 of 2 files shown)


api/api/settings.py
```diff
@@ -192,7 +192,7 @@
 
 # pdns accepts request payloads of this size.
 # This will hopefully soon be configurable: https://github.com/PowerDNS/pdns/pull/7550
-PDNS_MAX_BODY_SIZE = 16 * 1024 * 1024
+PDNS_MAX_BODY_SIZE = 32 * 1024 * 1024
```
nils-wisiol, line 194: Comment above seems stale

nils-wisiol (approved):

api/api/settings.py
peterthomassen: done (removed)

peterthomassen:
merged in 1dd187ccdacd1f9d87b374a772166d957ddc01a4 (via #1173)

PR #1214, opened 2026-07-28
chore(webapp): migration to vuetify 4

Ph0tonic:
Hey,
I tried to upgrade to vuetify 4 but I also needed to upgrade:

- vue-router 5
- pinia 4

I removed vue-demi which did not seemed to be used.

I also update the qrcode-vue inlined repo to a vue3 version, allowing to simplify it's vue integration.

Replaces #1211


peterthomassen:
Thanks for doing this, this looks pretty good!

I didn't find any functionality that broke, but there are several changes in how things look. Some of that is fine (for example, whether the sign-up button is uppercase or lowercase), but others are more significant for flow / smooth perception:

- [ ] In a large browser window, the padding to the left/right of the main body text is gone
- [ ] Sometimes there is weird spacing, such as between the checkboxes in the sign-up form
- [ ] Sometimes, fonts have wrong size. Example: in the domain list, click on the (i) to see configuration details. The "copy" buttons have significantly oversized labels

If you have some time, would you mind "playing through" as much GUI functionality as you can, and note down if you spot anything odd? There's no need for immediate fixes; the list will suffice for a start. We can then see what's easy to fix / significant or not / where assisted coding can help etc.

:pray:

Ph0tonic:
Hey @peterthomassen,

I played with the UI and fixed everything I could find.

I also made a change to the switches: I noticed it was hard to tell whether they were on or off, so I gave them a color (Vuetify 4 dropped the default accent color on switches, so they render grey either way now). It makes the state much easier to read at a glance.

Let me know what you think!

Ph0tonic:
Hey,
I just rebased this PR, let me know if you are still interested and I can do anything.

www/webapp/package.json
```diff
@@ -46,4 +46,4 @@
     "vite-plugin-vuetify": "^2.1.3",
     "vitest": "^3.2.4"
   }
-}
\ No newline at end of file
+}
```
peterthomassen, line 49: If this is all this commit does, then it's not doing what its title/message says

peterthomassen (changes requested):
thanks! one thing stood out.

www/webapp/package.json
Ph0tonic: Hello,

Are you referring to the addition of `@vue/devtools-api`, or the extra blank line?

For `@vue/devtools-api`: it's required because Pinia now moves it to a mandatory peer dependency:
- https://github.com/vuejs/pinia/commit/3cd16de69be3771a7d817abe513e978eff8e2248

For the extra line: fixed

Issue #1219, opened 2026-08-08
Security: chores for website


Rotzbua:
Tasks:

- [x] GH action update Nodejs to supported
- [x] use fixed nodejs in docker (so it does not breaks if lts changes version)
- [x] add cooldown for dependencies

PR #1220, opened 2026-08-08
feat(webapp): Use fixed Nodejs version v24

Rotzbua:
Node 24 is current supported version.

Lts version in docker build can break. Because Lts changes over time.

## Reference

mentioned in #1219


peterthomassen (approved):
lgtm, thanks!

PR #1224, opened 2026-08-08
fix(test): bump python to 3.13

### LINKED ISSUE #1223: Bug: test e2e2 and api do not use same python version
API uses 3.13 and test just 3.12.

Rotzbua:
Same as API uses.

## Reference

fixes #1223


peterthomassen (approved):
thanks!

PR #1225, opened 2026-08-08
chore: update Dependabot configuration

### LINKED ISSUE #1222: Review usage of dependabot
## Current state

Since dependabot is currently disabled due to inactivity, maybe review the usage?

Currently dependabot is configured for all dependencies and weekly updates.

## GH Actions

For the github workflows a quarterly interval is enough.
-> change interval

## NPM

The npm dependencies are not pinned by a package-lock file, so dependabot does only suggest mayor updates which are not useful because often migrations have to be done.
-> remove

## Docs

On docs there are no versioned dependencies. Dependabot has no effect.
-> Fix the the dependencies an pin them to mayors?
-> reduce interval. yearly?

## API & e2e2

Reduce interval or remove?

Rotzbua:
At the moment just a suggestion.

## Reference

fixes #1222

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,35 +1,17 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
-# Please see the documentation for all configuration options:
-# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# https://docs.github.com/code-security/reference/supply-chain-security/dependabot-options-reference
 
 version: 2
 updates:
-  - package-ecosystem: "github-actions" # See documentation for possible values
-    directory: "/" # Location of package manifests
+  - package-ecosystem: "github-actions"
+    directory: "/"
     schedule:
-      interval: "monthly"
+      interval: "quarterly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,39 +1,21 @@
...
-    groups:
-      npm-webapp:
-        update-types:
-          - "minor"
-          - "patch"
   - package-ecosystem: "pip"
     directory: "api/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
```
peterthomassen: api may have security implications, so weekly is preferred over monthly

peterthomassen (changes requested):
Thank you for taking this up!

.github/dependabot.yml
Rotzbua: I would call it stability implication rather than security implication.

The `dependabot.yml` configures the version upgrade dependabot. It only focuses on version upgrades which may include bug or security fixes.
The security dependabot can only be enabled in the repository setting `Advanced Security` -> `Dependabot security updates`. This version runs independent from the .yml config and is triggered if a security patch for a package is available.

peterthomassen (approved):
merci! looks good.

Issue #196, opened 2019-05-25
Wrap pdns write operations in two-phase commit


peterthomassen:
Currently, we

1. start a transaction and issue write commands to our API database
2. write to pdns using the pdns API
3. commit our transaction

The database commit needs to come after the pdns write operations because those may fail routinely (e.g. if record data are invalid).

However, the database commit is also expected to fail occasionally (e.g. if a concurring client creates an RRset while an identical transaction is pending). The result is that step 3 fails when 2 has succeeded, so that the API database and pdns end up out of sync.

We should mitigate this by switching to a two-phase protocol where a "prepare" stage is added between steps 1 and 2.

sshine:
While working on

- #1218
- #1226

I mapped the write path against pdns 4.8, which is what nslord and nsmaster pin.

The following notes are for dealing with this issue at some point in the future.

## The ordering premise

The issue gives the reason for committing after the pdns write as "those may fail routinely (e.g. if record data are invalid)". As far as I can tell that is no longer true:

- `RR_SET_TYPES_MANAGEABLE` is defined as `(BACKEND & VALIDATION) - UNSUPPORTED - AUTOMATIC`, so a type is only accepted from clients if dnspython can parse it.
- `desecapi/dns.py` mirrors pdns' stricter-than-RFC name checks on purpose, citing `dnsname.cc` line numbers, so local validation tracks pdns acceptance by design.
- A pdns rejection surfaces as 500, not 4xx. The only 4xx from that leg is 413, and it is raised before the request is sent.

So the ordering is buying defence in depth against validation drift, not a client-facing 4xx. That widens what is allowed to change.

## Divergence windows

The issue describes commit-fails-after-pdns-succeeded. Three more exist in `__exit__`:

- `api_do()` runs after `pdns_do()` in the same iteration, so a database failure there strands a pdns write before the commit is ever reached.
- Changes are applied one at a time, so a later change failing rolls back the database while earlier changes stay applied at pdns and PCH.
- `axfr_to_master()` and the `published` bump sit outside both the transaction and the `try`. `check-secondaries` selects zones by `published__gt=threshold`, so a zone that misses its bump is invisible to the watchdog, and `published` also feeds the `scavenge-unused` expiry path.

## Literal 2PC

Seems a little out of reach to me.

- pdns' `startTransaction`/`commitTransaction` are backend-internal and per-request. The HTTP API exposes no prepare/commit.
[...]

PR #1226, opened 2026-08-10
Fix stranded domain deletion (#1218)

### LINKED ISSUE #1218: api: a failed domain deletion strands the domain, permanently 500ing both DELETE and GET
## Summary

- A `DELETE /api/v1/domains/{name}/` can fail in a way that leaves the Django row in place while the nslord zone is gone.
- From then on `DELETE` and `GET /api/v1/domains/{name}/` both return `500` forever, while `GET /api/v1/domains/` still lists the domain.
- The domain holds a slot against `limit_domains` and cannot be removed through the API.
- ~Four~ five domains on my `simon+test@simonshind.dk` account are in this state right now and can be used to verify the problem.

I accidentally triggered this bug in a misconfigured test suite where a shared HTTP client got torn down mid-request as a race condition between parallel tests. I was able to repeatedly cause this with high probability (hence why there's 4 domains that fail to get deleted).

## To replicate

Run `DESEC_TOKEN=... just live-test` on ~[cd6d92a95fbbaf2a94e16b22b16987b086654546
](https://github.com/desec-community/desec-rs/commit/cd6d92a95fbbaf2a94e16b22b16987b086654546)~ [f483ba04517839ba5de7c7e17b3c74ccaaa17220](https://github.com/desec-community/desec-rs/commits/main/), but don't use your main account.

## Impact

- The account loses a domain slot with no self-service recovery.
- Clients that treat `DELETE` as idempotent, as RFC 9110 licenses, can trigger it by retrying a request whose response was lost.
- Silent: nothing distinguishes a stranded domain from a healthy one in the list view, which is where most clients look.

## Observed behaviour

~Four~ five names under `dedyn.io`, created and deleted through the API on 2026-08-07:

- `desec-rs-test-domain-18c996ffee6f27be.dedyn.io`
- `desec-rs-test-apex-18c996e967092783.dedyn.io`
- `desec-rs-test-import-18c996db68330be0.dedyn.io`
- `desec-rs-test-apex-18c996db6834403f.dedyn.io`
- `desec-rs-test-gone-18c9a5fd7eaa9309.dedyn.io`

The rrsets endpoint answers `200` while the detail endpoint answers `500`.

This puts the surviving data in the API database and the missing data in pdns.

## Suspected mechanism

Line references are against `bc6f3058d22e1422b77749ee814b8a770f4d0d99`.

Deletion makes three unguarded pdns calls in sequence, in [`pdns_change_tracker.py:98-101`](https://github.com/desec-io/desec-stack/blob/bc6f3058d22e1422b77749ee814b8a770f4d0d99/api/desecapi/pdns_change_tracker.py#L98-L101):

```python
def pdns_do(self):
    pdns.delete_zone_lord(self.domain_name)
    pdns.delete_zone_master(self.domain_name)
    pdns.update_catalog(self.domain_name, delete=True)
```

If any of them raises, [`PDNSChangeTracker.__exit__`](https://github.com/desec-io/desec-stack/blob/bc6f3058d22e1422b77749ee814b8a770f4d0d99/api/desecapi/pdns_change_tracker.py#L251-L256) rolls the database transaction back and re-raises as a bare `ValueError`, which surfaces as `500` rather than a typed error:

```python
except Exception as e:
    self.transaction.__exit__(type(e), e, e.__traceback__)
    exc = ValueError(...)
    raise exc from e
```

So a failure at `delete_zone_master` or `update_catalog`, after `delete_zone_lord` has already succeeded, drops the nslord zone and then restores the `Domain` row. The rollback covers the database but not the pdns writes already made, which is what the `# TODO introduce two phase commit protocol` on [line 240](https://github.com/desec-io/desec-stack/blob/bc6f3058d22e1422b77749ee814b8a770f4d0d99/api/desecapi/pdns_change_tracker.py#L240) anticipates and what's tracked in:

- #196

Every later `DELETE` then re-runs `delete_zone_lord` against a zone that is already gone, gets a non-2xx from pdns, and `_pdns_request` raises `PDNSException` ([`pdns.py:112`](https://github.com/desec-io/desec-stack/blob/bc6f3058d22e1422b77749ee814b8a770f4d0d99/api/desecapi/pdns.py#L112)). The state is self-perpetuating.

The read asymmetry follows from the same missing zone. `include_keys` is true only for `create` and `retrieve` ([`views/domains.py:95`](https://github.com/desec-io/desec-stack/blob/bc6f3058d22e1422b77749ee814b8a770f4d0d99/api/desecapi/views/domains.py#L95)), and the `keys` property calls `pdns.get_keys` ([`models/domains.py:181-183`](https://github.com/desec-io/desec-stack/blob/bc6f3058d22e1422b77749ee814b8a770f4d0d99/api/desecapi/models/domains.py#L181-L183)), which fetches `/zones/{id}/cryptokeys` from nslord. The list view never touches pdns, so it keeps working.

sshine:
## What

- Fixes #1218 
- Fixes HTTP 500s surfacing as Django HTML errors instead of JSON
- Fixes HTTP 500s when RequestEntityTooLarge should give HTTP 413

Does not introduce a two-phase commit:

- #196 

Let me know if you would rather have that than these fixes.

Or if you would like to have the two-phase commit afterwards.

## Pull request structure

The first commit contains tests that currently fail.

Subsequent commits fix those tests.

## Cleanup

I would love to have my account `simon+test@simonshine.dk` unbricked from the following undeletable domains:

- `desec-rs-test-domain-18c996ffee6f27be.dedyn.io`
- `desec-rs-test-apex-18c996e967092783.dedyn.io`
- `desec-rs-test-import-18c996db68330be0.dedyn.io`
- `desec-rs-test-apex-18c996db6834403f.dedyn.io`
- `desec-rs-test-gone-18c9a5fd7eaa9309.dedyn.io`

### DIFF
--- a/api/desecapi/tests/test_domains.py
+++ b/api/desecapi/tests/test_domains.py
@@ -285,9 +285,8 @@ class DomainOwnerTestCase1(DomainOwnerTestCase):
         self.assertStatus(response, status.HTTP_404_NOT_FOUND)
 
     def test_delete_my_domain_retry_after_partial_pdns_failure(self):
-        # A deletion that fails at nsmaster rolls the database back, but the nslord zone stays
-        # gone. The retry then hits a zone that no longer exists and must still succeed,
-        # otherwise the domain is stranded and keeps occupying a slot against limit_domains.
+        # A retried deletion must succeed against a zone the first attempt already removed,
+        # or the domain stays stranded and keeps occupying a slot against limit_domains.
         url = self.reverse("v1:domain-detail", name=self.my_domain.name)
         self.client.raise_request_exception = False
 
@@ -310,8 +309,8 @@ class DomainOwnerTestCase1(DomainOwnerTestCase):
             self.assertFalse(Domain.objects.filter(pk=self.my_domain.pk).exists())
 
     def test_delete_my_domain_pdns_failure_renders_json(self):
-        # Clients parse error bodies, so a failed commit must stay within the API's JSON
-        # error contract rather than falling through to Django's HTML error page.
+        # A failed commit must stay within the API's JSON error contract rather than falling
+        # through to Django's HTML error page.
         url = self.reverse("v1:domain-detail", name=self.my_domain.name)
         self.client.raise_request_exception = False
 
--- a/api/desecapi/tests/test_domains.py
+++ b/api/desecapi/tests/test_domains.py
@@ -876,8 +876,8 @@ import-me.example RRSIG A 13 2 3600 20220324000000 20220303000000 40316 @ 4wj6Zr
     def test_create_domain_atomicity(self):
         name = self.random_domain_name()
         with self.assertRequests(self.request_pdns_zone_create_422()):
-            with self.assertRaises(ValueError):
-                self.client.post(self.reverse("v1:domain-list"), {"name": name})
+            response = self.client.post(self.reverse("v1:domain-list"), {"name": name})
+            self.assertStatus(response, status.HTTP_500_INTERNAL_SERVER_ERROR)
             self.assertFalse(Domain.objects.filter(name=name).exists())
 
     def test_create_domain_punycode(self):
--- a/api/desecapi/tests/test_domains.py
+++ b/api/desecapi/tests/test_domains.py
@@ -284,6 +284,45 @@ class DomainOwnerTestCase1(DomainOwnerTestCase):
         response = self.client.get(url)
         self.assertStatus(response, status.HTTP_404_NOT_FOUND)
 
+    def test_delete_my_domain_retry_after_partial_pdns_failure(self):
+        # A deletion that fails at nsmaster rolls the database back, but the nslord zone stays
+        # gone. The retry then hits a zone that no longer exists and must still succeed,
+        # otherwise the domain is stranded and keeps occupying a slot against limit_domains.
+        url = self.reverse("v1:domain-detail", name=self.my_domain.name)
+        self.client.raise_request_exception = False
+
+        with self.assertRequests(
+            self.request_pdns_zone_delete(name=self.my_domain.name, ns="LORD"),
+            self.request_pdns_zone_delete_500(name=self.my_domain.name, ns="MASTER"),
+        ):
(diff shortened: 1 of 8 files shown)


peterthomassen:
Thank you for this. We're planning to change the internal flow so that api only talks to nslord, and nsmaster learns everything relevant from a catalog zone generated directly by nslord. This removes most of the problem. I'd therefore like to hold these fixes and first take a stab at the nslord/nsmaster API call flow.

api/desecapi/tests/test_domains.py
```diff
@@ -284,6 +284,45 @@ def test_delete_my_domain(self):
...
+        url = self.reverse("v1:domain-detail", name=self.my_domain.name)
+        self.client.raise_request_exception = False
+
+        with self.assertRequests(
+            self.request_pdns_zone_delete(name=self.my_domain.name, ns="LORD"),
+            self.request_pdns_zone_delete_500(name=self.my_domain.name, ns="MASTER"),
+        ):
+            response = self.client.delete(url)
+            self.assertStatus(response, status.HTTP_500_INTERNAL_SERVER_ERROR)
+            self.assertTrue(Domain.objects.filter(pk=self.my_domain.pk).exists())
```
peterthomassen, line 300: Did you actually run this? The way this is invoked here seems to have nothing to do with the logic that raises the error in the actual interaction with pdns.

sshine: At the time of submitting, I was still mocking PDNS and only ran Postgres.

I've managed to run all of Postgres, PDNS and its MariaDB.

What happens is PDNS gives a 404 on deleting a non-existent zone, and that is rethrown as a 500 in desec-stack.

sshine (commented):
> Thank you for this. We're planning to change the internal flow so that api only talks to nslord, and nsmaster learns everything relevant from a catalog zone generated directly by nslord. This removes most of the problem. I'd therefore like to hold these fixes and first take a stab at the nslord/nsmaster API call flow.

That makes a lot of sense. I agree that's a better approach.

Would you be interested in a different PR that only fixes the HTTP 500s that are sent as HTML?

PR #1232, opened 2026-08-15
tests(e2e2): cover the "..." RRset endpoint and bulk request ordering

### LINKED ISSUE #171: Remove or Test `rrset` endpoint (the Infamous "..." endpoint)

### LINKED ISSUE #226: Verify Tests for Bulk-Request RRset-Order
Bulk requests should be independent of the order the RRsets are given. Do we have tests to verify that? (This surfaced as in #220 we had a problem where pdns behavior depends on the order of RRsets given to the pdns API.)

acoseac:
Closes #171
Closes #226

Two commits, as discussed in review.

### 1. Drop the undocumented `{subname}@` URL form

The RRset detail endpoint accepted `rrsets/{subname}@/{type}/` as a side effect of the route that makes the zone apex addressable as `rrsets/@/{type}/`. Only the apex spelling is documented; for non-empty subnames the documented form is `rrsets/{subname}.../{type}/`, and nothing outside the test helpers used it.

The route is removed and the helpers in `base.py` now reverse the `...` form, which the suite already used elsewhere. The apex `@` form is unaffected. `rrsets/{subname}@/{type}/` falls through to the generic route, where `{subname}@` is not a valid subname, and returns 404.

### 2. Tests

The `...` form was covered by no test at all, since `base.py` reversed the `@` form throughout.

- `desecapi/tests/test_rrsets.py` covers the URL spellings over apex, plain, wildcard, dotted and underscore subnames, and that the removed form now returns 404
- `test/e2e2` keeps only what needs the full stack: that these characters survive the HTTP layer in front of the API, that the apex is not addressable by leaving the subname empty, and that a write made through such a URL reaches the DNS
- bulk ordering: the AAAA/CNAME replacement from #220 in both payload orders, for PATCH and PUT, in both directions, against the real pdns
- the positional error array turned out to be covered already by `test_bulk_patch_missing_invalid_fields_1`, so only the missing atomicity check is added next to it

`put()` is added to the e2e API client, which had `get`/`post`/`patch`/`delete` but no PUT.

### Verification

Both suites run against a full local stack, rebased on current main:

- API test suite: passes
- e2e2 suite: 338 passed, 2 skipped

### DIFF
--- a/test/e2e2/spec/test_api_rrset_bulk_order.py
+++ b/test/e2e2/spec/test_api_rrset_bulk_order.py
@@ -1,0 +1,59 @@
+import pytest
+
+from conftest import DeSECAPIV1Client
+
+
+# The outcome of a bulk request must not depend on the order in which the RRsets
+# are given. This does not come for free: pdns refuses to add a CNAME while a
+# conflicting RRset is still present, so the API applies deletions before
+# additions regardless of their position in the payload (desec-stack#220,
+# PowerDNS/pdns#7501). The unit tests cover this against a mocked pdns and for
+# one payload order only; these tests exercise both orders against the real
+# thing (desec-stack#226).
+
+TTL = 3600
+
+CNAME_RECORDS = ['elk.example.net.']
+AAAA_RECORDS = ['::1']
+
+
+def replacement_payload(subname: str, remove_type: str, add_type: str, add_records: list) -> list:
+    """Delete one RRset and create a conflicting one at the same subname, deletion first."""
+    return [
+        {'subname': subname, 'type': remove_type, 'ttl': TTL, 'records': []},
+        {'subname': subname, 'type': add_type, 'ttl': TTL, 'records': add_records},
+    ]
+
+
+@pytest.mark.parametrize("method", ['patch', 'put'])
+@pytest.mark.parametrize("deletion_first", [True, False])
+@pytest.mark.parametrize("init_rrsets", [{('kibana', 'AAAA'): (TTL, set(AAAA_RECORDS))}])
+def test_replace_rrset_with_cname(api_user_domain_rrsets: DeSECAPIV1Client, method: str, deletion_first: bool):
+    api = api_user_domain_rrsets
+    payload = replacement_payload('kibana', 'AAAA', 'CNAME', CNAME_RECORDS)
+    if not deletion_first:
+        payload.reverse()
+
+    response = getattr(api, method)(f"/domains/{api.domain}/rrsets/", data=payload)
+    assert response.status_code == 200
+    api.assert_rrsets({
+        ('kibana', 'AAAA'): (TTL, set()),
+        ('kibana', 'CNAME'): (TTL, set(CNAME_RECORDS)),
+    })
+
+
+@pytest.mark.parametrize("method", ['patch', 'put'])
+@pytest.mark.parametrize("deletion_first", [True, False])
+@pytest.mark.parametrize("init_rrsets", [{('kibana', 'CNAME'): (TTL, set(CNAME_RECORDS))}])
+def test_replace_cname_with_rrset(api_user_domain_rrsets: DeSECAPIV1Client, method: str, deletion_first: bool):
+    api = api_user_domain_rrsets
+    payload = replacement_payload('kibana', 'CNAME', 'AAAA', AAAA_RECORDS)
+    if not deletion_first:
+        payload.reverse()
+
+    response = getattr(api, method)(f"/domains/{api.domain}/rrsets/", data=payload)
+    assert response.status_code == 200
+    api.assert_rrsets({
+        ('kibana', 'CNAME'): (TTL, set()),
+        ('kibana', 'AAAA'): (TTL, set(AAAA_RECORDS)),
+    })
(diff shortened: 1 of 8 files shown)


test/e2e2/spec/test_api_rrset_bulk_order.py
```diff
@@ -0,0 +1,80 @@
...
+    payload[position] = {'subname': 'invalid', 'type': 'A', 'ttl': -1, 'records': ['1.2.3.4']}
+
+    response = api.patch(f"/domains/{api.domain}/rrsets/", data=payload)
+    assert response.status_code == 400
+    errors = response.json()
+    assert [bool(error) for error in errors] == [i == position for i in range(3)]
+    assert 'ttl' in errors[position]
+
+    # bulk requests are atomic, so the valid parts were not applied either
+    assert api.get(f"/domains/{api.domain}/rrsets/?type=A").json() == []
```
peterthomassen: This test hits the API only. Does it needs to live as an e2e test?

test/e2e2/spec/test_api_rrset_url.py
```diff
@@ -0,0 +1,106 @@
+import pytest
+
+from conftest import DeSECAPIV1Client
+
+
+# The subname part of an RRset URL can be written in up to three ways, see
+# https://desec.readthedocs.io/en/latest/dns/rrsets.html#accessing-the-zone-apex
+#
+#     {subname}/{type}/     plain
+#     {subname}@/{type}/    '@' terminates the subname
```
peterthomassen: This is actually only documented for the empty subname: `rrsets/@/<type>/`. For non-empty subname, `rrsets/<subname>.../<type>/` is document, but not `rrsets/<subname>@/<type>/` (although it is used in testing).

To minimize unneeded/undocumented or testing-only code, perhaps we should remove the `<subname>@` form (switching tests to `...` form). What do you think?

I personally also feel that URLs like `domains/example.com/rrsets/sub@/A/` are less intuitive than `domains/example.com/rrsets/sub.../A/` (albeit 2 characters shorter), but that's subjective.

test/e2e2/spec/test_api_rrset_url.py
```diff
@@ -0,0 +1,106 @@
...
+
+def test_apex_not_addressable_with_empty_subname(api_user_domain: DeSECAPIV1Client):
+    # This is the reason the '@' and '...' forms exist: the double slash does not
+    # survive URL normalization, so the apex cannot be addressed by omitting the
+    # subname.
+    api = api_user_domain
+    assert api.rr_set_create(api.domain, 'A', ['1.2.3.4'], ttl=TTL).status_code == 201
+
+    assert api.get(f"/domains/{api.domain}/rrsets//A/").status_code == 404
+    assert api.get(f"/domains/{api.domain}/rrsets/.../A/").status_code == 200
```
peterthomassen: These tests do not use the DNS; they hit the API only. Do they need to be e2e tests?

peterthomassen (commented):
Thank you for engaging with our backlog! <3

test/e2e2/spec/test_api_rrset_url.py
acoseac: Thanks for the review!

Agreed, let's drop it.

Removing the route itself and switching `base.py` to `...` is an API change, so probably its own PR. Only caveat: it's technically breaking for anyone who found it, undocumented or not.

test/e2e2/spec/test_api_rrset_bulk_order.py
acoseac: That one is pure API. Turns out `test_bulk_patch_missing_invalid_fields_1` already covers the positional errors, so I'll drop it here and add just the missing atomicity check to the unit suite.

test/e2e2/spec/test_api_rrset_url.py
acoseac: Proposal is to keep a thin e2e test for the URL shapes and move the exhaustive matrix to unit tests.

test/e2e2/spec/test_api_rrset_url.py
peterthomassen: In the last 31 days, we had 1 user doing a GET request like that. I've informed them and I think it's fine to drop it.

If you don't mind, I'd prefer this PR to address the full topic, that is, taking a fresh look at the situation, which leads to dropping the one route and adding tests for the other. I agree it should be different commits. How does that sounds to you?

test/e2e2/spec/test_api_rrset_url.py
peterthomassen: sure, doesn't hurt!

test/e2e2/spec/test_api_rrset_url.py
acoseac: Sounds good, will do it here in two commits: one dropping the route and switching the test helpers to the `...` form, one for the tests.

acoseac:
Pushed, rebased on current main, two commits as discussed.

One thing worth flagging: the positional error array is already covered by `test_bulk_patch_missing_invalid_fields_1`, so I dropped that test rather than porting it, and added only the missing atomicity check next to it.

Ran both suites against a full local stack: API tests pass, e2e2 338 passed and 2 skipped.

PR #1174, opened 2026-01-10
Adds delegation checker

nils-wisiol:
- periodically checks delegation status of all domains
- implements limits on insecure domains per user
- sets default limit to unlimited
- updates the UI to show the detected delegation status
- updates the UI to allow triggering of the delegation check for a given domain

delegation checker run times on my laptop on a 1000 domains sample (5 repeats):

| Threads | Mean (s) | Stdev (s) | Speedup vs 1x | 100k extrapolation |
|--------:|---------:|----------:|--------------:|--------------------|
| 1       | 178.07   | 21.53     | 1.00x         | ~296.8 min (4.95 h) |
| 10      | 39.12    | 13.79     | 4.55x         | ~65.2 min (1.09 h)  |
| 20      | 33.20    | 8.06      | 5.36x         | ~55.3 min (0.92 h)  |
| 50      | 23.89    | 6.38      | 7.46x         | ~39.8 min (0.66 h)  |
| 100     | 26.10    | 6.14      | 6.82x         | ~43.5 min (0.73 h)  |
| 200     | 27.16    | 6.15      | 6.55x         | ~45.3 min (0.75 h)  |

The delegation checker is not comparing keys against nslord to avoid making many requests to nslord.

PR #1235, opened 2026-08-16
Delegation checker and automatic limit_domains management

peterthomassen:
Different approach than #1174

Testing using production database on local machine takes about 79 minutes with 4 threads, 47 minutes with 8 threads, 44 minutes with 16 threads, and 39 minute with 64 threads. **8 threads is probably best**: the more threads, the more resolution errors (maybe rate limits at TLD auths or something).


peterthomassen:
Extended PR with automatic domain limit management. Effective domain limit is number of domains securely delegated to us (`s`, includes .dedyn.io domains), plus the square root of *external* domains securely delegated to us (`e`, without .dedyn.io domains; but: at least 1). This way, the headroom for creating more domains without immediately securing them (DS record) grows sensibly:

```
limit = s + max(1, h)    with  h = round(√e)
```

For users without .dedyn.io domains, this is equivalent to (e + sqrt(e)); .dedyn.io domains are added on top (effectively not influencing the limit besides offsetting it).

Illustrations:

# Account without dedyn.io domain
All secure domains `s` are external.
| Secure domains `s` | Headroom `h = round(√s)` | Effective limit |
| ---: | ---: | ---: |
| 0 | 0 | 1 |
| 1 | 1 | 2 |
| 2 | 1 | 3 |
| 3 | 2 | 5 |
| 4 | 2 | 6 |
| 5 | 2 | 7 |
| 7 | 3 | 10 |
| 10 | 3 | 13 |
| 12 | 3 | 15 |
| 15 | 4 | 19 |
| 20 | 4 | 24 |
| 25 | 5 | 30 |
| 30 | 5 | 35 |
| 40 | 6 | 46 |
| 50 | 7 | 57 |
| 75 | 9 | 84 |
| 100 | 10 | 110 |

# Account with dedyn.io domain
Secure domains `s` include the .dedyn.io domain. The limit is effectively raised by 1 for the first few domains. From ~10 domains on, the situation is the same as above.
| Secure domains `s` | Headroom `h = round(√(s−1))` | Effective limit |
| ---: | ---: | ---: |
| 1 | 0 | 2 |
| 2 | 1 | 3 |
| 3 | 1 | 4 |
| 4 | 2 | 6 |
| 5 | 2 | 7 |
| 7 | 2 | 9 |
| 10 | 3 | 13 |
| 12 | 3 | 15 |
| 15 | 4 | 19 |
| 20 | 4 | 24 |
| 25 | 5 | 30 |
| 30 | 5 | 35 |
| 40 | 6 | 46 |
| 50 | 7 | 57 |
| 75 | 9 | 84 |
| 100 | 10 | 110 |

peterthomassen:
I've performed a detailed/thorough review, and applied various improvements / rearranged commits. I did not find any problematic issues.

The first two commits initially might seem unrelated, but they are needed so that the later limit calculation logic works reliably.

Perhaps we should drop the commits containing the plans.

@nils-wisiol ready for final review

nils-wisiol:
# PR 1235 Review: Delegation Checks and Automatic Domain Limits

Compared: `origin/main...origin/pr/1235`

## Checklist

- [x] **High: Fix delegation false positives for nested/co-hosted zones.** `api/desecapi/delegation.py:214`, `api/desecapi/delegation.py:326`

  The checker queries parent nameserver addresses directly, but accepts NS RRsets
  from both answer and authority sections. If the same authoritative server hosts
  both parent and child zones, an authoritative child-zone answer can be mistaken
  for parent-side delegation. For quota purposes, only parent referral data should
  count. Add a co-hosted parent/child regression test.

- [ ] **High: Reserve execution capacity for ad-hoc delegation checks.** `docker-compose.yml:303`, `api/desecapi/tasks.py:27`, `api/desecapi/dns.py:154`, `api/desecapi/unbound.py`

  One `celery-delegation` worker consumes both `delegation_adhoc` and
  `delegation_bulk` with concurrency 4 and `--prefetch-multiplier 1`. That avoids
  deep prefetching, but does not bound user wait time when all slots are occupied
  by bulk checks. Run separate worker pools for `delegation_adhoc` and
  `delegation_bulk`; AMQP priority can be an extra ordering hint, but not the
  latency guarantee.

  Capacity note: with about 50k non-local-public-suffix domains, cron checks
  roughly `50,000 / 11 = 4,545` domains every 2h. With four workers, the system
  needs to average better than about 6.3s per domain to avoid falling behind.

- [x] **High: Prevent duplicate bulk-check backlog.** `api/cronhook/crontab:4`, `api/desecapi/management/commands/check-delegation.py:124`, `api/desecapi/tasks.py:116`

  Cron enqueues stale checks every 2h:

  ```text
  23 */2 * * * manage.py check-delegation --stale 79200 --concurrency 0
  ```

  Workers re-check staleness before DNS work, but if the queue falls behind,
  later cron runs can enqueue the same still-stale domains again. Celery does not
  provide enqueue-time uniqueness for "one pending check per domain". Add an
  app-level pending marker, e.g. `delegation_check_pending` or a pending task id,
  set it atomically when enqueueing, skip already-pending domains, and clear it
  after the task's freshness re-check completes. A Redis/RabbitMQ lock keyed by
  domain id and queue type would also work, with a TTL longer than worst-case task
  runtime.

- [ ] **Medium: Expose per-domain delegation status through the API.** `api/desecapi/models/delegation.py:71`, `api/desecapi/serializers/users.py:32`, `api/desecapi/serializers/domains.py:22`

  The PR stores per-domain security status, nameserver status, nameserver list,
  EDE, rcode, and timestamps internally, but public responses only expose
  aggregate `secure_domains` and effective `limit_domains` on the user object.
  Users need to know which domain is blocking quota, whether a re-check is
  pending, what failed, when it last ran, and what to fix.

- [x] **Medium: Align docs with the self-service limit flow.** `docs/auth/account.rst:270`, `docs/dns/domains.rst:162`, `api/desecapi/permissions.py:169`

  The account docs mention limits growing with secure delegation, but do not
  document the formula, freshness, status meanings, expected wait time, or how to
  inspect/fix specific domains. The domain docs still primarily say to contact
  support when hitting the limit, while the API error says the system is
  re-checking delegation.

nils-wisiol:
- [x] does ad hoc queue have precedence over bulk items so that users do not get blocked behind bulk workload?
- [x] can we increase concurrency on the workers (for ~free) to increase delegation checker workload headroom? AI memory estimate: `delegation worker PSS ~= 95 MiB + 12.5-13 MiB per concurrency slot`
- [x] close on LPS domain limit - implement a solution to increase it per account?
  - PR status quo: hardcoded LPS limit of 1 plus total domain limit as set by the per-account limit 
  - make it backwards compatible by disabling new logic when per-account limit is not null
  - plan to introduce separate LPS and regular domain limits later

peterthomassen:
> * [ ]  does ad hoc queue have precedence over bulk items so that users do not get blocked behind bulk workload?

It does not take precedence, but workers consume both queues round-robin, and bulk tasks are enqueued per domain. Thus, when an ad-hoc check task is added, a worker will pick it up quickly, then go back to pending bulk tasks.

Effectively, adhoc queue gets half the slots as long as it is not empty. Nothing to be gained by separating workers.

Hard cut-off for a check is 30s, expected time for any slot to free up is < 2s.

> * [ ]  can we increase concurrency on the workers (for ~free) to increase delegation checker workload headroom? AI memory estimate: `delegation worker PSS ~= 95 MiB + 12.5-13 MiB per concurrency slot`

Increased from 4 to 8. Potential constraints are TLD nameserver rate limits (probably far away), queuing within unbound (we give it `$(nproc)` workers, that's currently 8), and memory impact. We are not in a rush, so I think 8 is fine.

> * [ ]  close on LPS domain limit - implement a solution to increase it per account?
>   
>   * PR status quo: hardcoded LPS limit of 1 plus total domain limit as set by the per-account limit
>   * make it backwards compatible by disabling new logic when per-account limit is not null
>   * plan to introduce separate LPS and regular domain limits later

Done.

Issue #1190, opened 2026-02-10
Enable HTTP3


bwbroersma:
```bash
$ curl --http3 -sSI https://desec.io | head -n1
HTTP/2 200 
```

HTTP3 in nginx can be enabled (and `HTTPS` record set), in nginx:
```nginx
http3 on;
```
\+ duplicate the listen with a `quic` instance, see https://nginx.org/en/docs/quic.html#configuration

Then also a `HTTPS` record can be added to the DNS zone:
```zone
desec.io. IN HTTPS  1 . alpn=h2,h3 
```

---
At internet.nl we needed [these](https://github.com/internetstandards/Internet.nl/pull/1660/changes) changes (and forgot to expose [udp in docker](https://github.com/internetstandards/Internet.nl/pull/1680/changes)).

peterthomassen:
Done for static content under desec.io. Keeping this open until full rollout.

PR #1238, opened 2026-08-17
feat(www): enable http3

Rotzbua:
For first testing only website and minimal configuration.

### Reference

mentioned in #1190

### DIFF
--- a/www/90-desec.static.dev.location
+++ b/www/90-desec.static.dev.location
@@ -1,6 +1,7 @@
 location / {
     expires epoch;
     etag off;
+    add_header Alt-Svc 'h3=":443"; ma=86400' always;
     add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" always;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header Host "0.0.0.0";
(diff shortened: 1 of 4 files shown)


peterthomassen:
> For first testing only website

Includes API (just fyi).

> minimal configuration

What would be a non-minimal configuration?

Rotzbua:
> For first testing only website
> Includes API (just fyi).

I intentionally just activated http3 just on the static website for a really minimal first test 🔍 .

> What would be a non-minimal configuration?

Deployment the complete nginx server so it is deployed over all subdomains like update. checkip*.
Furthermore the deployment as https dns registry.

peterthomassen:
> > > For first testing only website
> > Includes API (just fyi).
> 
> I intentionally just activated http3 just on the static website for a really minimal first test 🔍 .

My point is that the API runs under the same hostname, so it affects the static website and the API. But I guess, why not.

peterthomassen:
I see now, you meant the Alt-Svc header.

www/90-desec.static.dev.location
```diff
@@ -1,6 +1,7 @@
 location / {
     expires epoch;
     etag off;
+    add_header Alt-Svc 'h3=":443"; ma=86400' always;
```
peterthomassen, line 4: I added this for local testing.

peterthomassen (approved):
thx!

PR #1244, opened 2026-08-19
Add Datadog Synthetic tests workflow

jorgemenendez142-sys:
This workflow triggers Datadog Synthetic tests on push and pull request events to the main branch, using the Datadog API and Application keys stored as GitHub secrets.


peterthomassen:
We do not use Datadog Synthetic tests, so I take it that this is a mistake. Let me know if I misunderstood.

PR #1247, opened 2026-08-20
TTL treatment, webapp fix

peterthomassen:
To facilitate lower TTL minimum for custom configured records, to reduce support load. GUI now fills in default TTL (3600), not minimum.


docs/dns/domains.rst
```diff
@@ -48,6 +49,17 @@ Field details:
     Timestamp of domain creation, in ISO 8601 format (e.g.
     ``2013-01-29T12:34:56.000000Z``).
 
+``default_ttl``
+    :Access mode: read-only
+
+    TTL that we recommend for :ref:`RRsets <RRset object>` in this domain, for
+    use when no specific TTL is desired.  Our web interface pre-fills the TTL
+    field with this value; you may want to do the same in your own tooling.
```
nils-wisiol: don't document the web ui in the api docs

nils-wisiol (approved):

nils-wisiol:
Updated this branch per the later direction and force-pushed a clean stack:

- `f2599f2b feat(api): lower default minimum TTL`
- `9d7fb638 feat(webapp): compile default TTL into record form`
- `cb2c17c1 fix(webapp): re-create views when route params change`

The PR no longer introduces a `Domain.default_ttl` field, serializer field, migration, API setting, or management-command handling. The record form default now comes from the webapp build-time config `DESECSTACK_WEBAPP_DEFAULT_TTL`, exposed as `VITE_APP_DEFAULT_TTL`, similar to the other compiled webapp settings. The webapp still reads the domain `minimum_ttl` from the API and pre-fills `max(VITE_APP_DEFAULT_TTL, minimum_ttl)` so the suggested TTL is valid for domains with a higher per-domain minimum.

Validation: `npm run lint` passed in `www/webapp`.

nils-wisiol (approved):

peterthomassen:
pushed for signing

PR #1231, opened 2026-08-14
feat(www): drop DHE TLS ciphers

Rotzbua:
Update nginx TLS config to TLSRef v6.0.
SSL version according to current nginx docker.

### Reference

closes #1189

PR #1250, opened 2026-08-25
fix(www): remove unused DH parameters

Rotzbua:
DHE TLS ciphers are dropped.

## Reference

Fixup #1231


peterthomassen (approved):
Indeed this file was now unused (`ssl_dhparam` was removed in #1231 and has no default), so this is removing the orphaned file. Thank you!

peterthomassen (approved):
thanks. (rebased)

PR #1254, opened 2026-09-01
A number of webapp / CI / docs fixes

peterthomassen:


docs/auth/tokens.rst
```diff
@@ -101,7 +101,7 @@ Field details:
 
 ``max_age``
     :Access mode: read, write
-    :Type: string (time duration: ``[DD] [HH:[MM:]]ss[.uuuuuu]``) or ``null``
+    :Type: string (time duration: ``[DD] [[HH:]MM:]ss[.uuuuuu]``) or ``null``
```
nils-wisiol, line 104: Is a bare number really interpreted as minutes?

nils-wisiol (approved):

docs/auth/tokens.rst
peterthomassen: no, seconds (I think you misread. - the change is that xx:yy was explained as HH:ss wheras it actually means MM:ss, as one would expect.)

PR #1257, opened 2026-09-02
Introduce registration gatekeeper

peterthomassen:


api/desecapi/exceptions.py
```diff
@@ -15,6 +15,14 @@ class RequestEntityTooLarge(APIException):
     default_code = "too_large"
 
 
+class RegistrationDenied(APIException):
+    status_code = status.HTTP_400_BAD_REQUEST
+    default_detail = (
+        "Registration denied. If you believe this is an error, please contact support."
```
nils-wisiol, line 21: "denied by policy"?

api/desecapi/gatekeeper.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    payload = {"event": event, **facts}
+    duration = metrics.get("desecapi_gatekeeper_request_duration_seconds").labels(event)
+    try:
+        with duration.time():
+            response = requests.post(
+                settings.GATEKEEPER_API,
+                json=payload,
+                headers={"User-Agent": "desecapi"},
+                timeout=settings.GATEKEEPER_TIMEOUT,
```
nils-wisiol: It's fair to wait 2s for the decision, but the tcp timeout should be much shorter to fail fast when the service is down

api/desecapi/gatekeeper.py
```diff
@@ -0,0 +1,53 @@
...
+        metrics.get("desecapi_gatekeeper_request_failure").labels(event).inc()
+        return ALLOW
+
+    metrics.get("desecapi_gatekeeper_verdict").labels(event, verdict).inc()
+    if verdict != ALLOW:
+        reason = str(answer.get("reason"))[:REASON_MAX_LENGTH]
+        logger.warning(
+            "Gatekeeper verdict on %s for %s: %s (%s)",
+            event,
+            facts.get("email"),
```
nils-wisiol, line 49: I think it's fair to do this but wanted to point out that in other places we avoid logging PII

api/desecapi/views/users.py
```diff
@@ -35,15 +36,26 @@ def create(self, request, *args, **kwargs):
...
+                raise RegistrationDenied
+            if verdict == gatekeeper.ALLOW:
+                # create user
+                user = serializer.save(is_active=None if activation_required else True)
+
+                # send email if needed
+                if domain or activation_required:
+                    serializers.AuthenticatedActivateUserActionSerializer.build_and_save(
+                        user=user, domain=domain
+                    )
```
nils-wisiol, line 57: Not sure why DROP doesn't show up here. If it is fall-thorugh, perhaps it's worth making that explicit

nils-wisiol (approved):

api/desecapi/gatekeeper.py
peterthomassen: yes, ack. conscious choice here

api/desecapi/views/users.py
peterthomassen: drop is silent, will add comment

peterthomassen: actually, moving to a match construction where DROP has pass

api/desecapi/gatekeeper.py
peterthomassen: changing to a timeout tuple where the first value is connect timeout

PR #1264, opened 2026-09-17
Fix canonicalization of LOC records to match PDNS' expectations

dairiki:
The PowerDNS auth server API wants record values to be in (PDNS') canonical form.  If they are not, requests for record creation or replacement via the PDNS API  will be rejected.
To accomodate this, the deSEC API uses `dnspython` to canonicalize record values.

In the case of LOC records, PDNS and dnspython disagree on what the canonical format looks like.
Specifically, if all of the last three fields of the record (those fields being _size_, _horizontal precision_, _vertical precision_) have their default values (1m, 10000m, 10m respectively), dnspython [omits those fields](https://github.com/rthalley/dnspython/blob/fb2979f909e79390682feace28b6c0e4b5bb81dc/dns/rdtypes/ANY/LOC.py#L175-L180) from its representation.  PDNS expects those fields to be present, even if they are all set to default values.

A symptom of this is that attempts to create or update a LOC record using deSEC's API will fail (with HTTP status 500) unless at least one of the size or precision fields is set to a non-default value.

This PR addresses that by fixing up dnspython's serialization for LOC records.


dairiki:
~Note that I haven't yet got the full stack running here locally, so I haven't verified, in practice, that this actually fixes the issue.  I'm still working on that.~

[ETA: I've gotten the full stack running, and have made a few basic tests creating and modifying LOC records via the deSEC API.  With these patches, things appear to work.]

api/desecapi/dns.py
```diff
@@ -148,6 +148,20 @@ class SRV(_NameMixin("target", allow_root=True), dns.rdtypes.IN.SRV.SRV):
...
+@dns.immutable.immutable
+class LOC(dns.rdtypes.ANY.LOC.LOC):
+    def to_text(self, *args, **kwargs):
+        text = super().to_text(*args, **kwargs)
+        word_count = len(text.split())
+        if word_count != 12:
+            # Add missing default-valued fields.
+            # Dnspython omits the last three fields (size, hprec, vprec) if they are
+            # all set to their default values.  Pdns wants all the fields, all the time.
+            assert word_count == 9
```
peterthomassen: Let's do `if len(text.split()) == 9: text += ...` so that cases with 10 or 11 get passed on, in case that would somehow happen.

By doing that instead of raising the assertion error here, Django will send us an error report, so we can look into it further. The current code would fail silently, but for unknown error modes we prefer being sent something so we can derive some insight.

peterthomassen (changes requested):
thx! Suggested a change.

api/desecapi/dns.py
dairiki: Done. I just force-pushed a revised version.

peterthomassen (approved):

peterthomassen:
Thank you very much! Will squash and sign, then merge.

peterthomassen:
deployed.

